PR #12, opened 2019-04-19
Add experimental custom pragma

jxnblk:
Trying out the idea of using a custom pragma for the Styled System `css` call. This uses Emotion's `jsx` create element function and is intended to be a replacement for that. I've use the `@jsx` pragma call in the docs to show explicitly what's happening and disabled the emotion plugin. Not sure if this is worth exploring more but wanted to give it a shot cc @johno @ChristopherBiscardi

### DIFF
--- a/theme-ui/index.js
+++ b/theme-ui/index.js
@@ -6,11 +6,16 @@ import { MDXProvider } from '@mdx-js/react'
 import css from '@styled-system/css'
 import merge from 'lodash.merge'
 import get from 'lodash.get'
+import omit from 'lodash.omit'
 
 export { css } from '@styled-system/css'
 
 // custom pragma
 const spaceRegExp = /^[mp][trblxy]?$/
+const spaceProps = [
+  'm', 'mt', 'mr', 'mb', 'ml', 'mx', 'my',
+  'p', 'pt', 'pr', 'pb', 'pl', 'px', 'py',
+]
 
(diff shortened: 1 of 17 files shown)


johno:
🔥🔥🔥

This is super interesting and I think being able to remove the `css={css({` is a substantial improvement in DX.

theme-ui/index.js
```diff
@@ -1,14 +1,72 @@
...
+const compose = (...funcs) => (type, props, ...children) => {
+  return funcs.reduce((el, fn) => {
+    return fn.apply(undefined, [ el.type, el.props, ...children ])
+  }, React.createElement.apply(undefined, [ type, props, ...children ]))
+}
+
+export const jsx = compose(
+  system,
+  emotion
+)
```
jxnblk: FWIW I tried throwing the `mdx` create element function in here as well, and it *seemed* to work as expected, replacing e.g. `<h1>` with a component from context – something to think about as an alternative to the `Styled.h1` components

ChristopherBiscardi:
Very promising direction IMO (sidenote: we've now all written custom pragmas for themes'mdx stuff lol).

theme-ui/index.js
ChristopherBiscardi: That is really awesome. We should put some thought into using the mdx createElement function more generally (even though it seems to work).

PR #13, opened 2019-04-19
Add Reset component

jxnblk:


ChristopherBiscardi:
While I agree box-sizing is a better default, We can't guarantee that these styles won't break someone :/

If we're going to include global resets, I think we need to take a stronger position and include it in core somehow. Otherwise we need to do these resets on a component level.

jxnblk:
Yeah, we have `box-sizing` on the layout components already so removed that. If we have a better plan for reseting body margin, I'm all ears

jxnblk:
Closing this after discussion that we might want to handle this with a different approach

PR #15, opened 2019-04-26
Add node_modules to gitignore

macklinu:
After cloning and installing dependencies, `node_modules` were being tracked by git. This prevents that.


johno:
Thanks!

PR #22, opened 2019-05-10
Use pragma directly

johno:
Current implementations using theme-ui are transpiling
theme-ui directly which causes gatsby-plugin-emotion to
override the jsx pragma. This uses the pragma directly
so that we can ensure this doesn't happen.

### DIFF
--- a/theme-ui/layout.js
+++ b/theme-ui/layout.js
@@ -1,4 +1,3 @@
-/** @jsx jsx */
 import { jsx } from './index'
 import styled from '@emotion/styled'
 import {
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
> which causes gatsby-plugin-emotion to
override the jsx pragma.

wild. I thought an inline pragma would take precedence

johno:
> wild. I thought an inline pragma would take precedence

I thought so, too. Turns out that emotion had already run and _then_ was calling theme-ui pragma. Didn't dig into the "why" yet tho. Hence the quick, hacky patch : )

jxnblk:
I think it was because the plugin adds the pragma with a Babel plugin *and* we’re using compile packages plugin on theme-ui but I might be wrong

PR #25, opened 2019-05-15
Add color modes and useColorMode hook

jxnblk:
This adds a color mode state to the ThemeProvider, allowing end users to change this state with the `useColorMode` hook. This is similar to the implementation in `use-dark-mode` but custom built to better support how theme-ui works with Emotion's theme context and its own internal context. This PR also includes updates to the docs

- [x] Add tests for nested ThemeProviders
- [x] Add support for initialization from the `(prefers-color-scheme: dark)` media query

### DIFF
--- a/docs/src/components/sidebar.js
+++ b/docs/src/components/sidebar.js
@@ -1,42 +1,68 @@
 /** @jsx jsx */
+import React from 'react'
 import { jsx, Box } from 'theme-ui'
+import { MDXProvider } from '@mdx-js/react'
+import NavLink from './nav-link'
 import Content from '../sidebar.mdx'
 
-export default props =>
-  <Box {...props}
-    css={{
-      position: 'sticky',
-      top: 0,
-      minWidth: 0,
-      width: [ 256 ],
-      flex: 'none',
-      px: 3,
-      py: 3,
-      maxHeight: '100vh',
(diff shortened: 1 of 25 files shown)


docs/src/components/sidebar.js
```diff
@@ -1,45 +1,68 @@
...
+          marginLeft: -256,
+          transition: 'transform .2s ease-out',
+          transform: props.open ? 'translateX(100%)' : 'translateX(0)',
+        },
+        ul: {
+          listStyle: 'none',
+          m: 0,
+          p: 0,
+        },
+        __a: {
```
ChristopherBiscardi: what is `__a`?

ChristopherBiscardi:
Still reading the PR, but it seems like this only handles colors for modes. Wouldn't it be reasonable to also change spacing, font size, etc for dark vs light themes (for ex: readability)?

docs/src/components/sidebar.js
jxnblk: dead code that needs ✂️

jxnblk:
My thinking for now is to solve this *only* for colors because: 1) it's a fairly common use case, 2) swapping colors on the client won't cause shifts in layout, and 3) wholesale theme swapping could still be implemented in userland

PR #26, opened 2019-05-21
Add inherit prop to ThemeProvider

jxnblk:
This allows parent ThemeProvider components to override child ThemeProviders' `theme` object – there's probably a better approach for this, but using this for trying out different ways to handle overriding themes


jxnblk:
Closing this out since this will likely be handled with gatsby plugins for now

PR #28, opened 2019-05-22
WIP Gatsby theme context plugin

jxnblk:
- Currently only supports one instance
- Ideally multiple instances of something like this could merge theme objects together and allow a root theme object to override any theme objects used in Gatsby themes


jxnblk:
Closing in favor of #30

PR #30, opened 2019-05-23
Add Gatsby plugin

### LINKED ISSUE #33: Gatsby Plugin
Add a Gatsby plugin to control how theme and component context is handled when using multiple Gatsby themes in a site. 

This plugin should:

- Provide a zero-config way to add a `src/theme` module to the theme context
- Merge multiple themes into a single theme object
- Allow the site author to completely override any value in a child theme
- Provide an option for enabling and disabling color mode state (or handle this with a separate plugin)
- Allow MDX components to be defined (or provide this as a separate plugin)

jxnblk:
- Adds a `gatsby-plugin-theme-ui` package
- The plugin looks for a `src/theme` module in a users site or in installed Gatsby themes
- Simplifies the color mode API
  - Color mode is enabled by including an `theme.initialColorMode` value
  - The `ThemeProvider` uses the color mode state hook
  - `ColorModeProvider` is deprecated
  - A new `ColorMode` component is used for adding color and background color styles to the `<body>` - currently this uses Emotion's `Global` component, but the implementation can be adjusted in the future
- The plugin uses an inline script to help prevent a flash of colors on page load (when a user has a non-default color mode set)

Fixes #33


gatsby-plugin-theme-ui/theme-loader.js
```diff
@@ -0,0 +1,20 @@
...
+  return `
+    export const themes = [
+      ${themes.map(filename => `require('${filename}').default`)}
+    ]
+
+    export default {
+      themes,
+    }
+
+    if (module.hot) module.hot.accept()
```
jxnblk: Hot reloading doesn't seem to pick up changes. This was an attempt to make it work

ChristopherBiscardi: you'd need `this.addDependency` to re-trigger the loader iirc https://webpack.js.org/api/loaders#thisadddependency

gatsby-plugin-theme-ui/README.md
```diff
@@ -0,0 +1,27 @@
...
+  plugins: [
+    {
+      resolve: 'gatsby-plugin-theme-ui',
+      options: {
+        // required
+        theme: require.resolve('./src/theme.js'),
+        // optional
+        components: require.resolve('./src/components.js'),
+        // optional, initial color mode
+        colorMode: 'light',
```
ChristopherBiscardi: `defaultColorMode` or something seems more obvious. It's also possible that this color mode doesn't exist if someone overrides the tokens right?

ChristopherBiscardi:
What's the intended effect here? I still feel weird about the magic file/global ThemeProvider merging API so want to make sure I'm understanding the intent correctly. It *seems* like you could get away with no config in the plugin, no writing to `.cache`, and instead passing `store.getState().plugins` to the webpack loader. The loader can be called in `wrapRootElement` like `theme-ui/loader!`. Note that there's no filename after `!` which means we use the store state to iterate over the plugins, checking if each of them (or the user's site) have the magic files at the right location. The webpack loader can return the existing `require`s and voila we have a merged object with the themes/mdx components imported into `wrapRootElement` for use in the Provider.

If there's no plugin config then there's no problem with multiple instances. Otherwise we have to implement the plugin option merging as well for this I think.

jxnblk:
I'm certainly unsure about this as-is too, which is why I wanted some eyes on it :)

I think the intent is to replicate the same functionality that you could do by manually adding a ThemeProvider in `wrapRootElement`, but as a plugin and with an implementation that is hidden away, which could mean we change the implementation to best suite our needs without causing breaking changes. Currently nested ThemeProviders deep merge the `theme` objects, so the behavior should be similar to how the theme context is added in e.g. `gatsby-theme-notes`. Part of the idea, too, is to make it easier for end-users to change the overall `theme` object for an entire site

> It seems like you could get away with no config in the plugin, no writing to .cache, and instead passing store.getState().plugins to the webpack loader.

I didn't get what you meant before, but this makes a ton of sense, and I think it's better than writing to cache 👍

gatsby-plugin-theme-ui/README.md
jxnblk: Renamed to match the prop name `initialColorMode` – if a color mode object isn't defined in the `theme.colors`, it will default to whatever is in `theme.colors` – a Gatsby theme author could enable this option and an end-user could opt-out of using color modes by not defining them in the theme object, and I guess it could make sense to allow the final instance of this plugin (e.g. the end users site) to disable the color mode state entirely

ChristopherBiscardi:
🎉

PR #41, opened 2019-06-01
Added Bold & Italic to theme generated components

sanandnarayan:
Added Bold & Italic to theme generated components so that one is able to style (add new fonts) to bold and italics specifically

### DIFF
--- a/packages/theme-ui/src/components.js
+++ b/packages/theme-ui/src/components.js
@@ -4,6 +4,8 @@ import themed from './themed'
 
 const tags = [
   'p',
+  'b',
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! The current list is based on https://mdxjs.com/getting-started#table-of-components, but I don't see any harm in adding a few more, especially ones that I think people might expect to exist

sanandnarayan:
oops my bad, may be i should add it to aliases then

sanandnarayan:
opened another PR #42 , if you think it helps.

PR #42, opened 2019-06-03
Converted bold and italics to aliases

sanandnarayan:
1. Removed bold and italics from individual components array 
2. Created aliases for bold and italics with `strong` and `em` respectively
3. Removed the extra `em` in components array

### DIFF
--- a/packages/theme-ui/src/components.js
+++ b/packages/theme-ui/src/components.js
@@ -4,8 +4,6 @@ import themed from './themed'
 
 const tags = [
   'p',
-  'b',
-  'i',
   'a',
   'h1',
   'h2',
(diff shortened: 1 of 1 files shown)


jxnblk:
Not sure what's best for this use-case TBH. The `aliases` are mostly for non-HTML names used in MDX so that for e.g. `theme.styles.inlineCode` applies to `<code>` (and avoids making a non-HTML element). With `<b>` and `<i>`, I could see how it'd make sense for them to use the same styles as `<strong>` and `<em>` but they are technically different HTML elements. Is there a particular use-case you have in mind for these changes? Might help shed some light on how to handle this

sanandnarayan:
I opened this PR as i felt people generally use bold in place of strong. So I felt the code would be cleaner to have them as aliases (in terms of how people use them). When i think of aliases as aliases for HTML elements the existing code makes sense.  Thank you for taking the time to explain

jxnblk:
Yeah no worries! With your other PR, you should be able to use `<Styled.b>` in addition to `<Styled.strong>`, which I think makes sense for the API

PR #65, opened 2019-06-10
Use deepmerge instead of lodash.merge

jxnblk:
This is intended to help reduce the core bundle size

**Note:** this includes a breaking change in the layout components, removing support for styled-system `flexbox` and `layout` props. The `css` prop should be used instead


jxnblk:
I'm planning on merging this and other PRs with breaking changes into the `next-02` branch and creating a single working PR for `v0.2.0`

packages/theme-ui/src/merge.js
```diff
@@ -0,0 +1,21 @@
+import deepmerge from 'deepmerge'
+
+const canUseSymbol = typeof Symbol === 'function' && Symbol.for
+
+const REACT_ELEMENT = canUseSymbol ? Symbol.for('react.element') : 0xeac7
+const FORWARD_REF = canUseSymbol ? Symbol.for('react.forward_ref') : 0xeac7
```
jxnblk: This could potentially be something to add upstream in `deepmerge`

PR #66, opened 2019-06-11
Change Main component to use HTML main element

amberleyromo:
Should be a `<main>` element


jxnblk:
Thanks! Yeah, I think *most* of the time it probably makes sense to render a `<main>` and edge cases can be handled with `as='div'`

amberleyromo:
Yeah, I think it would be very unexpected for a "Main" component to be a `<div>` instead of a `<main>`. At least it was for me!

PR #83, opened 2019-06-18
Add hooks to documentation

jordanoverbye:
Hello 👋 Thank you for building this awesome library - I have really been enjoying it so far and look forward to building more things with it in the future. 

One thing that came recently was that I wanted to access the theme object directly... and after poking around the source code I found the `useThemeUI` hook which worked perfectly. I noticed there wasn't any documentation on this so I have created a PR. I've written a little bit on info, but I'm sure you might want to tweak it 👍

### DIFF
--- a/packages/docs/src/pages/hooks.mdx
+++ b/packages/docs/src/pages/hooks.mdx
@@ -15,8 +15,9 @@ import React from 'react'
 import { useThemeUI } from 'theme-ui'
 
 export default props => {
-  const { theme } = useTheme
-  ...
+  const { theme } = useThemeUI()
+
+  return <pre>{JSON.stringify(theme, null, 2)}</pre>
 }
 ```
 
--- a/packages/docs/src/pages/hooks.mdx
+++ b/packages/docs/src/pages/hooks.mdx
@@ -1,0 +1,22 @@
+import Layout from '../components/layout'
+
+export default Layout
+
+# Hooks
+
+If you are interested in learning more about how React Hooks work please see the official React docs [here](https://reactjs.org/docs/hooks-intro.html). 
+
+## `useThemeUI`
+
+To access the `theme` object directly in a component, you can use the `useThemeUI` hook.
+
+```jsx
+import React from 'react'
+import { useThemeUI } from 'theme-ui'
+
+export default props => {
+  const { theme } = useTheme
+  ...
+}
+```
+
--- a/packages/theme-ui/README.md
+++ b/packages/theme-ui/README.md
@@ -229,8 +229,9 @@ import React from 'react'
 import { useThemeUI } from 'theme-ui'
 
 export default props => {
-  const { theme } = useTheme
-  ...
+  const { theme } = useThemeUI()
+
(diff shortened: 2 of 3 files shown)


packages/docs/src/pages/hooks.mdx
```diff
@@ -0,0 +1,22 @@
...
+## `useThemeUI`
+
+To access the `theme` object directly in a component, you can use the `useThemeUI` hook.
+
+```jsx
+import React from 'react'
+import { useThemeUI } from 'theme-ui'
+
+export default props => {
+  const { theme } = useTheme
```
johno:
```suggestion
  const { theme } = useThemeUI()
```

packages/theme-ui/README.md
```diff
@@ -216,6 +216,24 @@ To change the underlying component in `Styled`, use the `as` prop.
...
+### `useThemeUI`
+
+To access the `theme` object directly in a component, you can use the `useThemeUI` hook.
+
+```jsx
+import React from 'react'
+import { useThemeUI } from 'theme-ui'
+
+export default props => {
+  const { theme } = useTheme
```
johno:
```suggestion
  const { theme } = useThemeUI()
```

packages/theme-ui/README.md
```diff
@@ -216,6 +216,24 @@ To change the underlying component in `Styled`, use the `as` prop.
...
+  ...
```
johno:
```suggestion
  // ...
```

packages/docs/src/pages/hooks.mdx
```diff
@@ -0,0 +1,22 @@
...
+  ...
```
johno: We should probably make this a comment so that it works when cut and pasted.

```suggestion
  // ...
```

Perhaps we could even display the theme by default and make the component:

```js
export default () => {
  const { theme } = useThemUI()
  
  return <pre>{JSON.stringify(theme, null, 2)}</pre>
}
```

jxnblk:
Looks great! Thanks!

jordanoverbye:
Thanks for those suggestions @johno! I've just fixed all of those issues up.

packages/theme-ui/README.md
```diff
@@ -216,6 +216,25 @@ To change the underlying component in `Styled`, use the `as` prop.
 <Styled.a as={Link} to='/'>Home</Styled.a>
 ```
 
+## Hooks
```
jxnblk: I didn't notice this earlier, but I'm not sure if it makes sense to introduce hooks in the readme since it's meant to be sort of a high-level intro and hooks are a more advanced feature – what do you think @johno?

johno: I'd def agree there 👍

jordanoverbye: I've just removed this 👍

jordanoverbye:
Hey @jxnblk I've just removed the hooks section from the readme 👍

jxnblk:
Awesome, thank you!

PR #99, opened 2019-06-19
[WIP] Minor website tweaks

jordanoverbye:
Hello 👋 I've noticed a few really small things on the website that could be fixed up. What this PR adds/fixes:

- Small grid restructure
  -  On the homepage currently there is some a horizontal scrollbar. This is due to the pagination component having negative margins. This has been fixed.
  - I noticed a few things don't line up in a grid which would be nice. For example, the sidebar and footer are not left aligned. Also the second column in the home page banner is not aligned to the second column in the tiles just below.
- Document title page now gets updated on every route. To do this I have added frontmatter to each `.mdx` file inside `./pages/` directory and restructured the layout/root component so the document title can be updated (and other meta in the future).


jxnblk:
Thanks, the landing page definitely needs some work! 😅  I haven't looked too closely at the code yet, but as a heads up, the mobile nav layout doesn't left align in the deploy preview – the nav itself might have some issues, but it aligns differently in the docs pages

Screen Shot 2019-06-19 at 11 06 26 AM
Screen Shot 2019-06-19 at 11 06 48 AM

packages/docs/src/components/banner.js
```diff
@@ -16,14 +11,14 @@ export default props =>
...
-          mb: 4,
+          mb: 4
         },
         p: {
           fontWeight: 'bold',
           mt: 0,
           mb: 3,
           ':first-of-type': {
-            fontSize: [2, 3, 4],
+            fontSize: [2, 3, 4]
```
jxnblk: NB: We can enable hanging commas in the prettier setup to avoid diffs like this

packages/docs/src/components/banner.js
```diff
@@ -39,34 +34,45 @@ export default props =>
...
-          size='1em'
-          color='currentcolor'
+          py: [5, 6]
+        }}
+      >
+        <Box
           css={{
-            fontSize: [ 96, 96, 160 ],
+            display: 'grid',
+            gridTemplateColumns: ['1f', 'repeat(3, 1fr)'],
```
jxnblk: I feel like this whole banner layout could be improved up wrt how the logo is placed, etc. so feel free to be creative with the layout inside the banner

packages/docs/src/components/layout.js
```diff
@@ -6,30 +6,24 @@ import {
...
-  'deep',
-  'swiss',
-]
+const modes = ['light', 'dark', 'deep', 'swiss']
 
 export default props => {
-  const [ menuOpen, setMenuOpen ] = useState(false)
-  const [ mode, setMode ] = useColorMode()
+  const [menuOpen, setMenuOpen] = useState(false)
+  const [mode, setMode] = useColorMode()
```
jxnblk: This hook could go into the Header component now, couldn't it?

packages/docs/src/components/layout.js
```diff
@@ -38,70 +32,82 @@ export default props => {
...
+  const title = [
+    props._frontmatter ? props._frontmatter.title : false,
+    'Theme UI'
+  ]
+    .filter(Boolean)
+    .join(' — ')
+
   return (
     <Styled.root>
+      <Helmet>
```
jxnblk: Just conceptually thinking here, but I find it nice to keep the head component in `wrapPageElement`

packages/docs/src/components/nav-link.js
```diff
@@ -1,33 +1,26 @@
...
   const to = props.to || href
-  return (
-    <Link
-      {...props}
-      to={to}
-      css={styles}
-      activeClassName='active'
-    />
-  )
+  return <Link {...props} to={to} css={styles} activeClassName="active" />
```
jxnblk: Is there a prettier option to always break props to separate lines yet?

packages/docs/src/components/root.js
```diff
@@ -1,21 +1,8 @@
...
-      <meta name='twitter:card' content='summary' />
-      <meta name='twitter:site' content='@jxnblk' />
-      <meta name='twitter:image' content='https://theme-ui.com/icon.png' />
-      <meta name='twitter:title' content='Theme UI' />
-      <meta name='twitter:description' content={pkg.description} />
-    </Helmet>
-    {props.children}
-  </ThemeProvider>
+export default props => (
+  <ThemeProvider components={components}>{props.children}</ThemeProvider>
```
jxnblk: NB: This can be handled with `gatsby-theme-ui` now, so this will probably go away soon

jxnblk:
This is looking great! Thanks for helping out with this. Most of my comments aren't meant to block, but if you can look into making sure the mobile sidebar nave aligns the same across pages and that the nav link focus outlines aren't cropped off, then I think we can ship it

jordanoverbye:
Thanks for the feedback @jxnblk! Sorry about those super obvious ones, it was a bit late. Let me fix those issues up over the next day or so and get back to you.

jxnblk:
No worries, and personally I don't think those were super obvious :)

packages/docs/src/components/banner.js
jordanoverbye: Prettier is set up now 👍

packages/docs/src/components/nav-link.js
jordanoverbye: I don't think so :(. If there is, we add it to the config and apply to all files

packages/docs/src/components/layout.js
jordanoverbye: Yes it can 👍

packages/docs/src/components/layout.js
jordanoverbye: I don't think we can do that because `wrapPageElement` doesn't receive the props we need (`_frontmatter`). I've created a `Head` component to keep the `layout` component cleaner. What do you think?

PR #100, opened 2019-06-19
Formatting/prettier

jordanoverbye:
It would be great to use `prettier` so that all existing and future code is formatted the same. I've added a `format` script in the root dir so that all files can be formatted from a single command which I have included in this PR.


okonet:
Please add 🚫💩 lint-staged. You can do this by running `npx mrm lint-staged`. It will install everything and add the required config.

jxnblk:
Thanks! This would be great to have prettier work with `lint-staged` (like @okonet mentioned) and run as a precommit hook (see this option: https://prettier.io/docs/en/precommit.html#option-1-lint-staged-https-githubcom-okonet-lint-staged)

jordanoverbye:
Thanks @okonet and @jxnblk for your feedback! Would you mind reviewing this again please?

jordanoverbye:
I've just rebased with master to fix conflicts with `yarn.lock`

jxnblk:
Looks great, thanks!

PR #106, opened 2019-06-19
ThemeProvider refactor

### LINKED ISSUE #89: Splitting out root ThemeProvider and nested ThemeProviders
Currently the ThemeProvider component can be nested to create theming contexts inside of a parent theme. This allows components like the experimental [MDX Blocks](https://mdx-blocks.netlify.com/) to work and is a powerful feature that probably isn't leveraged much yet.

That said, there are a few issues with this in its current implementation:

- Color mode state is only kept at the root-level ThemeProvider
- ThemeProvider deeply merges theme objects, which can easily cause styling bugs and unreadable text when combined with things like syntax highlighting in `pre` tags
- The `gatsby-theme-ui` package is geared towards having one, single root theme
- The Chrome extension is also geared towards having a single root theme provider

This is a proposal to split the functionality of the ThemeProvider into two separate components.

1. The top-level, root ThemeProvider:
  - Handles the base theme object
  - Provides the styled MDX components in context
  - Includes hooks for stateful theme object changes which could be leveraged in the Chrome extension and other editing tools
  - Would not pick up outer context
  - Would warn when nested in another ThemeProvider
2. A nested context provider (name TBD):
  - Only includes a nested emotion context (no MDX)
  - Does not handle color modes
  - Can deep merge the base `theme` object (ignoring `theme.styles`)
  - Only shallow merges `theme.styles`

With the two separate components, an application tree could look something like the following:

```jsx
<ThemeProvider theme={theme}>
  <div>root-level context available</div>
  <NestedTheme theme={subTheme}>
    <div>nested context available</div>
  </NestedTheme>
</ThemeProvider>
```

jxnblk:
Fixes #89 

This is an attempt to refactor the ThemeProvider to:

- Only hold color mode state in the root-level ThemeProvider
- Set up the ThemeProvider to have a root-level stateful `theme` object for use with editing tools
- Simplify the merging of nested themes
- Potentially change how MDX components are provided in nested ThemeProviders


packages/docs/src/components/root.js
```diff
@@ -1,11 +1,9 @@
 import React from 'react'
 import { Helmet } from 'react-helmet'
-import { ThemeProvider } from 'theme-ui'
 import pkg from 'theme-ui/package.json'
-import components from './mdx-components'
 
 export default props =>
-  <ThemeProvider components={components}>
+  <>
```
jxnblk: MDX components are handled with `gatsby-theme-ui` now

packages/theme-ui/src/core.js
```diff
@@ -1,52 +1,82 @@
...
   components,
-  ...props
+  children,
 }) => {
+  // components are provided in the default Context
   const outer = useThemeUI()
-  const initialColorMode = outer.colorMode || (theme ? theme.initialColorMode : undefined)
-  const [ colorMode, setColorMode ] = useColorState(initialColorMode)
-  const context = merge({}, {
+  if (outer.colorMode) {
```
jxnblk: ✂️ This is unreachable code

packages/theme-ui/src/core.js
```diff
@@ -1,52 +1,82 @@
...
+    throw new Error(`
+      [ThemeProvider]: only one ThemeProvider component can be used at a time.
+    `)
+  }
+
+  const [ colorMode, setColorMode ] = useColorState(theme.initialColorMode)
+
+  const context = {
+    ...outer,
```
jxnblk:
```suggestion
```

packages/theme-ui/src/core.js
```diff
@@ -1,52 +1,82 @@
...
           value: context,
-          children: props.children
+          children,
         })
       )
     )
   )
 }
+
+// todo: should this provide components too?
```
jxnblk: It *might* make more sense to require using the official `MDXProvider` here instead, though that could make applying `theme.styles` a little trickier. Custom MDX components could be created with the Theme UI `Styled` components or add hooks for `theme.styles` using the `css` prop, which is more work...

I might keep the functionality but conditionally render the `MDXProvider` only *if* the `components` prop is truthy

PR #113, opened 2019-06-21
v0.2.0

### LINKED ISSUE #90: Breaking changes for v0.2
Based on a few issues and discussion, it might be worth bumping to v0.2 with some small breaking changes. Currently, I'm thinking the following:

- Rename the `css` prop in the JSX pragma to not collide with Emotion (#69)
- Remove some (or all) styled-system props in favor of using the `css` prop (#65)
- Split the `ThemeProvider` into two separate components (#89)
- Consider ditching `@emotion/styled` as a dependency and creating all built-in components with the `css` prop, which would help reduce the core bundle size (This *could* cause issues when certain props get accidentally passed through to MDX elements, but docs could be added to show some common pitfalls)
- Remove the `toStyle` API from `theme-ui-typography` in favor of using `toTheme` – this should reduce the overall bundle size
- Rename subpackages to use `@theme-ui` scope

jxnblk:
This PR includes a few breaking changes and is intended for a `0.2.0` release.

Fixes #90 

- Uses `deepmerge` instead of `lodash.merge` to reduce bundle size
- Uses Styled System v5 to reduce bundle size
- Removes layout and flexbox style props from `Box` and other layout components – using the `css` prop is the preferred approach for these now
- [x] Rename the prop used in the custom JSX pragma to `sx` (#69)
- [x] Refactor `ThemeProvider` (#106)
- [x] Remove the `toStyle` API from the typography.js package to reduce bundle size (#114)
- [x] Use the `@theme-ui` scope for `theme-ui-typography` (#114)
- [x] Look into removing `@emotion/styled` dependency (#116)
- [x] Consider removing style props and only using `css` or `sx` props (#118)
- [x] Changelog
- [x] Update color mode to `useEffect`
- [x] Update docs

Currently published as `theme-ui@next`


johno:
Should we include a TODO item for a quick migration guide from v0.1 to v0.2 that's linked in the changelog?

packages/docs/gatsby-node.js
```diff
@@ -0,0 +1,15 @@
+module.exports.createPages = ({ actions }) => {
+  actions.createRedirect({
+    fromPath: '/custom-pragma',
+    isPermanent: true,
+    redirectInBrowser: true,
+    toPath: '/sx-prop',
+  })
+
+  actions.createRedirect({
+    fromPath: '/css',
```
jxnblk: Not sure if we wanted to rename `/static-css` to replace this or just leave as-is

johno: It's early enough I'd say we can rework URLs without needing redirects, but no real opinion either way.

packages/theme-ui/src/provider.js
```diff
@@ -0,0 +1,67 @@
+import { ThemeContext as EmotionContext } from '@emotion/core'
```
jxnblk: This file was renamed, previously `core.js` – must've missed the `git mv` or something

packages/theme-ui/src/provider.js
```diff
@@ -0,0 +1,67 @@
...
+  return jsx(BaseProvider, {
+    context,
+    components: createComponents(components),
+    children,
+  })
+}
+
+export const ThemeProvider = props => {
+  const outer = useThemeUI()
+  if (outer.colorMode) {
```
johno: Dig using the existence of outer context to use the NestedProvider to avoid more API surface area. Though I wonder if we should add some sort of "private" key like `__ThemeUI__` to context to avoid overloading the `colorMode` key? 🤷‍♂

packages/theme-ui/src/merge.js
```diff
@@ -0,0 +1,24 @@
+import deepmerge from 'deepmerge'
+
+const canUseSymbol = typeof Symbol === 'function' && Symbol.for
```
johno: Is this legacy browser support?

jxnblk: Yeah, it's derived from https://github.com/TehShrike/is-mergeable-object/blob/master/index.js which is the default for `deepmerge`. Since this is to support `React.forwardRef`, it might make sense to PR this upstream

packages/theme-ui/src/provider.js
jxnblk: Yeah, that's smart 🤓

packages/docs/gatsby-node.js
jxnblk: Yeah, I'm not too worried about redirects – could rename the `/static-css` page later if needed

jxnblk:
@johno added a `/migrating` page to the docs (https://deploy-preview-113--theme-ui.netlify.com/migrating/) – not linked in the nav, but can be linked to from other places if needed

PR #119, opened 2019-06-22
Add new layout component: Grid

jordanoverbye:
I've added a new layout component called `Grid` which acts similar to the `Box` and `Flexbox` component but obviously it uses CSS grid. I think this would be handy to a lot of people who are using CSS Grid 👍

### DIFF
--- a/packages/docs/src/pages/layout.mdx
+++ b/packages/docs/src/pages/layout.mdx
@@ -49,6 +49,23 @@ export default props => (
 )
 ```
 
+## Grid
+
+The `Grid` layout component is similar to the `Flex` component but uses CSS Grid instead of flexbox.
+
+```jsx
+import React from 'react'
+import { Box, Grid } from 'theme-ui'
+
+export default props => (
+  <Grid p="2" gridTemplateColumns="repeat(3, 1fr)" gridGap="2">
+    <Box />
+    <Box />
+    <Box />
+  </Grid>
+)
+```
+
 ## Layout Theming
 
 These page layout components can be styled with the `theme.styles` object, similar to how other MDX components can be styled.
--- a/packages/theme-ui/src/index.js
+++ b/packages/theme-ui/src/index.js
@@ -4,7 +4,16 @@ export { ThemeProvider } from './core'
 export { Context, useThemeUI } from './context'
 export { ColorMode, useColorMode } from './color-modes'
 export { Styled, components } from './components'
-export { Box, Flex, Layout, Header, Main, Container, Footer } from './layout'
+export {
+  Box,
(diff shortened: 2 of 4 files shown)


packages/docs/src/pages/layout.mdx
```diff
@@ -49,6 +49,23 @@ export default props => (
...
+## Grid
+
+The `Grid` layout component is similar to the `Flex` component but uses CSS Grid instead of flexbox.
+
+```jsx
+import React from 'react'
+import { Box, Grid } from 'theme-ui'
+
+export default props => (
+  <Grid p="2" gridTemplateColumns="repeat(3, 1fr)" gridGap="2">
```
jxnblk: Thanks! While I do like CSS grid layout, I find it difficult to abstract into a nice component API. Even Flex is a bit of a stretch in this regard, but the smaller surface area of flexbox layout's API makes it work a little better.

We're considering ditching the style props (and `@emotion/styled`) altogether in favor of using the `sx` (currently `css`) prop, which would allow you to add things properties like `gridTemplateColumns` on any element, which I think works a little nicer for CSS grid layout. See #113 for those upcoming changes.

I suspect there are some other, probably higher-level layout components that we'd like to add to the library, and I suspect some of those could maybe take advantage of CSS grid layout, if the browser support includes IE 11.

jordanoverbye: @jxnblk okay no problem! Feel free to close this PR if this isn't the right thing for this package :). 

So with the new API you mentioned the above example could be done something like this, right?

```
<div sx={{
  display: 'grid',
  gridTemplateColumns: 'repeat(3, 1fr)',
  gridGap: 2,
  p: 2,
  }} 
/>
```

In my opinion, I actually find my example a little easier to read, just because I instantly know that the element is CSS Grid element. But that's just a personal opinion. Would you be interested in this API if you could use shorthand properties?

```
<Grid p="2" templateColumns="repeat(3, 1fr)" gap="2">
```

jxnblk: Yeah, I think we're leaning against style props altogether. This *could* be a component that's used like `<Grid sx={{ ...styles }} />` but even with the current `Flex` component, I'm unsure if that's all that useful TBH – i.e. it adds an additional API to do the same thing.

That said, if there's something that's a slightly higher level of abstraction (e.g. `GridTiles`, `GridRow`, etc.) that encapsulates some of the grid styles, I suspect that could fit in

jordanoverbye: Ok no problem! I think I might close this issue for now and have a think about how this could be abstracted further.

PR #124, opened 2019-06-25
Fix missing text in typography page option

alex-page:
Select is missing option text on:
https://theme-ui.com/typography/

### DIFF
--- a/packages/docs/src/components/typography-layout.js
+++ b/packages/docs/src/components/typography-layout.js
@@ -37,7 +37,9 @@ const ThemeSelect = props => (
       }}
     >
       {themeNames.map(name => (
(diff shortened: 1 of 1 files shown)


johno:
Thanks!

alex-page:
@jxnblk I was using firefox from my experience the `label` attribute can be a bit flakey in some browsers. I just default to old school html.

**Firefox** no labels at all
Screen Shot 2019-06-25 at 3 00 57 PM

**Chrome** no default label
Screen Shot 2019-06-25 at 3 01 12 PM

Code pen containing code.
https://codepen.io/alex-page/pen/vqJdBx

jxnblk:
Thanks! That's good to know that you can always count on firefox 😉

PR #125, opened 2019-06-25
Spike out a random theme page

johno:


packages/style-guide/src/ColorPalette.js
```diff
@@ -15,33 +15,33 @@ export const ColorRow = ({ colors, name, omit = ['modes'], ...props }) => {
           flexWrap: 'wrap',
         }}
       >
-        {Object.keys(colors).map(key => {
-          const color = colors[key]
-          if (!color || omit.includes(key)) return false
-          const id = join(name, key)
-          if (typeof color === 'object') {
+        {Object.entries(colors)
+          .sort(([_, colorA]) => (typeof colorA === 'string' ? -1 : 1))
```
jxnblk: Is this alphabetic sort? I sort of liked using `Object.keys` here

jxnblk: Nevermind, this is just moving non-objects first, right?

jxnblk:
:shipit:

packages/style-guide/src/ColorPalette.js
johno: Yeah, bumping up the direct values from scales/objects.

PR #148, opened 2019-07-02
Introduce code block component

johno:
In order to steer folks towards using composition and the "MDX Way" for
custom components in place of plugins we should make it easy to get up
and running with syntax highlighting.

This is a basic implementation using react-prism-renderer and the
useThemeUI context hook to extract styles to pass along.

### ~~Things still to consider~~

- ~~Should the title functionality be removed and we document how to
achieve it with a guide instead?~~
- ~~Should the component actually be exported by default? Or should we
require an explicit import for end users (it does come with some bundle
baggage)?~~
- ~~Should this live in `gatsby-plugin-theme-ui`? Should this live elsewhere like `@theme-ui/code-block` so the plugin itself is less opinionated?~~
---

Syntax highlighting on docs site after the change:

image


jxnblk:
> Should the title functionality be removed and we document how to
achieve it with a guide instead?

Haven't looked at code yet, but if it's already in there, it seems like a nice-to-have feature

> Should this live in gatsby-plugin-theme-ui? Should this live elsewhere like @theme-ui/code-block so the plugin itself is less opinionated?

Personally I think a separate package seems better since I'd guess that some people wouldn't use this at all depending on the type of site it is

packages/docs/src/components/prism.js
```diff
@@ -0,0 +1,21 @@
...
+  plain: {
+    color: 'black',
+  },
+  styles: [
+    {
+      types: ['comment', 'prolog', 'doctype', 'cdata'],
+      style: {
+        color: '#777',
+        fontStyle: 'italic',
+      },
```
jxnblk: Since this is based on the *props* in the `prism-react-renderer` component, I wonder if we should stick to using styles provided by Emotion – I might be misreading where this is imported and used

packages/docs/src/gatsby-plugin-theme-ui/index.js
```diff
@@ -127,6 +129,7 @@ export default {
         color: 'inherit',
       },
     },
+    prism,
```
jxnblk: This looks odd to me, I'd rather not mix in component-specific props with the `theme.styles` object

packages/gatsby-plugin-theme-ui/src/code-block.js
```diff
@@ -0,0 +1,40 @@
...
+    <Fragment>
+      {title ? <div sx={codeTitle}>{title}</div> : null}
+      <Highlight
+        {...defaultProps}
+        code={children.trim()}
+        language={language}
+        theme={prismTheme}
+      >
+        {({ className, style, tokens, getLineProps, getTokenProps }) => (
+          <pre className={className} sx={{ ...style, ...pre, ...extraStyles }}>
```
jxnblk: I think this should combine the outer `className` (maybe it is) so that theme-ui styles still apply

packages/gatsby-plugin-theme-ui/src/code-block.js
```diff
@@ -0,0 +1,40 @@
...
+            {tokens.map((line, i) => (
+              <div {...getLineProps({ line, key: i })}>
+                {line.map((token, key) => (
+                  <span {...getTokenProps({ token, key })} />
```
jxnblk: NB: I still don't like props getters

packages/gatsby-plugin-theme-ui/src/provider.js
```diff
@@ -3,7 +3,7 @@ import { jsx, ThemeProvider, ColorMode } from 'theme-ui'
 import theme from './index'
 import components from './components'
 
-export const wrapRootElement = ({ element }, opts) =>
+export const wrapRootElement = ({ element }, _opts) =>
```
jxnblk: I think this was probably a remnant from the other plugin approach or the color mode stuff

```suggestion
export const wrapRootElement = ({ element }) =>
```

jxnblk:
Also looks like you fell into my code coverage trap 😏

packages/gatsby-plugin-theme-ui/src/components.js
```diff
@@ -1,2 +1,7 @@
-// add custom MDX components here
-export default {}
+import React from 'react'
+import PrismCodeBlock from '@theme-ui/prism'
+
+export default {
+  pre: ({ children }) => <>{children}</>,
+  code: PrismCodeBlock
+}
```
jxnblk: Revert this back?

packages/gatsby-plugin-theme-ui/components.js
```diff
@@ -0,0 +1 @@
+export { default } from './src/components'
```
jxnblk: I think this file can ✂️

packages/docs/src/gatsby-plugin-theme-ui/components.js
```diff
@@ -22,6 +24,8 @@ const heading = Tag => props =>
   )
 
 export default {
+  code: CodeBlock,
```
jxnblk: This part looks good

packages/docs/src/components/prism.js
```diff
@@ -0,0 +1,21 @@
...
+    },
+  ],
+}
```
jxnblk: I *think* this file can go, right?

packages/prism/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@theme-ui/prism",
+  "version": "0.2.1",
+  "main": "index.js",
+  "author": "John",
```
jxnblk: "John"

packages/prism/README.md
```diff
@@ -0,0 +1,41 @@
...
+  <ThemeProvider theme={theme} components={components}>
+    {children}
+  </ThemeProvider>
+)
+```
+
+Then, all code blocks in MDX documents wrapped by Layout will be
+syntax highlighted.
+
+TODO: Document usage with theme.js
```
jxnblk: Maybe link here for now? https://theme-ui.com/theming/#syntax-highlighting

packages/prism/package.json
johno: ```git commit -m "Turns out I have a last name, too"```

packages/prism/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "license": "MIT",
+  "scripts": {
+    "prepare": "microbundle --no-compress -o dist",
+    "watch": "microbundle watch --no-compress -o dist"
+  },
+  "dependencies": {
+    "prism-react-renderer": "^0.1.7"
+  },
+  "peerDependencies": {
+    "theme-ui": "^0.1.2"
```
jxnblk:
```suggestion
    "theme-ui": "^0.2.0"
```

jxnblk:
:shipit:

PR #157, opened 2019-07-04
Update docs for color modes and adjust media query logic

jxnblk:


packages/docs/src/gatsby-plugin-theme-ui/index.js
```diff
@@ -15,10 +15,6 @@ export default {
     highlight: '#ffffcc',
     gray: '#777',
     purple: '#609',
-    header: {
-      text: 'inherit',
-      background: 'inherit',
-    },
```
jxnblk: This was for the old demo and isn't used anymore

packages/theme-ui/src/color-modes.js
```diff
@@ -23,9 +23,9 @@ export const useColorState = initialMode => {
   useEffect(() => {
     // initialize
     const stored = storage.get()
-    const dark = getMediaQuery()
-    if (dark) setMode('dark')
     document.body.classList.remove('theme-ui-' + stored)
+    const dark = getMediaQuery()
+    if (!stored && dark) return setMode('dark')
```
jxnblk: Only set the mode to dark if a value isn't stored in localStorage

PR #163, opened 2019-07-05
Ensure that the outer class is passed to the pre tag

johno:


packages/prism/src/index.js
```diff
@@ -4,18 +4,17 @@
...
     <Highlight
       {...defaultProps}
       code={children.trim()}
       language={language}
-      theme={undefined}
-    >
+      theme={undefined}>
       {({ className, style, tokens, getLineProps, getTokenProps }) => (
-        <Styled.pre className={className} sx={style}>
+        <Styled.pre className={`${outerClassName} ${className}`} sx={style}>
```
jxnblk: Also looks like this should be an inline style:
```suggestion
        <Styled.pre className={`${outerClassName} ${className}`} style={style}>
```

PR #166, opened 2019-07-06
add support for CSS properties

### LINKED ISSUE #134: Optional CSS custom properties support for color modes
Currently the color mode flash prevention only handles foreground and background colors for the `<body>` element, which means that colors used in other elements will still flash on page load. The styles generated by the [`ColorMode` component](https://github.com/system-ui/theme-ui/blob/master/packages/theme-ui/src/color-modes.js#L51-L72) could optionally include all values as CSS custom properties, and the `ThemeProvider` component could map the values used in context to their relevant custom properties instead of raw color values.

This should be an opt-in feature since custom properties are not supported in IE 11.

souporserious:
Adds support for toggling CSS properties by setting a `includesCSSProperties` key in theme 🎨. I'm not sure if this initial approach is the best since I'm setting a fallback value, but it seems to work nicely when building the theme-ui docs locally. Maybe there's a better way around that I'm just not seeing? I appreciate any feedback 😊.

closes #134

### DIFF
--- a/packages/docs/src/gatsby-plugin-theme-ui/index.js
+++ b/packages/docs/src/gatsby-plugin-theme-ui/index.js
@@ -5,7 +5,7 @@ const heading = {
 }
 
 export default {
-  includeCSSProperties: true,
+  useCustomProperties: true,
   initialColorMode: 'light',
   colors: {
     text: '#000',
--- a/packages/docs/src/gatsby-plugin-theme-ui/index.js
+++ b/packages/docs/src/gatsby-plugin-theme-ui/index.js
@@ -5,6 +5,7 @@ const heading = {
 }
 
 export default {
+  includeCSSProperties: true,
   initialColorMode: 'light',
   colors: {
     text: '#000',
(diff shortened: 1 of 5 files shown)


jxnblk:
Holy wow! This looks great – and it required way less code than I'd imagined it would've! I think the fallback values should be fine the way you have them (maybe I'm missing something?). My only nit is with the option name (of course, and I do apologize) – would `useCustomProperties` make more sense? `CSSProperties` sounds like it could mean any CSS property to me

souporserious:
Yeah! I was stoked how simple it ended up being. Having color modes already in place made it pretty trivial ✨.  No worries! Yeah, I like that name better 😊. I had initially stayed away from `use` so it didn't get confused with hooks, but as long as it's documented I think it should be fine.

souporserious:
I was going to add docs, but not sure where the best spot would be. Do you think this fits in with color modes? I can also update the CHANGELOG just let me know 😁.

jxnblk:
If you have time to update the docs, that'd be awesome! I think the Color Modes page probably makes most sense for now – probably good to include a note about lack of IE11 support - and feel free to add a line to the CHANGELOG, but I can handle that sorta stuff too after merging ✌️

souporserious:
Added some docs 😊

jxnblk:
Awesome, this looks great! Thanks for getting this working!

souporserious:
No problem! Excited to get this in 🙌 thanks!

PR #174, opened 2019-07-10
Style composition - Allow sx prop to accept arrays

phelma:
[Emotion's css prop](https://emotion.sh/docs/composition) can accept an array for style composition.

This adds the functionality to theme-ui's `sx` prop

Styles can be composed by using an array of style objects in the `sx` prop.

e.g.

```jsx
import React from 'react'
import { css } from 'theme-ui'

const base {
  backgroundColor: 'background',
  color: 'primary'
}

const danger = {
  color: 'red'
}

export default props => (
  <div>
    <div sx={base}>This will be the primary color</div>
    <div sx={[danger, base]}>
      This will be also be primary since the base styles
      overwrite the danger styles.
    </div>
    <div sx={[base, danger]}>This will be red</div>
  </div>
)
```


packages/theme-ui/src/jsx.js
```diff
@@ -3,7 +3,8 @@ import css from '@styled-system/css'
 
 const getCSS = props => theme => {
   if (!props.sx && !props.css) return undefined
-  const styles = css(props.sx)(theme)
+  const sx = Array.isArray(props.sx) ? props.sx : [props.sx]
+  const styles = sx.map(s => css(s)(theme))
   const raw = typeof props.css === 'function' ? props.css(theme) : props.css
   return [styles, raw]
```
jxnblk: Should `styles` also be spread into the array here now?

phelma: Not quite sure what you mean, `styles` is passed to emotions `css` prop which accepts an array and does the actual composition

jxnblk: With your change, this would be passing a nested array along to emotion – e.g. `[ [ a, b ], c ]`

phelma: Hi @jxnblk, What value as an `sx` prop would pass a nested array to emotion? Could you add a failing test?

phelma: Hi @jxnblk, I have spread the `styles` in the return array

phelma:
Hi @jxnblk I have updated the PR following your feedback

jxnblk:
Thanks! Was traveling over the weekend, but planning on merging this. As a heads up, #182 looks like it might cause merge conflicts, but I can take care of them

phelma:
hey @jxnblk do we need to do anything more with this PR to get it in?

domtronn:
+👍

lachlanjc:
Any update here? This looks awesome.

jonavila:
@jxnblk Any plans on adding the functionality of this PR? I've been relying on Emotion's css prop for style composition with arrays, but the would like to avoid the extra step of wrapping my style object with the `css` function.

RobStallion:
@jxnblk Is there any news on this PR? Would be a pretty handy feature to have.

atanasster:
@RobStallion - it is my understanding @jxnblk is not currently very active on this project and @hasparus is heading an effort to move things along while Brent is busy - mostly to close the backlog of issues and PRs. 
I didn't know this PR existed and since there were new issues asking for this feature, submitted a very similar PR in #1226 ( we can also try to resolve the conflicts on the PR here) - can you jump in with some comments and let's try to merge this feature merged.

lachlanjc:
Closing with #704 as a successor

PR #175, opened 2019-07-11
(chrome) docs and remove DS_Store

jordanoverbye:
- Update the download path in chrome README
- Remove DS_Store files from git


johno:
Thanks!

PR #182, opened 2019-07-14
JSX Pragma - In getCSS function, don’t always return a function

jordanoverbye:
I needed to use the `Global` component from `@emotion/core` and I kept getting the following warning:  `It looks like you're using the css prop on Global, did you mean to use the styles prop instead?`

So I was chatting with Mitchell Hamilton (core maintainer of emotion) about this issue and he suggested this was the best fix to get rid of that warning.

But this does not convert the styles in that component to theme values, which I think is expected. 

But I guess this leads to the question, would you be interested in exposing a `Global` component that allows values to be passed to the `sx` prop?

### DIFF
--- a/packages/theme-ui/src/jsx.js
+++ b/packages/theme-ui/src/jsx.js
@@ -1,11 +1,13 @@
 import { jsx as emotion } from '@emotion/core'
 import css from '@styled-system/css'
 
-const getCSS = props => theme => {
+const getCSS = props => {
   if (!props.sx && !props.css) return undefined
-  const styles = css(props.sx)(theme)
-  const raw = typeof props.css === 'function' ? props.css(theme) : props.css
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! I'd been meaning to create an issue for this but kept forgetting... I think this fix looks great!

> would you be interested in exposing a Global component that allows values to be passed to the sx prop?

I think generally we want to discourage the use of Global – we don't currently have a good built-in solution for reseting body margin, but other than that, it's really up to the end-user for how they handle global CSS. If someone *did* want to use the `Global` component with the theme, they can use the `css` function/utility

jordanoverbye:
@jxnblk yeah that's exactly what i needed to use the `Global` component for - styling the html/body :D

PR #192, opened 2019-07-21
fix breakpoints in bootstrap preset

bernharduw:
Breakpoints should include CSS units, but in the Bootstrap preset, breakpoints were unit-less, which leads to not-working media queries.

This PR fixes that by changing the breakpoint to px units.


jxnblk:
Thanks!

PR #197, opened 2019-07-22
Forward props to functional components

### LINKED ISSUE #193: Implement a way to add MDX shortcodes to gatsby-plugin-theme-ui
Right now shadowing `components.js` in `gatsby-plugin-theme-ui` [passes all components to `createComponents`](https://github.com/system-ui/theme-ui/blob/7ccfb6888e63ca99a1841a9c68225f9be89d89df/packages/theme-ui/src/provider.js#L77) which doesn't work properly for global shortcodes since props are lost in the factory.

I wonder if we should be able to detect if the given component is already a function and essentially forward it along, or maybe there's another way to add MDX-specific components?

Right now, the workaround is to shadow the layout and use MDXProvider directly, but that feels heavy handed for what _seems_ like it will be a common use case.

jxnblk:
Maybe fixes #193

### DIFF
--- a/packages/theme-ui/src/styled.js
+++ b/packages/theme-ui/src/styled.js
@@ -4,17 +4,21 @@ import isPropValid from '@emotion/is-prop-valid'
 import jsx from './jsx'
 
 export const styled = tag => (...args) => {
+  const shouldForwardProps = typeof tag === 'function'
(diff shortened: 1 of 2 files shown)


johno (approved):
😻 🙏

PR #210, opened 2019-07-23
Add two new presets based on current theme-ui docs, move presets to own packages

johno:


jxnblk:
Mind ignoring these packages in the coverage or adding some snapshots? https://github.com/system-ui/theme-ui/blob/master/package.json#L56

jxnblk:
Also which one is the one from the docs site?

johnletey:
@johno Mind also adding the “deep” theme from the docs (which is my personal favourite) ... I could also open a PR too

johno:
Presets are now ignored in code coverage and the deep preset added.

> Also which one is the one from the docs site?

The light mode isn't in there currently, just dark/deep/swiss.

johnletey:
@jxnblk @johno Was just previewing the Netlify deploy of this PR, and noticed that the "Raw JSON" section code block isn't highlighted:

Shot 2019-07-24 at 20 31 26

Is it supposed to be? If not, the black text looks kind of funny with the background color ... it's barely visible

Thought I'd report 😄!

jxnblk:
:shipit: 

Would be cool to automate generating Gatsby themes based on the presets now that there's a few in here

PR #218, opened 2019-07-24
fix: Unsupported Prism Languages

### LINKED ISSUE #160: Unsupported Prism Languages
`@theme-ui/prism` uses [`prism-react-renderer`](https://github.com/FormidableLabs/prism-react-renderer), [which does not include all languages supported by Prism](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) by default. So if you use one of the missing languages, it simply does not highlight:

```r
# r language
# no syntax highlighting on theme-ui
```

A fix would be to pass `Prism` as a prop on `Highlight` component, [as documented here](https://github.com/FormidableLabs/prism-react-renderer#prism). I think the best approach would be to allow passing `Prism` to [`index.js`](https://github.com/system-ui/theme-ui/blob/master/packages/prism/src/index.js#L7) like this:

```jsx
export default ({ children, className, title, Prism }) => {
  // ...
  
  return (
    <Highlight
	  {...defaultProps}	
      Prism={Prism || defaultProps.Prism}
      ... />
  )
}
```

So that it could be customized like:

```jsx
import CodeBlock from '@theme-ui/prism'
import Prism from 'prismjs/components/prism-core'
import 'prismjs/components/prism-r'

<CodeBlock Prism={Prism} ... />
```

johnletey:
@jxnblk I believe that I have fixed the issue that @chibicode reported! Let me know if there is anything else I need to do 😄

Closes #160

### DIFF
--- a/packages/prism/src/index.js
+++ b/packages/prism/src/index.js
@@ -12,7 +12,7 @@ export default ({ children, className: outerClassName, title, prism }) => {
   return (
     <Highlight
       {...defaultProps}
-      Prism={Prism || prism}
+      Prism={prism || Prism}
       code={children.trim()}
       language={language}
       theme={undefined}>
--- a/packages/prism/src/index.js
+++ b/packages/prism/src/index.js
@@ -4,12 +4,15 @@
 import Highlight, { defaultProps } from 'prism-react-renderer'
 import { jsx, Styled } from 'theme-ui'
 
-export default ({ children, className: outerClassName, title }) => {
+import Prism from 'prismjs/components/prism-core'
+
+export default ({ children, className: outerClassName, title, prism }) => {
   const [language] = outerClassName.replace(/language-/, '').split(' ')
 
   return (
     <Highlight
       {...defaultProps}
+      Prism={Prism || prism}
       code={children.trim()}
       language={language}
       theme={undefined}>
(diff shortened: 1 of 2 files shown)


packages/prism/src/index.js
```diff
@@ -4,12 +4,15 @@
...
-export default ({ children, className: outerClassName, title }) => {
+import Prism from 'prismjs/components/prism-core'
+
+export default ({ children, className: outerClassName, title, prism }) => {
   const [language] = outerClassName.replace(/language-/, '').split(' ')
 
   return (
     <Highlight
       {...defaultProps}
+      Prism={Prism || prism}
```
jxnblk: Thanks for looking into this! This *looks like* it would always evaluate to the imported module, but I don't think that's the intent, is it?

johnletey: No @jxnblk ... the intention is to check to see if the user passed the function a ‘prism’ parameter, and if they didn’t, we’ll use the one that we imported

johno: Right, but I think @jxnblk's point is that it should be `{prism || Prism}` if I understand this correctly?

johno:
```suggestion
      Prism={prism || Prism}
```

johnletey: I think you are right @johno

PR #220, opened 2019-07-24
Add docs around using the sx prop with MDX

### LINKED ISSUE #178: Using the sx prop in MDX
If you want to use the `sx` prop to style a component inside an MDX file you can create a wrapper component and use that instead:

```jsx
/** @jsx jsx */
import { jsx, Box as BaseBox, Flex as BaseFlex } from "theme-ui";

const Box = props => <BaseBox {...props} />;
const Flex = props => <BaseFlex {...props} />;

export { Box, Flex };
```

Is there a better way?

johno:
Closes #178

### DIFF
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -133,5 +133,5 @@ import { jsx } from 'theme-ui'
 export const Div = props => <Div {...props} />
 ```
 
-This will allow you to import and use the Div component
+This will allow you to import and use the `Div` component
 in MDX files with the `sx` prop.
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -117,3 +117,22 @@ export default props => (
   />
 )
 ```
+
+## Using the `sx` Prop in MDX
+
+In order to render your Markdown components, MDX uses its own
+custom pragma. Eventually we'd like to make it more seamless to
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/sx-prop.mdx
```diff
@@ -117,3 +117,22 @@ export default props => (
...
+use `sx` in MDX files, but for now you'll have to move all components
+that accept the prop to their own file.
+
+```js
+// src/my-sx-components.js
+/** @jsx jsx */
+import { jsx, Box as BaseBox, Flex as BaseFlex } from 'theme-ui'
+
+export const Box = props => <BaseBox {...props} />
+export const Flex = props => <BaseFlex {...props} />
```
jxnblk: I feel like this example looks more complex than it needs to to show what's happening – instead of using `Box` and `Flex`, I think we could have something like:

```jsx
// src/mdx-component.js
/** @jsx jsx */
import { jsx } from 'theme-ui'

export const Div = props => <div {...props} />
```

packages/docs/src/pages/sx-prop.mdx
```diff
@@ -117,3 +117,21 @@ export default props => (
...
+
+```js
+// src/my-sx-components.js
+/** @jsx jsx */
+import { jsx } from 'theme-ui'
+
+export const Div = props => <Div {...props} />
+```
+
+This will allow you to import and use the Div component
```
jxnblk:
```suggestion
This will allow you to import and use the `Div` component
```

PR #221, opened 2019-07-25
[@theme-ui/prism] Talk about unsupported Prism languages

johnletey:
@johno @jxnblk This adds info about how do deal with unsupported Prism languages!

Related #160

### DIFF
--- a/packages/prism/README.md
+++ b/packages/prism/README.md
@@ -38,7 +38,7 @@ export default ({ children }) => (
 Then, all code blocks in MDX documents wrapped by Layout will be
 syntax highlighted.
 
-Please note that `@theme-ui/prism` uses [`prism-react-renderer`](https://github.com/FormidableLabs/prism-react-renderer), [which does not include all languages supported by Prism](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) by default. So, if you happen to use one of the missing languages, your code block simply won't show up as highlighted.
+Please note that `@theme-ui/prism` uses [`prism-react-renderer`](https://github.com/FormidableLabs/prism-react-renderer), [which does not include all languages supported in Prism](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) by default. So, if you happen to use one of the missing languages, your code block simply won't show up as highlighted.
 
 For example, to add support for the `R` language, import the language and pass a custom `Prism` instance to the `CodeBlock` component:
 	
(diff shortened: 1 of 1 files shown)


packages/prism/README.md
```diff
@@ -38,5 +38,17 @@ export default ({ children }) => (
 Then, all code blocks in MDX documents wrapped by Layout will be
 syntax highlighted.
 
+Please note that `@theme-ui/prism` uses [`prism-react-renderer`](https://github.com/FormidableLabs/prism-react-renderer), [which does not include all languages supported by Prism](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) by default. So, if you happen to use one of the missing languages, your code block simply won't show up as highlighted.
+
+However, we have created a workaround. Say that you want to add support for the `R` language. Then you can do that as follows:
```
jxnblk: Thanks for this! It'd be good to focus the language on what the user needs to do, rather than making is sound like something *we've* done. Something like this, maybe?

```suggestion
For example, to add support for the `R` language, import the language and pass a custom `Prism` instance to the `CodeBlock` component:
```

johnletey: Love it! Thanks for the suggestion @jxnblk!

packages/prism/README.md
```diff
@@ -38,5 +38,17 @@ export default ({ children }) => (
 Then, all code blocks in MDX documents wrapped by Layout will be
 syntax highlighted.
 
+Please note that `@theme-ui/prism` uses [`prism-react-renderer`](https://github.com/FormidableLabs/prism-react-renderer), [which does not include all languages supported by Prism](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) by default. So, if you happen to use one of the missing languages, your code block simply won't show up as highlighted.
```
johnletey:
```suggestion
Please note that `@theme-ui/prism` uses [`prism-react-renderer`](https://github.com/FormidableLabs/prism-react-renderer), [which does not include all languages supported in Prism](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) by default. So, if you happen to use one of the missing languages, your code block simply won't show up as highlighted.
```

@jxnblk I think that two `by`'s right next to each other sounds weird, what do you think?

johnletey:
@jxnblk Should I also covert the text to a note like in the `sx-prop` page? Just a thought 💭

jxnblk:
For the readme's I wouldn't use the docs-specific `Note` component, so how you have it looks fine to me

PR #222, opened 2019-07-25
Update sidebar.mdx

MunifTanjim:
Add **Layout Components**


jxnblk:
Thanks! We're considering deprecating the current set of layout components in a later version, which is why the docs are no longer linked to. Instead of using these components directly, I'd recommend using the [variants feature](https://theme-ui.com/guides/variants#themeable-layout-components) and setting up your own layout styles. We'll be adding some more docs for how to do this soon

PR #225, opened 2019-07-26
feat: Add support for code titles, language info and copy button

### LINKED ISSUE #212: [@theme-ui/prism] Add support for code titles, language info and copy button
As mentioned in https://github.com/gatsbyjs/gatsby/pull/15834/files, it would be good to have code title and copy button support for theme-ui

johnletey:
- [x] Code titles
- [ ] Language info
- [ ] Copy button

Closes #212
Related https://github.com/gatsbyjs/gatsby/pull/15834

### DIFF
--- a/packages/prism/package.json
+++ b/packages/prism/package.json
@@ -10,6 +10,7 @@
     "watch": "microbundle watch --no-compress -o dist"
   },
   "dependencies": {
+    "gatsby-design-tokens": "^1.0.0",
     "prism-react-renderer": "^0.1.7",
     "prismjs": "^1.17.1"
   },
--- a/packages/prism/package.json
+++ b/packages/prism/package.json
@@ -10,7 +10,6 @@
     "watch": "microbundle watch --no-compress -o dist"
   },
   "dependencies": {
-    "gatsby-design-tokens": "^1.0.0",
     "prism-react-renderer": "^0.1.7",
     "prismjs": "^1.17.1"
   },
--- a/packages/prism/src/copy-to-clipboard.js
+++ b/packages/prism/src/copy-to-clipboard.js
@@ -1,0 +1,30 @@
+const copyToClipboard = str => {
+  const clipboard = window.navigator.clipboard
+  /*
+   * fallback to older browsers (including Safari)
+   * if clipboard API not supported
+   */
+  if (!clipboard || typeof clipboard.writeText !== `function`) {
+    const textarea = document.createElement(`textarea`)
+    textarea.value = str
+    textarea.setAttribute(`readonly`, true)
+    textarea.setAttribute(`contenteditable`, true)
+    textarea.style.position = `absolute`
+    textarea.style.left = `-9999px`
+    document.body.appendChild(textarea)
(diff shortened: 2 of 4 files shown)


packages/prism/package.json
```diff
@@ -10,6 +10,7 @@
     "watch": "microbundle watch --no-compress -o dist"
   },
   "dependencies": {
+    "gatsby-design-tokens": "^1.0.0",
```
jxnblk: Thanks for working on this! I don't think we want to use Gatsby brand colors for something like this. Instead, the Prism component should be themeable like the rest of the library. The `title` prop could be wrapped in a single `<div>` with a [variant](https://theme-ui.com/guides/variants/) to allow styles to be applied in the `theme` object – sorry that we don't have a lot of examples/docs around this stuff yet

johnletey: @jxnblk So are you saying do something like this?

```js
{title && (
  <div sx={{ variant: 'prism.code-title' }}>
      {title}
  </div>
 )}
```

And then have the user set the `prism.code-title` variant to whatever they want?

johnletey:
@jxnblk Also, what are your opinion on https://github.com/system-ui/theme-ui/issues/212#issuecomment-515028674, as we should include all of the language colors built in for the language info

packages/prism/package.json
jxnblk: Yes, something like that, though I'd avoid using `-` in a JS key name based on common conventions

jxnblk:
Personally, I'd rather see a text label for *copy*, but I think we can add a prop to the `Prism` component to allow any element to be passed in

packages/prism/package.json
johnletey: okay @jxnblk ... I'll call it `prism.codeTitle`

johnletey:
@jxnblk and whoever else is following this PR, I have just pushed a commit with code that I believe adds support for a copy button!

I'm not exactly quite if it works (as I have yet to find a way of testing the code locally), so would love feedback!

sonapraneeth-a:
@johnletey , Do you have a screenshot of the rendered output?

johnletey:
@sonapraneeth-a Like I said above, I don't know how to run the package locally ... would love some advice and I'd be happy to try to run it locally!

jxnblk:
@johnletey if you're able to run the docs site locally with `yarn start`, it should render all code examples with the Prism component, see: https://github.com/system-ui/theme-ui/blob/master/packages/docs/src/gatsby-plugin-theme-ui/components.js#L26

This line could change to pass props to the component:

```jsx
// example
code: props => <Prism {...props} copy={true} />,
```

johnletey:
Wow ... I'll have to try this out! Thanks @jxnblk 

Overall though, how does the code look?

jxnblk:
@johnletey are you still interested in working on this? If not, I think we can close this PR out for the time being

johnletey:
@jxnblk Sorry for the late reply. Unfortunately, I don't think I will have much time to work on this in the future, so I am going to go ahead and close this.

PR #232, opened 2019-07-26
@theme-ui/sidenav fix backgroundColor

MunifTanjim:
`bg` doesn't pick up values from `theme.colors`.
`backgroundColor` does.

### DIFF
--- a/packages/sidenav/src/index.js
+++ b/packages/sidenav/src/index.js
@@ -76,7 +76,7 @@ export const Sidenav = React.forwardRef(
               WebkitOverflowScrolling: 'touch',
               transition: 'transform .2s ease-out',
               transform: [open ? 'translateX(0)' : 'translate(-100%)', 'none'],
-              bg: ['background', 'transparent'],
+              backgroundColor: ['background', 'transparent'],
             }}
           />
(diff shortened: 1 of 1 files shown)


packages/sidenav/src/index.js
```diff
@@ -76,7 +76,7 @@ export const Sidenav = React.forwardRef(
               WebkitOverflowScrolling: 'touch',
               transition: 'transform .2s ease-out',
               transform: [open ? 'translateX(0)' : 'translate(-100%)', 'none'],
-              bg: ['background', 'transparent'],
+              backgroundColor: ['background', 'transparent'],
```
jxnblk: Thanks! But this *should* work as-is and appears to be working on https://theme-ui.com – were you running into issues where this wasn't working?

MunifTanjim: For me, `bg: 'background'` was generating `background: background`.

jxnblk: Ah! If you see `background-color: background` being generated, that's usually a case of mis-matched theming context. I'm looking into better warnings when that happens, but if a value is missing from React context, it can cause unexpected issues like this

Issue #128, opened 2019-06-26
Implement Theme UI to CSS variables utility transformer


alex-page:
Given a theme-ui object, create usable css variables for ingestion in a front end framework.

jxnblk:
Hey @alex-page! Have you started any work on this? If not, I might pick this up soon

alex-page:
@jxnblk yes let me share my progress. Just been having a hectic few weeks.

alex-page:
Here is a gist with what I was working on. I have some questions around if this package should be creating units for things like `px` and `rem` or if that should be provided in the theme object.

https://gist.github.com/alex-page/fc52e2aea996782e2ef56981e721be1f#file-themeui-css-variables-js

alex-page:
I opened a pull request with a few things to discuss 👍

alex-page:
This is done 🚀

PR #235, opened 2019-07-28
Implement Theme UI to CSS variables utility transformer

alex-page:
@jxnblk had some time to dig into #128. Let me know if this should be a package in this project or split a part. Some things I would like feedback on or would like to investigate further my self before merging:

1. Is there a possibility of having more then two levels of depth in a theme object? Is there a collection of theme object examples?
1. Do the generated CSS variable names feel right? 
1. Should there be different options for `kebabcase`, `camelcase` etc?
1. I could provide a map for the variable names or try to stay aligned to theme specification? A good example might be mapping plural to singular `--colors-text` -> `--color-text`.
1. Should variable names that are generated from arrays append a numeric value or something else? If yes, should they always start at zero?
1. I added an optional prefix for the variable name. Should themeui provide a default prefix or is no prefix preferred by default? Happy to add a test for an optional prefix.
1. Should the function return a string of CSS or an object that could be applied with JS `element.style.setProperty(key, value);`? I personally don't mind but would love to know how other people want to use this.
1. What happens when a user doesn't add a unit like in spacing below? Should we just output the value based on what is provided, always add a unit or give an option to provide one?

Here is an example of how the following theme object turns into css:
```js
{
  colors: {
    text: '#000',
    background: '#fff',
    primary: '#07c',
    secondary: '#05a',
    accent: '#609',
    muted: '#f6f6f6',
  },
  fonts: {
    body: 'system-ui, sans-serif',
    heading: 'system-ui, sans-serif',
    monospace: 'Menlo, monospace',
  },
  fontWeights: {
    body: 400,
    heading: 700,
    bold: 700,
  },
  lineHeights: [1.5, 1.125],
  space: [0, 2, 3, 4, 5, 6],
  size: ['10em', '20em', '30em', '40em'],
}
```

Example of css variables generated:
```
--colors-text: #000;
--colors-background: #fff;
--colors-primary: #07c;
--colors-secondary: #05a;
--colors-accent: #609;
--colors-muted: #f6f6f6;
--fonts-body: system-ui, sans-serif;
--fonts-heading: system-ui, sans-serif;
--fonts-monospace: Menlo, monospace;
--fontweights-body: 400;
--fontweights-heading: 700;
--fontweights-bold: 700;
--lineheights-0: 1.5;
--lineheights-1: 1.125;
--space-0: 0;
--space-1: 2;
--space-2: 3;
--space-3: 4;
--space-4: 5;
--space-5: 6;
--size-0: 10em;
--size-1: 20em;
--size-2: 30em;
--size-3: 40em;
```


jxnblk:
Awesome, thanks for working on this! I'll try to address some of your questions below, but generally this is looking great!

> Is there a possibility of having more then two levels of depth in a theme object? Is there a collection of theme object examples?

The `theme` object itself isn't intended to include child `theme` objects, but something working above the level of what's currently going on in Theme UI could. I think for the implementations in Theme UI itself, this can generally be ignored for now – if we stick to a flat object convention across the board, then it'll be easier to handle stuff like that in the future with separate components without making the initial code base more complex.

As for individual scales within the theme object, e.g. `colors`, they can be deeply nested, but the naming convention is up to the user.

> Should there be different options for kebabcase, camelcase etc?

CSS custom properties are case sensitive, so using the same convention should be okay. For dot-notation, I think keys could be separated by hyphens, e.g. `--color-blue-2`. An option for converting to kebab-case seems like a nice-to-have though.

> Do the generated CSS variable names feel right?
> I could provide a map for the variable names or try to stay aligned to theme specification? A good example might be mapping plural to singular --colors-text -> --color-text.

Since the theme scales are objects and arrays, the plural naming convention makes sense in the theme object, but for the custom properties, we can probably use something like `pluralize` to use a singular naming convention. I.e. I think they should be named like `--color-background` and `--fontSize-0`

> Should variable names that are generated from arrays append a numeric value or something else? If yes, should they always start at zero?

Yeah, I think using the object keys as-is is the most straight-forward naming convention, and since arrays are zero-based, `fontSizes[0]` would become `--fontSize-0`.

> I added an optional prefix for the variable name. Should themeui provide a default prefix or is no prefix preferred by default? Happy to add a test for an optional prefix.

An optional prefix would be fantastic, but I don't think it's necessary to ship this. The default could be un-prefixed, even though there might be potential for collisions. If you want to add that to the PR, feel free to do so.

> Should the function return a string of CSS or an object that could be applied with JS element.style.setProperty(key, value);? I personally don't mind but would love to know how other people want to use this.

For the initial use-case, I think being able to pass a style object to either the React `style` prop, the Emotion `css` prop, or the `Global` component would cover most use-cases. If someone needs a CSS string version, there are existing utilities that can convert the value, and I don't think we need to support that in this utility. It looks like you've implemented it as a string currently, but hopefully it's not too difficult to switch to an array.

> What happens when a user doesn't add a unit like in spacing below? Should we just output the value based on what is provided, always add a unit or give an option to provide one?

For number values, we'll probably want to convert *some* CSS properties to use pixel units and leave properties like `lineHeight` as unitless values. If the numbers are consumed *only* within Emotion, we wouldn't need this, but I think part of the idea here is to make the theme object consumable in other CSS contexts.

packages/css-variables/README.md
```diff
@@ -0,0 +1,25 @@
+# @theme-ui/css-variables
```
jxnblk: I think the package name could be `@theme-ui/custom-properties` to be a little more explicit about what it's intended for

packages/css-variables/test/test.js
```diff
@@ -0,0 +1,33 @@
...
+import toCssVariables from '../src'
+
+const theme = {
+  colors: {
+    text: '#000',
+    background: '#fff',
+    primary: '#07c',
+    secondary: '#05a',
+    accent: '#609',
+    muted: '#f6f6f6',
```
jxnblk: Don't worry about it for this PR, but support for setting colors modes via a class selector is something I've thought might be nice in a utility like this

packages/css-variables/test/test.js
```diff
@@ -0,0 +1,33 @@
...
+    monospace: 'Menlo, monospace',
+  },
+  fontWeights: {
+    body: 400,
+    heading: 700,
+    bold: 700,
+  },
+  lineHeights: [1.5, 1.125],
+  space: [0, 2, 3, 4, 5, 6],
+  size: ['10em', '20em', '30em', '40em'],
```
jxnblk:
```suggestion
  sizes: ['10em', '20em', '30em', '40em'],
```

alex-page:
Thanks for all the great points. I am out for a few days but will jump back into this later in the week.

packages/css-variables/test/test.js
alex-page: That would be a nice feature. I will create an issue for this after the PR is merged.

alex-page:
I have gone through and fixed up the following based on the feedback:

- [x] Nested properties in theme object
- [x] Remove plurals in theme object
- [x] Test for optional prefix
- [x] Returns object ( I think this will work for `React style prop` but I haven't tested it )

> For number values, we'll probably want to convert some CSS properties to use pixel units and leave properties like lineHeight as unitless values. If the numbers are consumed only within Emotion, we wouldn't need this, but I think part of the idea here is to make the theme object consumable in other CSS contexts.

I feel like there could be a linter to make sure that certain properties have units. `space` for example I think should have a unit. For now the function returns the values based on the input. I would feel a bit dirty appending `px` to the end if there isn't a unit but maybe that is intentional.

jxnblk:
Awesome! Will take another look in a few!

> Not including units may be intentional ( if it is I would love to know why )

With Emotion, unitless number values are converted to pixel values for CSS properties where it makes sense. This is also how the `style` prop in React works, and Styled Components supports the same thing. In Theme UI, you can define values in e.g. `theme.space` as numbers, but they render as pixels. For this to work *outside* of emotion, I think we'll want a way to convert numbers to pixel string values. This is *not* needed in this PR though, and it's something I can look into later. (I suspect `@emotion/core` might have this utility already)

packages/custom-properties/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+  return customProperties;
+}
+```
+
+## Parameters
+
+The @theme-ui/custom-properties function takes two parameters:
+
+feColor( $theme, $prefix );
```
jxnblk: Looks like a typo?

packages/custom-properties/test/test.js
```diff
@@ -0,0 +1,47 @@
+import 'babel-polyfill'
```
jxnblk: Is this needed for something? If so, we should probably fix it in the root-level Jest configuration

packages/custom-properties/test/__snapshots__/test.js.snap
```diff
@@ -0,0 +1,63 @@
...
+  "--🍭-color-dark-background-surface": "#fff",
+  "--🍭-color-dark-foreground-text": "#000",
+  "--🍭-color-muted": "#f6f6f6",
+  "--🍭-color-primary": "#07c",
+  "--🍭-color-secondary": "#05a",
+  "--🍭-color-text": "#000",
+  "--🍭-font-body": "system-ui, sans-serif",
+  "--🍭-font-heading": "system-ui, sans-serif",
+  "--🍭-font-monospace": "Menlo, monospace",
+  "--🍭-font-weight-body": 400,
```
jxnblk: Not sure what the best thing to do here is, and it might be an edge case. With hyphenated theme object keys, `fontWeights`, `fontSizes`, *could* collide with `fonts.size.body` (I doubt people would use this sort of naming convention)

packages/custom-properties/README.md
alex-page: My bad 😂

packages/custom-properties/test/test.js
alex-page: Good question I will see what happens when I remove it. I was copying the `tachyons` package.

alex-page: Removed it without any problems.

alex-page:
> With Emotion, unitless number values are converted to pixel values for CSS properties where it makes sense. This is also how the style prop in React works, and Styled Components supports the same thing. In Theme UI, you can define values in e.g. theme.space as numbers, but they render as pixels. For this to work outside of emotion, I think we'll want a way to convert numbers to pixel string values. This is not needed in this PR though, and it's something I can look into later. (I suspect @emotion/core might have this utility already)

This is super interesting! Thanks for the context 💯

packages/custom-properties/test/__snapshots__/test.js.snap
alex-page: Is the confusion that they might be using `fonts.size.body` then `font-size-body`. I would prefer to keep it consistent despite the fact that it might not be logically correct. It also means one less dependency.

jxnblk: Yeah, I'm suggesting the CSS name could be for e.g. `--fontSize-0` and `--fontWeight-bold` instead of hyphenating the original key. That way `-` is only used in place of `.`

alex-page: Sounds good to me. It also feels more consistent.

alex-page: I have made the change. It now keeps the camelCase key but still removes the plural.

alex-page:
This feels pretty polished now. Let me know if there are any additional changes 👍

jxnblk:
This looks great! Thanks for working on this!

packages/css-variables/test/test.js
alex-page: New issue created https://github.com/system-ui/theme-ui/issues/257

eschaefer:
Hey all, just for awareness, this broke compatibility with IE11, which does not support CSS variables. I love this lib but I imagine quite a few people still need to support this unfortunate browser...

alex-page:
@eschaefer you can use theme-ui in multiple ways this is just one option. You could try using [tachyons](https://github.com/system-ui/theme-ui/tree/master/packages/tachyons) or [tailwind](https://github.com/system-ui/theme-ui/tree/master/packages/tailwind) package if you want to generate css classes which would work in IE11.

jxnblk:
@eschaefer you can turn off custom properties with this flag in your theme: https://theme-ui.com/color-modes#turn-off-custom-properties

eschaefer:
Thanks @jxnblk! This solves the issue perfectly! I was searching for the wrong text ("css variables") in the documentation. 🙄

PR #246, opened 2019-07-29
Rework color mode API documentation

johno:
Color modes can be difficult to grasp, and I think it might be partially
due to the "happy path" requiring configuration. This attempts to rework
the API to _maybe_ be more straightforward.

Please note that this doesn't yet address any implementation concerns,
but is attempting to document the most desirable API before we
implement. This will be able to be done in a backwards-compatible way.

I'm open to any and all suggestions.

---

[Rendered version](https://github.com/system-ui/theme-ui/blob/1056e08fccea355d32bc9fe6f7304ea7606c14e9/packages/docs/src/pages/color-modes.mdx)


jxnblk:
Generally LGTM! 

Couldn't "suggest" this but I think this like should be removed:
> If you support CSS custom properties, we recommend setting `useCustomProperties` in your theme.

Technically, this isn't 100% right:
> This approach isn't required when using gatsby-plugin-theme-ui because after rehydration, the custom property implementation will be replaced with CSS values.

`gatsby-plugin-theme-ui` [renders the ColorMode component](https://github.com/system-ui/theme-ui/blob/master/packages/gatsby-plugin-theme-ui/src/provider.js#L14-L17), but it still needs the `useCustomProperties` flag at the moment. The only additional thing the plugin adds is the render-blocking script in [`gatsby-ssr.js`](https://github.com/system-ui/theme-ui/blob/master/packages/gatsby-plugin-theme-ui/gatsby-ssr.js)

johno:
> gatsby-plugin-theme-ui [...] still needs the useCustomProperties flag at the moment

Though it'd no longer require the flag if we make it the default, right?

jxnblk:
@johno What I mean is, if we change it to use custom properties by default, but allow you to disable them with `useCustomProperties: false`, this shouldn't really have anything to do with what `gatsby-plugin-theme-ui` is doing, and I don't think we need to call it out

jxnblk:
I say let's go full-steam ahead with this! 🚂

jxnblk:
Due to the confusion around the current behavior with the `prefers-color-scheme` media query, I think we should disable that or make it opt-in with this change as well

jxnblk:
:shipit:

PR #259, opened 2019-07-31
Refactor custom properties implementation in color modes

### LINKED ISSUE #224: Enabling "useCustomProperties" breaks dot selector for colors
Can be replicated in https://github.com/KyleAMathews/gatsby-theme-hyde

If you install and run and navigate to http://localhost:8000/hello-world which has some code blocks — all the prism color tokens aren't getting replaced by their actual values but are left as strings e.g. `background-color: prism.background;`

jxnblk:
Fixes #224


packages/theme-ui/src/custom-properties.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+const toPixel = (key, value) => {
+  if (typeof value !== 'number') return value
+  if (numberScales[key]) return value
+  return value + 'px'
+}
+
+// convert theme values to custom properties
+export const toCustomProperties = (obj, parent, themeKey) => {
```
jxnblk: This *should* work for the entire theme object, and could potentially be replaced by the work in `@theme-ui/custom-properties` in the future

packages/theme-ui/src/custom-properties.js
```diff
@@ -0,0 +1,97 @@
...
+      styles[key] = {
+        color: modes[mode].text,
+        bg: modes[mode].background,
+      }
+    } else {
+      styles[key] = objectToVars('colors', modes[mode])
+    }
+  })
+
+  if (!theme.useCustomProperties) {
```
jxnblk: The logic here could be simplified if we only support the FOUC issue in browser that support custom properties, which I suspect is an okay tradeoff

packages/theme-ui/src/provider.js
```diff
@@ -69,7 +37,11 @@ const RootProvider = ({ theme: propsTheme = {}, components, children }) => {
   const outer = useThemeUI()
   const [colorMode, setColorMode] = useColorState(propsTheme.initialColorMode)
   const [themeState, setThemeState] = useReducer(mergeState, propsTheme)
+
   const theme = applyColorMode(themeState, colorMode)
+  if (theme.useCustomProperties) {
+    theme.colors = toCustomProperties(theme.colors, 'colors')
```
jxnblk: Currently this is still only converting the `colors` object to custom properties

packages/theme-ui/src/custom-properties.js
johno: Yeah, I'd say it's fine to have FOUC for legacy browsers since it will fall back to a small flash on initial load but otherwise work fine.

loganpowell:
I don't know if it's just my setup, but I'm still getting FOUC with `"theme-ui": "^0.2.26"` (Chrome)

jxnblk:
@loganpowell currently only sites with `gatsby-plugin-theme-ui` have the full flash prevention setup. We'd like to have an equivalent for Next.js as well, if that's anything you're interested in helping out with

loganpowell:
@jxnblk How can I help? If it's within my abilities, I'm up for it

loganpowell:
I have a couple of questions that might help me get oriented (please forgive me if they seem naive):

1. wouldn't it be easier to set the theme dynamically if the modes lived at the same level within `theme`?
2. if `useThemeUI` returned the current `mode` then couldn't you set the mode as a `prop` parent then :
```jsx
<div
  sx={{
    bg: `${props.mode}.base.primary`
  }}
>
```
I know this would be a breaking change, so feel free to say "out of the question"

loganpowell:
On second thought, that's a stupid idea. The whole point is not to have to `useThemeUI` everywhere

loganpowell:
I think the key to getting the last set `mode` into next would be to do it in `getInitialProps` in `_app.js` as is done with material: https://github.com/mui-org/material-ui/blob/master/examples/nextjs/pages/_document.js

LekoArts:
@jxnblk 	This PR was a breaking change in one of my themes:

```jsx
const { theme, colorMode } = useThemeUI()

const bg = colorMode === `light` ? theme.colors.background : theme.colors.modes.dark.background
const shade = rgba(bg, 0.4)
```

In the past `bg` would be `#fff`, now it's `var(--theme-ui-colors-background, #fff)` which breaks the rgba() function.

How can I get only the `#fff` value?

jxnblk:
@LekoArts sorry about that! Opened #271 to try to figure out the best fix for this

jxnblk:
@loganpowell I've opened #270 for discussion. I know someone else expressed interest in working on this, so I'll try to find that issue

LekoArts:
Thanks! No worries, my Cypress tests catched it — in the meantime I'm importing the file manually

PR #261, opened 2019-08-01
Forward ref in styled HOC

### LINKED ISSUE #260: Can not get ref from the <Box>
don't know what I am missing, I checked the code is indeed `forwardRef`, but in this example, 3 inputs there, when typing in the inputs, except for the last one which uses raw `<input>`, the 2 `Box` from `theme-ui` does not have ref.

https://codesandbox.io/s/fervent-swanson-t8wsu

jxnblk:
Fixes #260

### DIFF
--- a/packages/theme-ui/test/layout.js
+++ b/packages/theme-ui/test/layout.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
+import { render, cleanup } from '@testing-library/react'
 import {
   Box,
   Flex,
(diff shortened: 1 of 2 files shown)


packages/theme-ui/test/layout.js
```diff
@@ -22,3 +25,9 @@ test.each([
   const json = renderer.create(<Component />).toJSON()
   expect(json).toMatchSnapshot()
 })
+
+test('forwards ref to div', () => {
+  const ref = React.createRef(null)
+  const tree = render(<Box ref={ref} />)
+  expect(ref.current.tagName).toBe('DIV')
```
Albert-Gao: Cool, I was thinking about how to test the ref passing when doing a shallow rendering,,, just this easy... assert it directly :D

PR #272, opened 2019-08-06
fix(preset-tailwind): Add missing keys

LekoArts:
Hi!

While using the Tailwind preset I realized that some keys are missing. As this PR also changes the order of some variables (to order it alphabetically – I look into the preset a lot and want to find things quicker), here a list of what I changed (when referring to "original" I mean the [defaultConfig](https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js)):

- Change the `breakpoints` from `rem` to `px` values like in the original (they were correct for a root font size of 16px but e.g. I personally always use 18px or some people use 10px)
- Add `tighter` to `letterSpacings`
- Add a bunch of keys to `sizes`. In the original they are used for things like max-width, width
- Add `snug` and `normal` to `baseLineHeights`

---

I wanted to change the `fontSizes`, too, as they're incorrect (see [original](https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js#L253)) but that would be a breaking change and it's not widely different from the original.

### DIFF
--- a/packages/preset-tailwind/src/index.js
+++ b/packages/preset-tailwind/src/index.js
@@ -1,6 +1,14 @@
 // Based on https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js
 
-export const breakpoints = ['40rem', '48rem', '64rem', '80rem']
+export const borderWidths = {
+  px: '1px',
+  '0': '0',
+  '2': '2px',
+  '4': '4px',
+  '8': '8px',
+}
+
+export const breakpoints = ['640px', '758px', '1024px', '1280px']
 
 export const baseColors = {
   transparent: 'transparent',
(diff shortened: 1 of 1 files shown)


jxnblk:
> I wanted to change the fontSizes

What would you change specifically? There might be a non-breaking way to adjust them if the keys differ.

Also, did you want to throw in some of the button variants you mentioned in this PR or another one?

LekoArts:
Original: https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js#L253

---

So it would be:

```js
export const fontSizes = [
  '0.75rem',
  '0.875rem',
  '1rem',
  '1.125rem',
  '1.25rem',
  '1.5rem',
  '1.875rem',
  '2.25rem',
  '3rem',
  '4rem',
  '4.5rem',
]
```

---

> Also, did you want to throw in some of the button variants you mentioned in this PR or another one?

Will do this in another one :)

jxnblk:
Hmmm... The difference in the font sizes looks like that was a mistake, but yeah technically this would be somewhat a breaking change

LekoArts:
I'm fine with waiting for that when another breaking change needs to be done

PR #277, opened 2019-08-07
examples/next: fix unstyled flash

### LINKED ISSUE #270: Next.js plugin
Add a plugin (or similar) for Next.js to help prevent the flash of unstyled colors. See the implementation in `gatsby-plugin-theme-ui` for reference

jordanoverbye:
Hello 👋

This PR attempts to fix #270 

So I don't think it's possible to create a `theme-ui-next` plugin as we have to do things that plugins aren't able to do (like wrapping the App in a ThemeProvider). So instead I have done something else:

1. Extract the `noflash` function in `gatsby-plugin-theme-ui` into its own function that `theme-ui` exports called `applyColorModeFromLocalStorage`. 
2. Update the next example to use light/dark mode and also use the `applyColorModeFromLocalStorage` which prevents the unstyled flash.

### DIFF
--- a/examples/next/pages/_document.js
+++ b/examples/next/pages/_document.js
@@ -1,0 +1,27 @@
+import React from 'react'
+import Document, { Html, Head, Main, NextScript } from 'next/document'
+import { applyColorModeFromLocalStorage } from 'theme-ui'
+
+class MyDocument extends Document {
+  static async getInitialProps(ctx) {
+    const initialProps = await Document.getInitialProps(ctx)
+    return initialProps
+  }
+
+  render() {
+    return (
+      <Html>
+        <Head />
+        <body>
+          <script
+            dangerouslySetInnerHTML={{ __html: applyColorModeFromLocalStorage }}
+          />
+          <Main />
+          <NextScript />
+        </body>
+      </Html>
+    )
+  }
+}
+
+export default MyDocument
--- a/examples/next/pages/_document.js
+++ b/examples/next/pages/_document.js
@@ -9,7 +9,6 @@ class MyDocument extends Document {
   }
 
   render() {
-    console.log(ApplyColorModeFromLocalStorage)
     return (
       <Html>
         <Head />
--- a/examples/next/pages/_document.js
+++ b/examples/next/pages/_document.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import Document, { Html, Head, Main, NextScript } from 'next/document'
-import { ApplyColorModeFromLocalStorage } from 'theme-ui'
+import { InitializeColorMode } from 'theme-ui'
 
 class MyDocument extends Document {
   static async getInitialProps(ctx) {
@@ -13,7 +13,7 @@ class MyDocument extends Document {
       <Html>
         <Head />
         <body>
-          <ApplyColorModeFromLocalStorage />
+          <InitializeColorMode />
           <Main />
           <NextScript />
         </body>
--- a/examples/next/pages/_document.js
+++ b/examples/next/pages/_document.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import Document, { Html, Head, Main, NextScript } from 'next/document'
-import { applyColorModeFromLocalStorage } from 'theme-ui'
+import { ApplyColorModeFromLocalStorage } from 'theme-ui'
 
 class MyDocument extends Document {
   static async getInitialProps(ctx) {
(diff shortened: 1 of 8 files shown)


examples/next/pages/_document.js
```diff
@@ -0,0 +1,27 @@
...
+    const initialProps = await Document.getInitialProps(ctx)
+    return initialProps
+  }
+
+  render() {
+    return (
+      <Html>
+        <Head />
+        <body>
+          <script
```
jxnblk: Thanks for looking into this! I wonder if it might make more sense from an API standpoint to make this a component, since this and `gatsby-ssr.js` both use a script element

johnletey:
@jxnblk @jordanoverbye When I run your code ... I get a module not found error for `@emotion/core`

Should I install that?

examples/next/pages/_document.js
jordanoverbye: Yeah that's a good idea! I was thinking that after I submitted this PR. The reason I didn't do that is because in `gatsby-ssr.js` it's using a js function and not JSX. But I will update

jordanoverbye:
@johnletey good pickup! I'll fix that.

jordanoverbye:
Hey @jxnblk , do you mind taking another look at this please?

packages/theme-ui/src/color-modes.js
```diff
@@ -68,4 +68,17 @@ const bodyColor = theme => ({
 
 export const ColorMode = () => <Global styles={bodyColor} />
 
+export const ApplyColorModeFromLocalStorage = () => (
+  <script
+    key="theme-ui-noscript"
```
jxnblk: Think I made this typo originally, but... 
```suggestion
    key="theme-ui-no-flash"
```

packages/theme-ui/src/color-modes.js
```diff
@@ -68,4 +68,17 @@ const bodyColor = theme => ({
 
 export const ColorMode = () => <Global styles={bodyColor} />
 
+export const ApplyColorModeFromLocalStorage = () => (
```
jxnblk: As far as names go, what about
```suggestion
export const InitializeColorMode = () => (
```

jxnblk:
This is looking great! Not sure about what would make most sense for the component name, but other than that this seems great! We can add docs for Next.js usage after this is merged in

jordanoverbye:
Thanks @jxnblk , I've just updated those changes

johnletey:
@jordanoverbye @jxnblk When I use the newest update `_document.js`, I get the following error:

```
at Object.renderToHTML (/Users/john/Desktop/learn-anything/node_modules/next-server/dist/server/render.js:277:16) name: 'Invariant Violation', framesToPop: 1 }
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check your code at _document.js:16.
    in MyDocument
    in Context.Provider
```

jordanoverbye:
Hey @johnletey 

So because this example lives outside the `/packages` folder, we have to manually link the latest version of `theme-ui` into `examples/next`. I think the problem is you haven't done this step and because we have added a new API which doesn't exist in the current version of `theme-ui`, you are getting that error.

```
cd ~/path/to/theme-ui/packages/theme-ui
yarn prepare
yarn link
cd ../../examples/next 
yarn link "theme-ui"
yarn dev
```

Let me know if that fixes it for you.

johnletey:
No @jordanoverbye ... I'm literally just coping-n-pasting your code into a different Next.js project that I'm working on and I need Theme UI for, and the newest changes to `_document.js` cause errors when running the site

jordanoverbye:
Hey @johnletey 

Oh right that makes sense - but the error you are seeing is caused by the same thing though. In your NextJS project you have `theme-ui` installed via NPM. The latest version of `theme-ui` on NPM does not export the component `InitializeColorMode` which I have introduced in this PR. So it will fail when it tries to render something that doesn't exist.

What you can do for now is just copy the source code of `InitializeColorMode` and paste it into `_document.js`. Then, once this PR is merged in you can upgrade your `theme-ui` package and remove this hack.

```
import React from 'react'
import Document, { Html, Head, Main, NextScript } from 'next/document'
// import { InitializeColorMode } from 'theme-ui'

// Remove this once this PR is merged
const InitializeColorMode = () => (
  <script
    key="theme-ui-no-flash"
    dangerouslySetInnerHTML={{
      __html: `(function() { try {
        var mode = localStorage.getItem('theme-ui-color-mode');
        if (!mode) return
        document.body.classList.add('theme-ui-' + mode);
      } catch (e) {} })();`,
    }}
  />
)

class MyDocument extends Document {
  static async getInitialProps(ctx) {
    const initialProps = await Document.getInitialProps(ctx)
    return initialProps
  }

  render() {
    return (
      <Html>
        <Head />
        <body>
          <InitializeColorMode />
          <Main />
          <NextScript />
        </body>
      </Html>
    )
  }
}

export default MyDocument
```

johnletey:
Wonderful ... thank you so much @jordanoverbye

Now I don't have an unstyled flash!

jxnblk:
@jordanoverbye Thanks for working on this, this looks great!

PR #282, opened 2019-08-11
docs/pages/theming.mdx: update link to base preset

HerbCaudill:
Each preset now has its own package, but the URL on this link needed to be updated.


jxnblk:
Thanks!

Issue #284, opened 2019-08-12
Double quotes from Typography js theme makes font declaration invalid


RVMendoza:
I'm setting this up on Storybook. I'm trying to apply fonts to a component/story, and the double quotes causing the browser to read it as a string instead of a font declaration.

```
import { configure, addDecorator } from '@storybook/react';
import React from 'react'
import { ThemeProvider } from 'theme-ui'
import Theme from './theme'
import { toTheme } from '@theme-ui/typography'
import moragaTheme from 'typography-theme-moraga'

const theme = { ...toTheme(moragaTheme), ...Theme };
console.log(theme);

// automatically import all files ending in *.stories.js
const req = require.context('../stories', true, /\.stories\.js$/);
function loadStories() {
  req.keys().forEach(filename => req(filename));
}

configure(loadStories, module);

const ThemeDecorator = storyFn => <ThemeProvider theme={theme}>{storyFn()}</ThemeProvider>;
addDecorator(ThemeDecorator);

```
When `console.log` the theme and it outputs this:
<img width="294" alt="Messages Image(404214036)" src="https://user-images.githubusercontent.com/4342363/62841551-90eb9880-bc77-11e9-84f7-fb133072e428.png">

I looked in the styles of the dev tools, and I'm seeing font-family being read as a string rather than an actual font-family declaration.
<img width="308" alt="Messages Image(1056516193)" src="https://user-images.githubusercontent.com/4342363/62841562-b37db180-bc77-11e9-9f57-80c6c6f19241.png">

## Possible source of problem
Perhaps this is something in the `toTheme` component? Something about it is adding double quotes to the font declaration?

## Temporary Workaround
Explicitly set the fonts in your theme file.
```
  fonts: {
    body: "Source Sans Pro, sans-serif",
    heading: "Source Sans Pro, sans-serif",
  },
```

Dev tool style should look correct now:
![image](https://user-images.githubusercontent.com/4342363/62841578-cabc9f00-bc77-11e9-9877-923c82d940ea.png)

dburles:
Looks like a bug! I think it's because the [generic font name is also being wrapped in quotes](https://github.com/system-ui/theme-ui/blob/master/packages/typography/src/to-theme.js#L50). I think the solution might be to [transform them in a similar way to typography.js](https://github.com/KyleAMathews/typography.js/blob/master/packages/typography/src/utils/createStyles.js#L33-L48)

dburles:
[...]

PR #285, opened 2019-08-12
Don't wrap font keywords in quotes

dburles:
for #284

### DIFF
--- a/packages/typography/src/to-theme.js
+++ b/packages/typography/src/to-theme.js
@@ -59,6 +59,7 @@ const genericFontFamilies = [
   'fantasy',
   'cursive',
   '-apple-system',
+  'system-ui',
 ]
 
 const wrapFontFamily = fontFamily =>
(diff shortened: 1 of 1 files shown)


packages/typography/src/to-theme.js
```diff
@@ -47,7 +47,24 @@ export const getSpace = (result, opts) => {
...
+// See https://stackoverflow.com/a/13752149 and https://www.w3.org/TR/CSS2/fonts.html#font-family-prop
+const genericFontFamilies = [
+  'inherit',
+  'default',
+  'serif',
+  'sans-serif',
+  'monospace',
+  'fantasy',
+  'cursive',
+  '-apple-system',
```
jxnblk: Adding this since it's meant to be the more standard name
```suggestion
  '-apple-system',
  'system-ui',
```

dburles:
Updated

jxnblk:
Thanks!

PR #291, opened 2019-08-12
feat(preset-tailwind): Add button/input variants

LekoArts:
As per https://tailwindcss.com/components I'm adding some button/input variants.

Preview:

image

Also changed the `primary` color to something more usable, and added a `primaryHover` key.


jxnblk:
Thanks for adding the button variants! The colors on these look *really* low-contrast to me. Are these the colors that Tailwind uses? https://colorable.jxnblk.com/ffffff/5199dd

LekoArts:
Good point! Yes, it's the same colors. But I guess we can do better :) Therefore I changed both primary colors to make it AA accessible.

image

PR #297, opened 2019-08-14
Add warning for conflicting emotion versions

jxnblk:


ChristopherBiscardi:
Seems like it could work

PR #303, opened 2019-08-15
Correct accidental capitalization of HTML tag

xenophilicibex:
I spent about 10 minutes trying to hunt down why this was breaking my build. Eventually discovered it was because I copy pasted the following code. `Div` needs to be `div`.


jxnblk:
Thanks!

PR #370, opened 2019-09-17
Add transparentize to polished color operations

robmoggach:
This is a suggestion... great when `backgroundColor` wants to be transparent.


robmoggach:
I currently achieve this with the following...

```jsx
import { get } from '@styled-system/css'
import { transparentize as Ptrans } from 'polished'
const g = (t, c) => get(t, `colors.${c}`, c)
export const transparentize = (c, n) => t => Ptrans(n, g(t, c))
```

jxnblk:
Thanks! Would you mind adding a simple unit test for this one?

robmoggach:
never written one before...

jxnblk:
No worries! It's not that hard, and I can help point you in the right direction if you get stuck. See the tests here for examples of the other color utilities: https://github.com/system-ui/theme-ui/blob/master/packages/color/test/index.js#L25-L28 – we'll want a test that takes one color value and checks to make sure that `transparentize` returns the correct value

jxnblk:
@robmoggach any luck with the unit test for this? If not, I can take a stab at adding one in

robmoggach:
I’d suggest you do it - I’m on python vfx work this month and next

PR #374, opened 2019-09-19
@theme-ui/color convert var to hex when useCustomProperties is true

### LINKED ISSUE #360: @theme-ui/color is broken when useCustomProperties is set to true
I'm getting the error:
`Error: Couldn't parse the color string. Please provide the color as a string in hex, rgb, rgba, hsl or hsla notation`

Looks like that as the colors in the theme are in the format: `var(--theme-ui-colors-primary, #FFB013)`, this is the breaking the polished functions.
A possible solution would be to check if useCustomProperties is true and in this case extract the color with a regex.
I can come up with a PR if needed!

johnpolacek:
Uses a regex to extract the hex value from the CSS Custom Variable value generated from a theme when useCustomProperties is true. Added test coverage. Verified that CSS Color Names are also processed correctly.

### DIFF
--- a/packages/color/src/index.js
+++ b/packages/color/src/index.js
@@ -1,7 +1,7 @@
 import * as P from 'polished'
 import { get } from '@styled-system/css'
 
-const g = (t, c) => get(t, `colors.${c}`, c).replace(/var\(--(\w+)(.*?), /,'').replace(/\)/,'')
+const g = (t, c) => get(t, `colors.${c}`, c).replace(/^var\(--(\w+)(.*?), /,'').replace(/\)/,'')
 
 export const darken = (c, n) => t => P.darken(n, g(t, c))
 export const lighten = (c, n) => t => P.lighten(n, g(t, c))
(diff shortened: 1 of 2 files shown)


packages/color/src/index.js
```diff
@@ -1,7 +1,7 @@
 import * as P from 'polished'
 import { get } from '@styled-system/css'
 
-const g = (t, c) => get(t, `colors.${c}`, c)
+const g = (t, c) => get(t, `colors.${c}`, c).replace(/var\(--(\w+)(.*?), /,'').replace(/\)/,'')
```
jxnblk: Awesome, thanks! Wondering if the regex should use `^`??
```suggestion
const g = (t, c) => get(t, `colors.${c}`, c).replace(/^var\(--(\w+)(.*?), /,'').replace(/\)/,'')
```

johnpolacek: Sure couldn't hurt.

jxnblk:
Awesome, thanks!

NickChristensen:
Hey, the second `replace` here is stripping the `)` in color values with parentheses, such as hsla or rgba. They get passed to polished as `hsl(240, 0%, 100%` instead of `hsl(240, 0%, 100%)`.

Issue #280, opened 2019-08-09
[Feature Request]: Responsive values outside of `sx` prop


worldeggplant:
I have a component that takes a `size` prop that cannot be set using the `sx` prop, which I want to pass a responsive value based on the current breakpoint.

Currently I've made a makeshift solution that reads `theme.breakpoints` and uses `window.matchMedia` to match the current breakpoint.
I think that theme-ui could benefit by having something similar.

```javascript
import { useThemeUI } from "theme-ui";

function matchBreakpoint(breakpoint) {
  const test = window.matchMedia(`(max-width: ${breakpoint})`);

  return test.matches ? true : false;
}

function getCurrentBreakpointIndex(breakpoints) {
  let matchedIndex = breakpoints.findIndex(matchBreakpoint);

  if (matchedIndex === -1) {
    matchedIndex = breakpoints.length - 1;
  }

  return matchedIndex;
}

const useResponsiveValue = responsiveValues => {
  const { theme: { breakpoints } } = useThemeUI();
  const responsiveIndex = getCurrentBreakpointIndex(breakpoints);

  // @TODO: update on window resize event

  return responsiveValues[responsiveIndex];
};

// ...

const SomeComponent = () => {
  const size = useResponsiveValue([10, 20, 30]);

  return <OtherComponent size={size} />;
}
```

jxnblk:
Thanks! I could certainly see the value in adding an optional hook package for using `window.matchMedia` with the theme object

dburles:
I also ran into needing this! Taking some of the code @worldeggplant posted, as a starting point, this is what I ended up with:

```js
import { useState, useEffect, useCallback } from 'react';
import { useThemeUI } from 'theme-ui';

const useBreakpoints = array => {
  const {
    theme: { breakpoints },
  } = useThemeUI();

  const getValue = useCallback(() => {
    let index = breakpoints.findIndex(
      breakpoint => window.matchMedia(`(max-width: ${breakpoint})`).matches,
    );
    if (index === -1) {
      index = breakpoints.length - 1;
    }
    return array[index];
  }, [array, breakpoints]);

  const [value, setValue] = useState(getValue);

  useEffect(() => {
    const onResize = () => {
      const newValue = getValue();
      if (value !== newValue) {
        setValue(newValue);
      }
    };
    window.addEventListener('resize', onResize);
    return () => window.removeEventListener('resize', onResize);
  }, [array, breakpoints, getValue, value]);

  return value;
};

[...]

PR #375, opened 2019-09-19
Add match-media package

dburles:
This PR adds a new package `@theme-ui/match-media` which provides two React hooks, `useBreakpointIndex` and `useResponsiveValue`.

Todo:

- [x] Documentation
- [x] Tests
- [ ] Support objects?

For https://github.com/system-ui/theme-ui/issues/280


packages/match-media/src/useResponsiveValue.js
```diff
@@ -0,0 +1,11 @@
+import { useThemeUI } from 'theme-ui'
+import useBreakpointIndex from './useBreakpointIndex'
+
+const useResponsiveValue = values => {
```
jxnblk: Thanks for working on this! Why not keep all the code in a single `src/index.js` file?

packages/match-media/src/useBreakpointIndex.js
```diff
@@ -0,0 +1,35 @@
...
+  const [value, setValue] = useState(getIndex)
+
+  useEffect(() => {
+    const onResize = () => {
+      const newValue = getIndex()
+      if (value !== newValue) {
+        setValue(newValue)
+      }
+    }
+    window.addEventListener('resize', onResize)
```
jxnblk: What were the issues with the match media listeners? I'm fairly sure that's a more performant way to handle this

dburles: It was actually my first approach but I found it quite tricky and edge-casey, so for the sake of time constraints on a project that required it, I decided to instead opt for this simpler implementation, and in reality, it should still be very performant. It will only update state if the value actually changes, and I don't think calculating the breakpoints with `window.matchMedia` is a particularly CPU intensive process. It definitely seems more correct to use the listeners, but then again, for this particular use of them their API doesn't seem to match up very nicely.

packages/match-media/src/useResponsiveValue.js
dburles: No real reason I guess, I usually like to write one hook per file. Happy to do that if you prefer.

packages/match-media/src/useBreakpointIndex.js
jxnblk: Yeah, no worries, just more curious what the edge cases might've been etc.

packages/match-media/src/useBreakpointIndex.js
```diff
@@ -0,0 +1,35 @@
...
+import { useThemeUI } from 'theme-ui'
+
+const useBreakpointIndex = () => {
+  const {
+    theme: { breakpoints },
+  } = useThemeUI()
+
+  const getIndex = useCallback(() => {
+    let index = breakpoints.findIndex(
+      breakpoint => window.matchMedia(`(max-width: ${breakpoint})`).matches
```
jxnblk: These should be `min-width` right? The `0` in responsive array values corresponds to no media query, so maybe it'd make sense to pad the `theme.breakpoints` array with an item for that one if this uses `findIndex` – I suspect mixing `min-width` and `max-width` would yield different results when viewport is at exactly the width of the breakpoint (e.g. at `640px` viewport when `breakpoints = [ '40em', '52em', '64em' ]`)

packages/match-media/src/useBreakpointIndex.js
```diff
@@ -0,0 +1,35 @@
+import { useState, useEffect, useCallback } from 'react'
+import { useThemeUI } from 'theme-ui'
+
+const useBreakpointIndex = () => {
+  const {
+    theme: { breakpoints },
```
jxnblk: Also wonder if providing a default value for `breakpoints` would be good here since `@styled-system/css` will fall back to defaults

packages/match-media/src/useResponsiveValue.js
jxnblk: Mostly just curious here – personally, I think it makes sense to keep these together until there is a good reason to split into multiple files, but this is fine as-is

packages/match-media/src/useBreakpointIndex.js
dburles: Yes you're right, there is a px difference. I've made an update to address this, and it simplifies the logic too.

packages/match-media/src/useBreakpointIndex.js
dburles: Yes actually I was wondering about that, should we reference it from here https://github.com/styled-system/styled-system/blob/master/packages/css/src/index.js#L10 or just embed it.

packages/match-media/src/useBreakpointIndex.js
dburles: Here's a demo https://codesandbox.io/s/theme-uimatch-media-tzv1s

packages/match-media/src/useResponsiveValue.js
dburles: I figured I might as well update it, in this case I also prefer them in a single file

packages/match-media/src/useBreakpointIndex.js
jxnblk: I think it's fairly harmless to hard-code the values in this case

dburles:
@jxnblk Just added a first go at some documentation

packages/match-media/README.md
```diff
@@ -0,0 +1,38 @@
...
+React hooks for referencing theme-ui breakpoints.
+
+## Overview
+
+This package provides two React hooks (`useResponsiveValue` and `useBreakpointIndex`) for referencing responsive values outside of the `sx` prop.
+
+## Installation
+
+```sh
+yarn add @theme-ui/match-media
```
jxnblk: For consistency with the rest of the docs:
```suggestion
npm i @theme-ui/match-media
```

jxnblk:
Awesome! It's looking great! I think this will also need some unit tests to pass CI – you should be able to mock the `window.matchMedia` method, but let me know if you get stuck

dburles:
I started working on the tests and I've ran into the context issue again 😞. As far as I can tell, it seems the underlying cause of the problem is that the match-media package has a peerDependency on theme-ui, and imports `useThemeUI` from it in order for `useBreakpointIndex` to read the user defined breakpoints. It then creates a seperate context with the theme property as `null` (as it is set by default).

As you might expect, the context is correct when importing `useThemeUI` directly (`import { useThemeUI } from '../../theme-ui/src'`).

I don't really have the time to dig into the provider/context code myself. If it helps, I could push a simple test that reproduces the problem if you would like to take a look at it?

jxnblk:
Yeah, context can be annoying like that sometimes... Feel free to push up what you have, and I can try to take a look later

dburles:
Ended up writing the full test suite, to reproduce the issue, swap the comment on these imports: 
https://github.com/dburles/theme-ui/blob/match-media/packages/match-media/src/index.js#L2-L3

dburles:
Everything should be good to go! The context issue with the tests [has been addressed with this commit](https://github.com/system-ui/theme-ui/pull/375/commits/9ce1c8a9bda370231846cc25d566f331b31c3b11#diff-3b2cdbb88ba01d559592867a7284f54dR3).

dburles:
ping @jxnblk!

jxnblk:
Sorry for the delay! This looks fantastic! Thanks for working on this. Will merge and publish shortly :shipit:

PR #378, opened 2019-09-20
fix: edit link to git editor page

JCofman:
Hi and thx a lot your work 👋,

When I tried to look at the source code of the doc pages I was redirected to the wrong GitHub URL. This should fix it :)

### DIFF
--- a/packages/docs/src/components/edit-link.js
+++ b/packages/docs/src/components/edit-link.js
@@ -3,8 +3,19 @@ import { Location } from '@reach/router'
 import { jsx } from 'theme-ui'
 
 const getHREF = (base, location) => {
-  if (location.pathname === '/') return false
-  return base + location.pathname.replace(/\/+$/, '') + '.mdx'
+  if (location.pathname === '/') return
+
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/edit-link.js
```diff
@@ -32,7 +31,7 @@ export const EditLink = ({ base, children, ...props }) => (
 
 EditLink.defaultProps = {
   base:
-    'https://github.com/system-ui/theme-ui/edit/master/packages/docs/src/pages',
```
jxnblk: Thanks, but this looks like it'd "fix" the link for pages under `/recipes` but break other pages. I think the edit link should either work for both cases or just be disabled for pages in `/recipes`

JCofman: Thanks a lot for your review sorry for that 🙈 I updated the pull request. It's not a pretty solution so far but I think it works for now and I like the idea of quickly creating pull-request with such a link :)

jxnblk:
Thanks! I think this'll work for now

PR #386, opened 2019-09-23
Update  NextJS example

### LINKED ISSUE #379: Issues with Next.js example
Hi, I started working with Next.js and Theme UI and found two issues:

- Can we remove the code inside `getInitialProps` in `_app.js` (and perhaps `_document.js`)?
- Can we remove `<Container />` in `_app.js`?

With `getInitialProps` in `_app.js`, every page becomes server-rendered when we `next build` and we don't get automatic static optimization (see related [Next.js blog](https://nextjs.org/blog/next-9#automatic-static-optimization) and [docs](https://nextjs.org/docs#custom-app)). I looked into the pull request (#277) but not sure if this code is necessary. 

Also when I run the example, it warns `<Container />` is deprecated. 

I am pretty new to Next.js so I am not sure how those two works in the context of Theme UI. Perhaps @jordanoverbye knows? When I removed those everything seems to work as expected but maybe I am missing something.

jordanoverbye:
I've made some optimisations to the NextJS example so its more inline with `create-next-app` defaults. 

This PR removes a deprecation warning and also allows pages to be statically optimised by next.

Closes #379

### DIFF
--- a/examples/next/pages/_app.js
+++ b/examples/next/pages/_app.js
@@ -1,34 +1,33 @@
 import React from 'react'
-import App, { Container } from 'next/app'
+import App from 'next/app'
 import { ThemeProvider, Styled, ColorMode } from 'theme-ui'
 
 import Header from '../components/Header'
 import theme from '../src/theme'
 
 class MyApp extends App {
-  static async getInitialProps({ Component, ctx }) {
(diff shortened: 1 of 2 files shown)


jxnblk:
Seems legit, thanks!

PR #389, opened 2019-09-25
Use line height from theme for headings

johno:


johno:
An additional thought, should we adjust this to use a `textStyles` variant @jxnblk? I guess that could be another pass to make the presets more consistent?

jxnblk:
Yeah, I feel like we could come up with some more conventions around using variants, especially for cases like this

PR #398, opened 2019-10-04
Clean up unused argument

michaellopez:
The code for `useColorMode` got changed to make the argument `initialMode` unused. This tripped me up because the typings say it exists, but the code doesn't use it. So I propose to remove it here first and then update the typings at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/theme-ui/index.d.ts#L222

This PR also includes some formatting done by Prettier on commit.


jxnblk:
Thanks!

michaellopez:
@jxnblk thanks. I'll continue to send a PR to the typings

PR #406, opened 2019-10-08
Fix missing key warning

millette:
I was seeing

> Warning: Each child in a list should have a unique "key" prop.
> Check the top-level render call using <EmotionCssPropInternal>.

Since `key` is special and isn't passed down in `props`, I moved it directly in `<ColorSwatch>`.

### DIFF
--- a/packages/style-guide/src/ColorPalette.js
+++ b/packages/style-guide/src/ColorPalette.js
@@ -39,6 +39,7 @@ export const ColorRow = ({
           const swatch = (
             <ColorSwatch
               {...props}
+              key={key}
               name={id}
               color={id}
               size={size}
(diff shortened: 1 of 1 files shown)


packages/style-guide/src/ColorPalette.js
```diff
@@ -51,7 +52,6 @@ export const ColorRow = ({
             return render({
               swatch,
               color,
-              key,
```
jxnblk: Thanks! I *think* this argument is needed in the `@theme-ui/editor` package – can you keep this in the render prop, please?

millette: Here you go.

PR #407, opened 2019-10-10
Remove ColorModeProvider from Dark-Mode example

jasonrundell:
`ColorModeProvider` no longer available in `theme-ui` package so the example will not actually run. Followed the migration instructions and bingo! Back in action 😉


jxnblk:
Thanks!

PR #411, opened 2019-10-17
Add components package

jxnblk:


epilande:
@jxnblk this looks great! 

The components here look identical to `rebass`. 
If you don't mind me asking, are there other plans for `rebass`?

dburles:
Hey @jxnblk wondering why the decision that some of the components expose a few styling props rather than simply using the sx prop? e.g. https://github.com/system-ui/theme-ui/pull/411/files#diff-5507edef875ba633e03cb339c111d617R70

Grsmto:
>Hey @jxnblk wondering why the decision that some of the components expose a few styling props rather than simply using the sx prop?

I'm also wondering this. After migrating from system props to `sx` I kinda supposed the system props were deprecated since you can just do everything within the `sx` prop without the downsides of props clash/discoverability.
That's also confusing to have 2 different APIs to style a component (as it's currently in the [new components doc](https://theme-ui.com/components/flex)). 
Would be interesting to know more about your thoughts on this!

jxnblk:
The inclusion of style prop shorthands is simply for ergonomics. Adding e.g. `mb={2}` is a shortcut to using the `sx` prop

dburles:
@jxnblk I feel like the downside there is that you then have two ways to do the same thing, which can be a bit of cognitive overhead.  In building out latest design system I've managed to make `sx` prop work in every case, that way the only props the components are left with are not styling related. This to me was one of the main advantages of theme-ui and the `sx` prop over styled-system.

PR #415, opened 2019-10-21
fix theme-ui/provider typo

lxynox:
webpack define plugin or rollup replace plugin didn't work as expected due to this typo.


jxnblk:
Good catch, thanks!

PR #421, opened 2019-10-24
Add match-media package docs under hooks

dburles:
I think it makes sense to go here


jxnblk:
Thanks! We'll probably want to revisit how the optional packages are organized in the docs at some point, but, yeah, I think this makes sense for now

PR #441, opened 2019-11-08
[@theme-ui/colors] Implement alpha color function

colebemis:
I noticed the beginning of an `alpha` color function was commented out so I took a stab at finishing it. The function changes the transparency of a given color using the `rgba` function from Polished. 

**Example:**

```js
const theme = {
  colors: { primary: '#f00' }
}

alpha('primary', 0.5)(theme) // rgba(255, 0, 0, 0.5)
```

Let me know if this is how you intended for the `alpha` function to work 😸 

### To do
- [x] Add documentation
- [x] Write tests

### DIFF
--- a/packages/color/README.md
+++ b/packages/color/README.md
@@ -116,6 +116,16 @@ import { tint } from '@theme-ui/color'
 // tint('primary', amount)
 ```
 
+### `alpha`
+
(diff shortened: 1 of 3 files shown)


jxnblk:
Awesome, thanks!

PR #460, opened 2019-11-15
[match-media] Add default breakpoint index arg to match-media hooks

### LINKED ISSUE #438: [@theme-ui/match-media] Breaks gatsby builds due to undefined window
Using `@theme-ui/match-media`'s `useResponsiveValue` hook in Gatsby causes the build to fail, due to its use of `window.matchMedia`.

I'd normally add a `typeof window === 'undefined'` check, but am not allowed to conditionally call a hook. Is there an alternative userland solution for this that I'm not thinking of?

I could add the check to the hook - is returning `breakpoints[0]` a sensible default?

joestrouth1:
Closes #438 

Should it also throw if the `defaultIndex` is not in range of the breakpoints array?

### DIFF
--- a/packages/match-media/README.md
+++ b/packages/match-media/README.md
@@ -58,3 +58,22 @@ const MyComponent = () => {
 }
 ```
 
+Each hook uses `breakpoints[0]` by default, if not specified.
+
+```js
+import { useResponsiveValue, useBreakpointIndex } from '@theme-ui/match-media'
+
+const MyComponent = () => {
+  const color = useResponsiveValue(['red', 'green', 'blue'])
+  const index = useBreakpointIndex()
+
+  return (
+    <div>
+      The current color is: {color}, and the current index is: {index} 
+   // Gatsby will output: 
+   // The current color is: red, and the current index is: 0
+    </div>
+  )
+}
+```
+
--- a/packages/match-media/README.md
+++ b/packages/match-media/README.md
@@ -36,3 +36,23 @@ const MyComponent = () => {
   )
 }
 ```
+
+Each hook also accepts a default breakpoint index to use when rendering on the server.
+
+```js
+import { useResponsiveValue, useBreakpointIndex } from '@theme-ui/match-media'
+
+const MyComponent = () => {
+  const color = useResponsiveValue(['red', 'green', 'blue'], 1)
+  const index = useBreakpointIndex(0)
+
+  return (
+    <div>
+      The current color is: {color}, and the current index is: {index} 
+   // Gatsby will output: 
+   // The current color is: green, and the current index is: 0
+    </div>
+  )
+}
+```
+
--- a/packages/match-media/README.md
+++ b/packages/match-media/README.md
@@ -48,7 +48,7 @@ const MyComponent = () => {
   const color = useResponsiveValue(['red', 'green', 'blue'], {
     defaultIndex: 1
   })
-  const index = useBreakpointIndex({ defaultIndex: 1 })
+  const index = useBreakpointIndex({ defaultIndex: 0 })
 
   return (
     <div>
--- a/packages/match-media/README.md
+++ b/packages/match-media/README.md
@@ -39,14 +39,16 @@ const MyComponent = () => {
 
 ### Server side rendering
 
-Each hook also accepts a default breakpoint index to use when rendering on the server.
+Each hook also accepts an options object, used to set a default breakpoint index when rendering on the server.
 
 ```js
 import { useResponsiveValue, useBreakpointIndex } from '@theme-ui/match-media'
 
 const MyComponent = () => {
-  const color = useResponsiveValue(['red', 'green', 'blue'], 1)
-  const index = useBreakpointIndex(0)
+  const color = useResponsiveValue(['red', 'green', 'blue'], {
+    defaultIndex: 1
+  })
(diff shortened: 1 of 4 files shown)


dburles:
Yeah it would make sense to also throw if the index is out of range. I'm also wondering if the second argument should be an object, as it will make it easier to support additional options in future.

packages/match-media/README.md
```diff
@@ -36,3 +36,23 @@ const MyComponent = () => {
   )
 }
 ```
+
+Each hook also accepts a default breakpoint index to use when rendering on the server.
```
dburles:
```suggestion
### Server side rendering

Each hook also accepts a default breakpoint index to use when rendering on the server.
```

packages/match-media/src/index.js
```diff
@@ -4,19 +4,32 @@ import { useThemeUI } from 'theme-ui'
...
+      if (typeof defaultIndex === 'number') {
+        if (defaultIndex < 0 || defaultIndex > breakpoints.length - 1) {
+          throw new RangeError(
+            `Default breakpoint index out of range. Theme has ${breakpoints.length} breakpoints, got index ${defaultIndex}`
+          )
+        }
+        return defaultIndex
+      }
+      throw new TypeError(
+        'To use @theme-ui/match-media hooks on the server, you must pass a default index. Got: ' +
```
jxnblk: Thanks for looking into this. I might've missed this part of the conversation, but why not default to `0` on the server? Seems like a sensible default IMO

packages/match-media/test/ssr.js
```diff
@@ -0,0 +1,54 @@
+/**
```
jxnblk: I think using a separate `ssr.js` test file for this is a-okay 👍

packages/match-media/src/index.js
joestrouth1: That was my initial thought until @dburles mentioned that it could cause unexpected re-rendering when loaded at larger screen sizes. Could be jarring for desktop users to see content replaced shortly after load

jxnblk: My guess is that that would be an issue regardless. I think having the ability to change the default makes sense, but also that this should have a sensible default

joestrouth1: Oh - I misunderstood. A default for the default 😅 updated

dburles:
I’m not sure the default is a good idea as it could yield some unexpected rendering, which I’d imagine would be difficult for the user to understand why. 

Previously if you used the hook on the server it would force you to acknowledge the need for specifying a default index, that way you’ll understand what’s going on.

dburles:
Also I really think the second argument should take an object instead of the index, eg: `{ defaultIndex: 0 }`. Right now we would be stuck if we ever wanted to allow any additional options.

jxnblk:
@dburles I suspect we could cover the behavior in documentation. Personally, I expect the server to provide whatever the non-media-query based state is, which should be `0` with the way this library works.

The single argument signature is also similar to the official `useState()` hook which accepts the default value as its first argument. If we do need to add more options later, I think a second options argument wouldn't be too terrible. That said, I haven't used this hook at all yet, so I might be glossing over some valid concerns

dburles:
> @dburles I suspect we could cover the behavior in documentation. Personally, I expect the server to provide whatever the non-media-query based state is, which should be `0` with the way this library works.

Yeah that's valid, so long as it's clearly documented, I can definitely imagine it being quite a confusing thing to debug otherwise.

> The single argument signature is also similar to the official `useState()` hook which accepts the default value as its first argument. If we do need to add more options later, I think a second options argument wouldn't be too terrible. That said, I haven't used this hook at all yet, so I might be glossing over some valid concerns

That's true, I guess in our case we have some freedom in handling it a little differently since it's not a direct mapping to the state value in the case of `useResponsiveValue`, but instead of course a numerical index. 

Plus it's also only applicable to the server. I can't think ahead as to what additional options we might require (if any of course), but either way if we did ever require any, SPA code might end up littered with `useResponsiveValue(undefined, { someFutureOption: true })` :).

In any case, I'm not terribly fussed, just my thoughts on it.

packages/match-media/src/index.js
```diff
@@ -34,9 +47,9 @@ export const useBreakpointIndex = () => {
   return value
 }
 
-export const useResponsiveValue = values => {
+export const useResponsiveValue = (values, defaultIndex) => {
```
dburles:
```suggestion
export const useResponsiveValue = (values, options) => {
```

packages/match-media/src/index.js
```diff
@@ -34,9 +47,9 @@ export const useBreakpointIndex = () => {
...
   const { theme } = useThemeUI()
   const array = typeof values === 'function' ? values(theme) : values
-  const index = useBreakpointIndex()
+  const index = useBreakpointIndex(defaultIndex)
```
dburles:
```suggestion
  const index = useBreakpointIndex(options)
```

packages/match-media/src/index.js
joestrouth1: Good catch 😅

dburles:
This is all looking really good, @joestrouth1 is there anything else remaining?

joestrouth1:
Not that I'm aware of.

jxnblk:
Awesome! Thanks for working on this! 😎

Issue #412, opened 2019-10-18
Global jsx/sx pragma


alexluong:
Hi. Is there anyway we can use `sx` prop without defining the custom pragma on the top of the files? Is there any drawback of using a babel plugin for that?

epilande:
Not sure if this is the recommended approach but this is how I build my components.

Create a base `Box` component or use the one from [reflexbox](https://github.com/rebassjs/rebass/tree/master/packages/reflexbox).

Here's an abstract example:
```js
import styled from '@emotion/styled'
import css from '@styled-system/css';

const sx = props => css(props.sx)(props.theme);

export const Box = styled('div')(sx)
```

Usage:
```jsx
```

Then have everything extend `Box`.

jxnblk:
@alexluong you can set up a custom babel config to automatically add the jsx pragma comment, but it will affect all files, which might or might not be what you want. See https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#pragma

@epilande this is essentially coming in #411

alexluong:
@jxnblk thank you, that is what I’m asking. Is there a situation where I don’t want it to affect all files? What is the downside?

adamsoffer:
@jxnblk I haven't been able to successfully get theme-ui to work with the custom babel config 

Check out the error below:
https://codesandbox.io/s/theme-ui-starter-4ydj0

Any idea what I'm doing wrong?

jxnblk:
@adamsoffer I don't see the `jsx` import in your codesandbox, did you forget to import that?

adamsoffer:
@jxnblk Even when I import jsx, the sx attributes still get ignored. The theme only is applied when a pragma is present.

adamsoffer:
Gotcha. How do I configure the babel plugin so that I don't have to add a pragma to every file?

adamsoffer:
@jxnblk I tried a few different configurations but no luck. Have you confirmed using this plugin works on your  end?

lachlanjc:
The mentioned issues/PRs I think cover this functionality for setting the JSX pragma & the Components have since been published/exported from the default package; @hasparus I’m cool with closing this issue

PR #479, opened 2019-11-24
babel-preset-theme-ui

lewisacidic:
This PR adds a small package to the monorepo representing a preset to configure the theme-ui babel pragma automatically.   This addresses #412.

I called it `babel-preset-theme-ui`, but it could be `@theme-ui/babel-preset`.

It could do with a docs section, and maybe an example in `examples/`, but I don't have time right now - thought it would be good to kick it off.

### DIFF
--- a/packages/babel-preset-theme-ui/package.json
+++ b/packages/babel-preset-theme-ui/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "babel-preset-theme-ui",
+  "version": "0.2.49",
+  "description": "Babel preset to enable the theme-ui custom pragma by default.",
+  "keywords": [
+    "babel",
+    "babel-preset",
+    "theme-ui"
+  ],
+  "author": "Rich Lewis <opensource@richlew.is>",
+  "license": "MIT",
+  "main": "src/index.js",
+  "repository": "system-ui/theme-ui",
+  "dependencies": {
+    "@babel/plugin-transform-react-jsx": "^7.7.4",
+    "@wordpress/babel-plugin-import-jsx-pragma": "^2.3.0"
+  }
+}
--- a/packages/babel-preset-theme-ui/package.json
+++ b/packages/babel-preset-theme-ui/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "babel-preset-theme-ui",
-  "version": "0.2.52",
+  "name": "@theme-ui/babel-preset-sx-prop",
+  "version": "0.3.1",
   "description": "Babel preset to enable the theme-ui custom pragma by default.",
   "keywords": [
     "babel",
--- a/packages/babel-preset-theme-ui/package.json
+++ b/packages/babel-preset-theme-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "babel-preset-theme-ui",
-  "version": "0.2.49",
+  "version": "0.2.52",
   "description": "Babel preset to enable the theme-ui custom pragma by default.",
   "keywords": [
     "babel",
@@ -13,6 +13,6 @@
   "repository": "system-ui/theme-ui",
   "dependencies": {
     "@babel/plugin-transform-react-jsx": "^7.7.4",
-    "@wordpress/babel-plugin-import-jsx-pragma": "^2.3.0"
+    "@emotion/babel-plugin-jsx-pragmatic": "^0.1.5"
   }
 }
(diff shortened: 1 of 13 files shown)


packages/babel-preset-theme-ui/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "babel-preset",
+    "theme-ui"
+  ],
+  "author": "Rich Lewis <opensource@richlew.is>",
+  "license": "MIT",
+  "main": "src/index.js",
+  "repository": "system-ui/theme-ui",
+  "dependencies": {
+    "@babel/plugin-transform-react-jsx": "^7.7.4",
+    "@wordpress/babel-plugin-import-jsx-pragma": "^2.3.0"
```
vxna: Why not just use `@emotion/babel-plugin-jsx-pragmatic` since `@emotion/babel-preset-css-prop` [does](https://github.com/emotion-js/emotion/blob/master/packages/babel-preset-css-prop/src/index.js#L19-L24) exactly that?

lewisacidic: I hadn't seen the preset (haven't used emotion much), the wp one came up first in my quick google search. Look like they do the same thing, but the emotion one seems much more sensible to use. I can update.

cweekly:
It looks like the only thing holding up this PR is the lockfile conflicts. Anyone able to do that? Global jsx pragma (w/ help from `@emotion/babel-plugin-jsx-pragmatic`) would be so great.

lewisacidic:
Still needs lazy me to update, I'll have a go now (sorry all I was conferencing most of Dec).

lewisacidic:
OK I've changed to `@emotion/babel-plugin-jsx-pragmatic` and added an example using gatsby.  Hopefully this fixes the yarn.lock conflict too.

I guess we stil need a section in the docs.

jxnblk:
@lewisacidic Thanks for the update! I think the README should suffice for the time-being. It's not hard to add a page to the docs using that. My only question now is should the package be named `@theme-ui/babel-preset` to be more consistent with this library?

cweekly:
> @lewisacidic Thanks for the update! I think the README should suffice for the time-being. It's not hard to add a page to the docs using that. My only question now is should the package be named `@theme-ui/babel-preset` to be more consistent with this library?

Yes, thanks @lewisacidic!
FWIW, namespacing under @theme-ui makes sense to me too.

lewisacidic:
Hi @jxnblk - I chose `babel-preset-theme-ui` at the start to be consistent with other babel presets like `babel-preset-gatsby`, and to distance from the `@theme-ui/preset-*` packages, but I think `@theme-ui/babel-preset` would be distinct enough and as you said, it's nice to use the same namespace scope as the other theme-ui packages.

Up to you!

jxnblk:
Sorry I dropped the ball on this. I think `@theme-ui/babel-preset` would make the most sense. Would you have time to update the name? If not, I can use suggested changes in the relevant spots

lewisacidic:
No worries, I'll update once I'm off work this eve.

vxna:
@jxnblk for me `@theme-ui/babel-preset-sx-prop` makes the most sense. This perfectly aligns with `@emotion/babel-preset-css-prop` package and current ecosystem. 

Babel also accepts short names so users could just write it as `@theme-ui/sx-prop`.

Thoughts?

lewisacidic:
@vxna I agree, more explicit, more tightly follows babel preset naming conventions, would further break any confusion with the theme-ui presets.

jxnblk:
I think `@theme-ui/babel-preset-sx-prop` seems fine

krishnagopinath:
Thanks for working on this, @lewisacidic! Would it possible to take this to completion? I'd love to be able to use this babel plugin in the app I'm building. Would you need any assistance to finish this? 👐

lewisacidic:
Oh oops yeah - I guess I missed the last update. Once I'm back from work, I'll rebase and change the name to `@theme-ui/babel-preset-sx-prop` then ping.

lewisacidic:
@jxnblk: I made the changes about a minute before falling asleep and didn't check the ci results - I'm not sure what's going on exactly?

jxnblk:
@lewisacidic CI looks like an unrelated snapshot change -- probably due to a dependency update. I think it's safe to ignore and can double check locally

dburles:
Aside from not requiring the pragma comment, it would be great if we could get CSS source maps working too. I've been playing around with `@emotion/babel-preset-css-prop` the config and unable to get it to function with the `sx` prop, glancing over the source it's likely because it's expecting a `css` prop, (which as you would expect works just fine!). Haven't really gone any further than that.

clehene:
This is great! I hope it gets merged soon.
One question - currently we have to use the full `<React.Fragment></React.Fragment>` instead of the shorthand `<></>`. Does this PR solve this requirement too?

lachlanjc:
Thank you so much for this contribution! I’m closing this because our latest versions have support for the new JSX automatic runtime, which obviates the need for this. (Documentation coming with #1335.)

PR #498, opened 2019-12-03
@theme-ui/editor: Use tinycolor in ThemeColorPicker

jordanoverbye:
This fixes a bug in Blocks UI: https://github.com/blocks/blocks/issues/74

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -13,7 +13,8 @@
     "@theme-ui/style-guide": "^0.2.49",
     "color": "^3.1.2",
     "react-color": "^2.17.3",
-    "reakit": "^1.0.0-beta.6"
+    "reakit": "^1.0.0-beta.6",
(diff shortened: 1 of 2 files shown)


packages/editor/package.json
```diff
@@ -13,7 +13,8 @@
     "@theme-ui/style-guide": "^0.2.49",
     "color": "^3.1.2",
     "react-color": "^2.17.3",
-    "reakit": "^1.0.0-beta.6"
+    "reakit": "^1.0.0-beta.6",
+    "tinycolor2": "^1.4.1"
```
jxnblk: This is the same lib that `react-color` uses, right?

jxnblk:
LGTM :shipit:

packages/editor/package.json
jordanoverbye: Yeah it is 👍

PR #523, opened 2019-12-22
Fix overflow error in LiveError component

IsAmrish:
Signed-off-by: isamrish <askmaurya48@gmail.com>

This commit will fix overflow error in LiveError component
in docs package

--

I was playing with code in theme-ui and I got this overflow issue in LiveError components

Before:

overflow-issue

After being fixed:

overflow-solution


jxnblk:
Thanks!

PR #541, opened 2020-01-04
Add custom columns support to Grid component

lachlanjc:
I've found while using Theme UI Components' Grid component that the `columns` prop is frustrating because it assumes I only want equally-sized columns. ([Example here](https://github.com/lachlanjc/gunfunded/blob/f88ce280d71ee2aedb916962fad9eb7d939391d7/pages/contributors.js#L88)) This PR changes it to assume equally-sized columns when a number is provided, otherwise allow freeform [grid column syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns).

Let me know if you want anything changed!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- `@theme-ui/components`: on Grid component, allow custom `columns` definitions via strings #541
+
 ## v0.2.53 2019-12-19
 
 - `@theme-ui/color`: add `transparentize` function #370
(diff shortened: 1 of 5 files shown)


jxnblk:
Thanks! My thinking was that this could be achieved with the `sx` prop if you wanted to use native CSS syntax, but this could make sense. Is there any way that this might cause a breaking change for anyone currently using the component?

lachlanjc:
That makes sense. I don’t think anyone could really use meaningful string syntax. The current syntax/docs only show using numbers or null, & if you passed a string with a unit (`256px` or something), I don’t think that’d work (result would be `repeat(256px, 1fr)`). So I guess if they were passing the number of columns as a string (`columns=“8”`) & expecting it to expand into the repeat syntax, that could break, but it seems unlikely.

jxnblk:
Thanks for clarifying! I think this is GTG then :shipit:

PR #542, opened 2020-01-04
Add note on frameworks to Theme Color Meta Tag page

lachlanjc:
(I believe we'll need to update this with `rawColors` with the next release, too?)


jxnblk:
Thanks! (And yes, looks like that will need to be updated with v0.3)

Issue #555, opened 2020-01-08
Community Discord?


Sonicrida:
**Describe the solution you'd like**
I briefly mentioned the idea of a theme-ui discord to @KyleAMathews and he suggested that ask here to see if I could help in setting one up! I think it would be a great idea to help form a community around the project. I've setup a few programming discords in the past and also help to moderate the Reactiflux discord. I would love to help get one setup if you all feel as though this is a good idea!

**Additional context**
I think that discords are great supplementary tools for open source projects. I've seen it being extra helpful with Gatsby and I'm sure that Theme-UI could benefit from the same.

jxnblk:
I've never used Discord myself. If you'd like to set one up and add a link in the readme, we could give it a shot

Sonicrida:
Awesome! I'll set it up over the next few days 👍

mrmartineau:
Don't want to step on any toes, but [Spectrum](https://spectrum.chat/) might be a better option. It has SEO by default where Discord doesn't. I know the two services serve different purposes, but as a channel for support and sharing, in my opinion, it is better than discord.

Sonicrida:
That's a fair point to bring up. The main reason I shy away from spectrum is that I've seen way too many open source communities migrate from things like slack and discord to have their traffic and discussion go way down and sometimes die out. A few seem to thrive but I have a hard time recommending the platform because of this. I don't really understand why this happens but I've heard a fair bit of UX complaints. I think the lack of real time nature results in less discussion happening. Dedicated apps like Discord or slack have more control over customizing the user experience as well. 

Based on my experiences helping to run Reactiflux and being apart of a few of the other open source communities on Slack and Discord like the ones listed here, https://discordapp.com/open-source I think that any loss of SEO is worth the amount of help and overall learning, interaction, and community building that is gained by using Discord.

CanRau:
Thought about spectrum, too. Yet it feels kinda less stable, it always flickers after the first seconds, I just started reading, then everything disappears for a sec just to return as it looked before^^ that's kinda annoying 😅 
SEO discoverability sounds great though, Discord and Slack have search functions yet I like to be able to just search Google and know it'll search the chat as well..
Just some thoughts 😉 I'm joining whatever will be used ❤️

dcastil:
Now that GitHub Discussions are in beta, how about trying that? That would fit right in.

I didn’t find a marketing page for the feature, but Next.js is using it already: https://github.com/zeit/next.js/discussions

[...]

PR #557, opened 2020-01-09
Add link to community discord

Sonicrida:
As discussed in #555 , I have created a community discord so that there can be a place to build a real time discussion community around the project. 

I have setup a few channels for discussions for now.

### DIFF
--- a/packages/theme-ui/README.md
+++ b/packages/theme-ui/README.md
@@ -160,5 +160,6 @@ export default props => (
 - [Theme Spec](https://theme-ui.com/theme-spec)
 - [Gatsby Plugin](https://theme-ui.com/gatsby-plugin)
 - [API](https://theme-ui.com/api)
+- [Community Discord](https://discord.gg/wAD6p8q)
 
 MIT License


packages/theme-ui/README.md
```diff
@@ -160,5 +160,6 @@ export default props => (
 - [Theme Spec](https://theme-ui.com/theme-spec)
 - [Gatsby Plugin](https://theme-ui.com/gatsby-plugin)
 - [API](https://theme-ui.com/api)
+- [Community Discord](https://discord.gg/wAD6p8q)
```
jxnblk: Is there not an option for a pretty URL with discord? Also I can’t seem to open this link in the browser on my phone...

Sonicrida: I don't believe that discord supports phone browsers. When I open it on my phone, it deeplinks to the app when I click the accept button. 

As far as pretty URLs go, I think you need to either earn that with server boosts (people join your server and basically donate with an in app feature) *or* you have to get the server verified like what the Gatsby Discord did https://discordapp.com/verification

The boost option isn't as reliable because users have to keep them active and it takes 30 boosts for the url so verification would make the most sense I think if we wanted this.

jxnblk: Thanks for the info! Sounds pretty limiting, but I guess this thing is really intended for gaming not OSS communities...

CanRau: why not slack? just asking don't have an opinion, just curios 😉

Sonicrida: Eh It's not as big of an issue as it seems I'd say. I've seen and helped plenty of successful open source discords.

You can see that a lot of projects have grown pretty well here https://discordapp.com/open-source.

As for slack, it's a lot easier to not have to make accounts among other reasons. You have a bit more control over categorization and you don't have any message history limits to deal with for free servers are other things.

hasparus:
Can I revive the topic? 

While I kinda like Discord and find their [engineering blog](https://blog.discordapp.com/tagged/engineering) pretty cool, there are [mixed feelings](https://stallman.org/discord.html) about them in the open source community.

Seeing that this Discord thing didn't really lift off, may I propose Spectrum? :smile:

Sonicrida:
I haven't really seen another platform generating large open source communities as effectively as Discord has. A lot of the points in that article, especially the ones about TOR.....I'd say don't represent the majority of users. 

Either way, it doesn't seem like there isn't much interest in this from @jxnblk . It may be better to close this and just use the official discussions tab on this repo once Github enables that site-wide.

jxnblk:
I was happy to leave this open to see if others found value in a discord community, but yeah, I kind of suspect we'd have more traction with the official GitHub discussions, whenever that's available

PR #563, opened 2020-01-12
Add Styled.article

lachlanjc:
I feel like this might be incredibly dumb, so apologies if so. But I'd love a simple way to style text content (possibly user-generated, like Markdown in comments, but not MDX) in Theme UI, separately from application/`root` styles.

The article tag here (I hope) would allow doing `<Styled.article>` & having `theme.styles.article` give those styles, keeping the code (as you can see) super concise. Thoughts?

### DIFF
--- a/packages/theme-ui/src/components.js
+++ b/packages/theme-ui/src/components.js
@@ -33,6 +33,7 @@ const tags = [
   'thematicBreak',
   // other
   'div',
+  'article',
   // theme-ui
   'root',
 ]


jxnblk:
Thanks! I don't want to shoot this down immediately, but want to provide some context for discussion, so bear with me 🙏 

The `Styled` component is largely intended as a way to reuse styled elements from MDX, so the only elements currently in that API are ones that can be rendered in markdown, and since markdown doesn't have a notion of an `<article>` element, it's not currently in the `Styled` component.

To avoid adding a lot of potential dead code for users who wouldn't use something like `Styled.article` (objects can't really be treeshaken and there are a *lot* more HTML elements than what `Styled` has), I'd recommend using variants and/or creating your own component to handle something like this. Note that you can still use the `theme.styles` object to store these variants if you'd like:

```jsx
// Example
<article sx={{ variant: 'styles.article' }} />
```

Let me know if that makes sense

lachlanjc:
Ok, that makes sense. I think mostly I should reconsider what's in my `styles` & I can probably work something out. Thanks for the thoughts.

Issue #115, opened 2019-06-21
Looking for additional preset themes


jxnblk:
Theme UI includes a `@theme-ui/presets` package, which is intended to serve as a base for creating custom themes on top of. It'd be great to have more presets added to the package. Presets can be as minimal as a basic color palette and typographic styles, or include MDX element styles, syntax highlighting, or any other useful design tokens.

If you'd like to contribute a new preset, but aren't sure how to get started, feel free to leave a comment here, and we'll try to help you out

lemes:
@jxnblk do you think it would make sense to have a Material Design preset?

jxnblk:
@lemes yes, that would be great!

yurm04:
@jxnblk 👋 I added https://github.com/system-ui/theme-ui/pull/567 as a basic preset loosely based on the Polaris design system.  It's pretty simple right now so let me know if there's any interest in building out its typography or anything else.

TomFreudenberg:
Hi, hopefully I may post this question here.

We are preparing a new app with gatsby and are not sure to choose the right design fundament. Before looking on theme-ui we intend to go for material-ui.

We like the idea of theme-ui but as far as I get it right - if we would also like to use (something like) the material-ui components - there had to be a materialize theme and components library based on theme-ui. Currently existing themes and components for theme-ui are too primitive. Is this a correct assessment?

Another approach is discussed on that issue https://github.com/system-ui/theme-ui/issues/819 - is this a right one?

Thanks for your feedback
Tom

jxnblk:
@TomFreudenberg could you please open a new issue since your question is off topic for this issue?

beerose:
I recently did _sketchy_ Theme UI preset: https://www.npmjs.com/package/theme-ui-sketchy-preset.

Would you be interested in a PR that adds it to  the `@theme-ui/presets`?

jxnblk:
@beerose I *love* this preset! It's so good! We're certainly open for a PR; whether you want to link out to your own repo from the docs or add it to this one, I'll leave that up to you (i.e. it's totally fine if you want to keep this on your own GitHub or domain)

lachlanjc:
We’d still love for contributions of additional preset themes, but due to the lack of activity on these we’re closing for now.

PR #567, opened 2020-01-13
Add new theme preset based on Shopify's Polaris design system

yurm04:
Addresses Issue https://github.com/system-ui/theme-ui/issues/115 

[Polaris](https://polaris.shopify.com/design/colors) is a great design system from Shopify so here's a very basic theme preset to use as a jumping-off point.  If there's interest I can continue to further develop this preset to more closely match Polaris's typography.

Screen Shot 2020-01-12 at 7 53 41 PM


packages/preset-polaris/src/index.js
```diff
@@ -0,0 +1,43 @@
+import base from '@theme-ui/preset-base'
+
+const text = '"San Francisco", Roboto, "Segoe UI", "Helvetica Neue", sans-serif'
```
lachlanjc: "San Francisco" isn't the correct font name. Instead:
```suggestion
const text = 'system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", sans-serif'
```

packages/preset-polaris/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+export const polaris = {
+  ...base,
+  colors: {
+    text: '#454f5b',
+    background: '#fff',
+    primary: '#5C6AC4',
```
lachlanjc: These hex codes are capitalized inconsistently :) other themes seem to be using all-lowercase

packages/presets/src/index.js
```diff
@@ -11,3 +11,4 @@ export { default as tosh } from '@theme-ui/preset-tosh'
 export { default as bootstrap } from '@theme-ui/preset-bootstrap'
 export { default as bulma } from '@theme-ui/preset-bulma'
 export { default as tailwind } from '@theme-ui/preset-tailwind'
+export { default as polaris } from '@theme-ui/preset-polaris'
```
lachlanjc: Let's keep this list alphabetized as above—just move this above Tailwind

lachlanjc:
FYI the deploy preview seems to have failed for unrelated reasons—Yarn couldn't download `repng`. No idea what's happening there, it's still alive: https://www.npmjs.com/package/repng

yurm04:
thanks, @lachlanjc for the feedback! Think I addressed all of your initial suggestions

jxnblk:
Looks great! Thanks for putting this together!

PR #568, opened 2020-01-13
Use backwards-compatible font stack for base theme

lachlanjc:
You may disagree with me on this since most sites will override this anyway, but I think our default system font stack should be more backwards-compatible. I not-infrequently run into browsers that don't yet support `system-ui` (like older versions of macOS, iOS <11), & there are other system fonts we can use. Browsers are & will continue adopting `system-ui`, but it's at 85% globally with no support on IE or Edge (https://caniuse.com/#feat=font-family-system-ui), & this is a simple fix.


CanRau:
I, at least^^, totally agree as I always think about that when I stumble upon the font stack just without opening an issue 😅 👍

jxnblk:
Thanks! I think this makes total sense from an ease-of-use standpoint. Would you mind updating the other presets that use the `system-ui` font stack?

lachlanjc:
Done. Let me know if there's anywhere I shouldn't have changed—I mostly kept it off the docs site pages just for brevity but perhaps consistency is more important?

jxnblk:
I think for docs, keeping the example font stack shorter is probably less distracting. Thanks for putting this together!! :shipit:

jxnblk:
Looks like some snapshots are failing. Would you mind running `yarn test -u` to update those?

lachlanjc:
Oops, sorry I forgot about that. Done!

Issue #553, opened 2020-01-08
Site fails if LocalStorage is explicitly disallowed with option useColorSchemeMediaQuery


c-kirkeby:
**Describe the bug**

**To Reproduce**
(Similar issues in both Firefox v72 and Chrome v79.0.3945.88 Windows 10 64-bit)
1. Go to https://gatsby.norse.codes
2. Go to your settings to block LocalStorage and Cookies on the site)
*Firefox:* about:preferences#privacy > Manage Permissions, add the site https://gatsby.norse.codes and click 'Block' and 'Save Changes'.
*Chrome:* chrome://settings/content/cookies > Add and enter the site gatsby.norse.codes then click 'Add'
3. Go to https://gatsby.norse.codes and refresh.
The error will differ depending on the browser but they both complain about not being able to access LocalStorage.

Firefox:
```
Request to access cookies or storage on “<URL>” was blocked because of custom cookie permission. 4
Request to access cookies or storage on “https://gatsby.norse.codes/blog” was blocked because of custom cookie permission.
2 blog:1250:8
Request to access cookies or storage on “https://gatsby.norse.codes/blog” was blocked because of custom cookie permission.
blog:2153:40
Request to access cookies or storage on “https://gatsby.norse.codes/blog” was blocked because of custom cookie permission.
blog:2:19
Request to access cookies or storage on “https://gatsby.norse.codes/blog” was blocked because of custom cookie permission.
blog:1:31821
DOMException: "The operation is insecure." d618faa047230c1c52d10b908bd3b5a512e1c154.js:1:381107
    fl https://d33wubrfki0l68.cloudfront.net/bundles/d618faa047230c1c52d10b908bd3b5a512e1c154.js:1
    callback https://d33wubrfki0l68.cloudfront.net/bundles/d618faa047230c1c52d10b908bd3b5a512e1c154.js:1
    vi https://d33wubrfki0l68.cloudfront.net/bundles/d618faa047230c1c52d10b908bd3b5a512e1c154.js:1
    yi https://d33wubrfki0l68.cloudfront.net/bundles/d618faa047230c1c52d10b908bd3b5a512e1c154.js:1
    vu https://d33wubrfki0l68.cloudfront.net/bundles/d618faa047230c1c52d10b908bd3b5a512e1c154.js:1
    unstable_runWithPriority https://d33wubrfki0l68.cloudfront.net/bundles/d618faa047230c1c52d10b908bd3b5a512e1c154.js:1
[...]

PR #576, opened 2020-01-13
Fix uncaught localStorage exception

alexpermyakov:
This is a fix for [#553](https://github.com/system-ui/theme-ui/issues/553):

**SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.**


alexpermyakov:
@jxnblk it might look better with optional chaining enabled, not sure if it's worth it tough:
`
const getStorage = () => {
  try {
    return window.localStorage
  } catch (e) {}
}

const storage = {
  get: init => getStorage()?.getItem(STORAGE_KEY) || init,
  set: value => getStorage()?.setItem(STORAGE_KEY, value),
}
`

packages/theme-ui/src/color-modes.js
```diff
@@ -5,11 +5,18 @@ import { useThemeUI } from './context'
 import { createColorStyles } from './custom-properties'
 
 const STORAGE_KEY = 'theme-ui-color-mode'
-const HAS_STORAGE = typeof Storage !== 'undefined'
+const hasStorage = () => {
```
jxnblk: Thanks! I suspect it would be better to keep this as a constant rather than calling a function for every time this stores something in `localStorage`. Does a check for the `window` object not work in the environment you're seeing this in?

alexpermyakov: Thank you for the quick response! Yep, it's not enough, because if you block cookies and site data (chrome://settings/content/cookies) then just an attempt of accessing localStorage will throw an exception (SecurityError).

I don't it as a big issue - even though it's a function call - it's indistinguishable in terms of performance comparing to window.localStorage. It takes the way more time to actually save the string inside the storage.

alexpermyakov: Another way we can do it - wrap up window.localStorage into try/catch for both get/set methods. Then we don't need this hasStorage check at all.

jxnblk: I like the idea of the whole thing being wrapped in a try catch -- it could still show a warning if localStorage fails for some reason, right?

alexpermyakov: Yes, we can do something like this:

```
const storage = {
  get: init => {
    try {
      return window.localStorage.getItem(STORAGE_KEY) || init
    } catch (e) {
      console.warn(e)
    }
  },

  set: value => {
    try {
      window.localStorage.setItem(STORAGE_KEY, value)
    } catch (e) {
      console.warn(e)
    }
  },
}
```

This try/catch inside each method is kinda ugly but simple and probably won't get more complex.

alexpermyakov: The warning message in the console will be:
_DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document._

or we can use something more custom like:
_localStorage is disabled and theming might not work as expected. Please check your Site Settings._

jxnblk: Yeah, I like this because it would also fail silently if *anything* else happens to not work with localStorage

alexpermyakov: Updated. I'm fine with the warning message duplication (React itself doest it a lot) or we can put it as a const (something line STORAGE_WARNING_MESSAGE)

alexpermyakov: Probably better to add a real error as well, like this:

console.warn(
    'localStorage is disabled and theming might not work as expected.',
    'Please check your Site Settings.',
    e
  )

fyi @jxnblk

jxnblk: The only thing localStorage is used for is keeping the color mode across sessions. Theming should still work if localStorage is disabled

alexpermyakov: So, 'localStorage is disabled and color mode might not work as expected.' then?

jxnblk: I think that sounds fine -- otherwise, this looks ready to go. Would be nice to have a test for this, but no worries if you don't have time

alexpermyakov: I'll try to create a test. It might be harder than usual - Jest mocks a method and we can throw an exception as an implementation. However, in our case, we have to throw an error during property access and not a function call. I think [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) can solve this case. Let me try.

alexpermyakov: I've added a test. Finally found a practical use-case for [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) 😃

packages/theme-ui/test/color-modes.js
```diff
@@ -518,3 +518,41 @@ test('raw color values are passed to theme-ui context when custom properties are
   )
   expect(color).toBe('tomato')
 })
+
+test('warns when localStorage is disabled', () => {
+  const originalWindow = window
+
+  const mockWindow = new Proxy(window, {
```
jxnblk: What's the reason for using this instead of `jest.fn()`?

alexpermyakov: Indeed, Object.defineProperty with a custom descriptor is enough. No need for Proxy.
Fixed.

Issue #403, opened 2019-10-06
Creating Multiple Variants


moshemo:
I am looking for a way to create multiple variants for buttons.  For example, let's say I want to have the following three variants:

1. Size (small, medium, large, etc). 
2. Look (primary, secondary, disabled, etc).  
3. Kind (default, outline, text, etc).

So that I could style a button like this:

```
<button
 sx={{
    size="buttons.size.large"
    kind="buttons.kind.outline"
    look="buttons.look.secondary"
 }}
/>
```

This IS possible with style system -- here is an example:

* https://mathanpec.github.io/component-library/?path=/story/button--size-variant
* https://github.com/mathanpec/component-library/blob/basic-setup/src/theme/buttonStyle.ts

But I can't figure out how to do with with Theme-UI.

Is this possible with Theme-UI?  If so, can how can I do it?

Thanks.

timkindberg:
Seems like an array could work, where it would just `merge` the variants.

```
sx={{
  variant: [
    'buttons.size.large',
    'buttons.kind.outline',
    'buttons.look.secondary'
  ]
}}
```

jxnblk:
Currently the `variant` property only accepts a single argument, but it's something I've considered. @timkindberg the array syntax would allow you to use different variants at different breakpoints

moshemo:
Okay, thanks for letting me know.

jferrettiboke:
@jxnblk Any solution to this? I have the same need. Could you consider to re-open this issue?

jferrettiboke:
I get it working doing this. However, it would be cooler to support this internally.

```js
import { useResponsiveValue, useBreakpointIndex } from "@theme-ui/match-media";
import get from "lodash.get";
import theme from "../theme";

function variants(...variantBlocks) {
  const variants = useResponsiveValue(variantBlocks);
  let styles = {};
  variants.map(variant => {
    styles = { ...styles, ...get(theme, variant) };
  });
  return styles;
}

function Example() {
  return (
    <Button
      sx={{
        ...variants(
          ["buttons.colors.secondary", "buttons.sizes.xs"],
          ["buttons.colors.primary", "buttons.sizes.lg"]
        )
      }}
    >
      Breakpoint index: {index}
    </Button>
  );
}
```

CanRau:
How about comma separated strings? @jxnblk 

Merry Christmas to y'all 🎅 🎉

Pirastrino:
[...]

PR #586, opened 2020-01-17
Add support for combining multiple variants

lachlanjc:
Inspired by #403, I've added support for combining multiple variants by passing `variants` an array. It's a new prop on components, or a property inside `sx`.

- Docs & tests are included for both `components` & `css`.
- The new test on `css` is failing because something about my implementation in `css` doesn't work—to be honest, I have no idea what's wrong & would super appreciate some assistance there.
- Targeting v0.3 because I don't want to create more work with upgrading this :)
- Added to changelog

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -14,6 +14,7 @@
     "@styled-system/color": "^5.1.2",
     "@styled-system/should-forward-prop": "^5.1.2",
     "@styled-system/space": "^5.1.2",
+    "@theme-ui/core": "^0.3.1",
     "@theme-ui/css": "^0.3.1"
   },
   "peerDependencies": {
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -14,8 +14,8 @@
     "@styled-system/color": "^5.1.2",
     "@styled-system/should-forward-prop": "^5.1.2",
     "@styled-system/space": "^5.1.2",
-    "@theme-ui/core": "^0.3.1",
-    "@theme-ui/css": "^0.3.1"
+    "@theme-ui/css": "^0.3.1",
+    "deepmerge": "^4.2.2"
   },
   "peerDependencies": {
     "react": "^16.8.0"
--- a/packages/components/test/index.js
+++ b/packages/components/test/index.js
@@ -147,6 +147,27 @@ describe('Box', () => {
     expect(json).toHaveStyleRule('color', 'white')
   })
 
+  test('renders with variant and variants prop', () => {
+    const json = renderJSON(
+      <ThemeProvider theme={theme}>
+        <Box variant="text.heading" variants={['boxes.beep', 'boxes.boop']} />
+      </ThemeProvider>
+    )
+    expect(json).toHaveStyleRule('font-size', '32px')
+    expect(json).toHaveStyleRule('padding', '32px')
+    expect(json).toHaveStyleRule('background-color', 'tomato')
+    expect(json).toHaveStyleRule('color', 'white')
+  })
+
+  test('renders with incorrect type of variants prop', () => {
+    const json = renderJSON(
+      <ThemeProvider theme={theme}>
+        <Box variants="boxes.beep" />
+      </ThemeProvider>
+    )
+    expect(json).toMatchSnapshot()
+  })
+
   test('renders with base styles', () => {
     const json = renderJSON(
       <Box
--- a/packages/components/test/index.js
+++ b/packages/components/test/index.js
@@ -48,6 +48,10 @@ const theme = {
       p: 4,
       bg: 'highlight',
     },
+    boop: {
+      bg: 'tomato',
+      color: 'white',
+    },
   },
   cards: {
     primary: {
@@ -130,6 +134,17 @@ describe('Box', () => {
     )
     expect(json).toHaveStyleRule('background-color', 'highlight')
     expect(json).toHaveStyleRule('padding', '32px')
+  })
+
+  test('renders with variants prop', () => {
+    const json = renderJSON(
+      <ThemeProvider theme={theme}>
+        <Box variants={['boxes.beep', 'boxes.boop']} />
+      </ThemeProvider>
+    )
+    expect(json).toHaveStyleRule('padding', '32px')
+    expect(json).toHaveStyleRule('background-color', 'tomato')
+    expect(json).toHaveStyleRule('color', 'white')
   })
 
   test('renders with base styles', () => {
--- a/packages/css/test/index.js
+++ b/packages/css/test/index.js
@@ -230,6 +230,21 @@ test('handles responsive variants', () => {
   })
 })
 
+test('returns multiple variants from theme', () => {
+  const result = css({
+    variants: ['text.caps', 'text.title'],
+  })(theme)
+  expect(result).toEqual({
+    fontSize: 16,
+    letterSpacing: '-0.01em',
+    textTransform: 'uppercase',
+    '@media screen and (min-width: 40em)': {
+      fontSize: 16,
+      letterSpacing: '-0.02em',
+    },
+  })
+})
+
 test('handles negative margins from scale', () => {
   const result = css({
     mt: -3,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Adjust media query sort logic #600
 - Fixed link to Gatsby Plugin page in `getting-started` page. Issue #602
+- Adds support for combining multiple variants (pass `variants` an array in `sx` or as a prop on components)
 
 ## v0.3.0 2019-01-22
 
--- a/packages/components/src/Box.js
+++ b/packages/components/src/Box.js
@@ -1,9 +1,9 @@
 import styled from '@emotion/styled'
 import { css, get } from '@theme-ui/css'
-import { merge } from '@theme-ui/core'
 import { createShouldForwardProp } from '@styled-system/should-forward-prop'
 import space from '@styled-system/space'
 import color from '@styled-system/color'
+import deepmerge from 'deepmerge'
 
 const shouldForwardProp = createShouldForwardProp([
   ...space.propNames,
@@ -12,10 +12,17 @@ const shouldForwardProp = createShouldForwardProp([
 
 const sx = props => css(props.sx)(props.theme)
 const base = props => css(props.__css)(props.theme)
+
 const variant = ({ theme, variant, __themeKey = 'variants' }) =>
   css(get(theme, __themeKey + '.' + variant, get(theme, variant)))
-const variants = ({ theme, variants = [], __themeKey = 'variants' }) =>
-  css(merge(variants.map(v => get(theme, __themeKey + '.' + v, get(theme, v)))))
+const variants = ({ theme, variants = [], __themeKey = 'variants' }) => {
+  if (!Array.isArray(variants)) return {}
+  return css(
+    deepmerge.all(
+      variants.map(v => get(theme, __themeKey + '.' + v, get(theme, v)))
+    )
+  )
+}
 
 export const Box = styled('div', {
   shouldForwardProp,
--- a/packages/components/src/Box.js
+++ b/packages/components/src/Box.js
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled'
 import { css, get } from '@theme-ui/css'
+import { merge } from '@theme-ui/core'
 import { createShouldForwardProp } from '@styled-system/should-forward-prop'
 import space from '@styled-system/space'
 import color from '@styled-system/color'
@@ -13,6 +14,8 @@ const sx = props => css(props.sx)(props.theme)
 const base = props => css(props.__css)(props.theme)
 const variant = ({ theme, variant, __themeKey = 'variants' }) =>
   css(get(theme, __themeKey + '.' + variant, get(theme, variant)))
+const variants = ({ theme, variants = [], __themeKey = 'variants' }) =>
+  css(merge(variants.map(v => get(theme, __themeKey + '.' + v, get(theme, v)))))
 
 export const Box = styled('div', {
   shouldForwardProp,
@@ -23,6 +26,7 @@ export const Box = styled('div', {
     minWidth: 0,
   },
   base,
+  variants,
   variant,
   space,
   color,
--- a/packages/components/test/__snapshots__/index.js.snap
+++ b/packages/components/test/__snapshots__/index.js.snap
@@ -180,6 +180,18 @@ exports[`Box renders 1`] = `
 </div>
 `;
 
+exports[`Box renders with incorrect type of variants prop 1`] = `
+.emotion-0 {
+  box-sizing: border-box;
+  margin: 0;
+  min-width: 0;
+}
+
+<div
+  className="emotion-0"
+/>
+`;
+
 exports[`Button renders 1`] = `
 .emotion-0 {
   box-sizing: border-box;
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -8,6 +8,9 @@
     "prepare": "microbundle --no-compress",
     "watch": "microbundle watch --no-compress"
   },
+  "dependencies": {
+    "deepmerge": "^4.2.2"
+  },
   "author": "Brent Jackson",
   "license": "MIT",
   "publishConfig": {
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -1,3 +1,5 @@
+import { all } from 'deepmerge'
+
 export const get = (obj, key, def, p, undef) => {
   key = key && key.split ? key.split('.') : [key]
   for (p = 0; p < key.length; p++) {
@@ -188,6 +190,12 @@ export const css = args => (props = {}) => {
     if (key === 'variant') {
       const variant = css(get(theme, val))(theme)
       result = { ...result, ...variant }
+      continue
+    }
+
+    if (key === 'variants' && typeof val === 'array') {
+      const variants = css(all(val.map(v => get(theme, v))(theme)))
+      result = { ...result, ...variants }
       continue
     }
 
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -193,8 +193,8 @@ export const css = args => (props = {}) => {
       continue
     }
 
-    if (key === 'variants' && typeof val === 'array') {
-      const variants = css(all(val.map(v => get(theme, v))(theme)))
+    if (key === 'variants') {
+      const variants = css(get(theme, val))(theme)
       result = { ...result, ...variants }
       continue
     }
--- a/packages/docs/src/pages/components/variants.mdx
+++ b/packages/docs/src/pages/components/variants.mdx
@@ -64,6 +64,17 @@ For example, a secondary button style can be added to `theme.buttons.secondary` 
 </Button>
 ```
 
+## Combining Multiple Variants
+
+If you want to combine multiple variants on the same element, pass an array of variants with the `variants` prop.
+If styles conflict, those of the variant passed later will render.
+
+```jsx live=true
+<Text variants={['caps', 'heading']}>
+  Title text
+</Text>
+```
+
 ## Example Theme
 
 ```js
--- a/packages/docs/src/pages/guides/variants.mdx
+++ b/packages/docs/src/pages/guides/variants.mdx
@@ -29,6 +29,9 @@ For example, you can define `primary` and `secondary` variants for buttons and u
       color: 'white',
       bg: 'secondary',
     },
+    small: {
+      fontSize: 1
+    }
   },
 }
 ```
@@ -50,6 +53,15 @@ When using the built-in [`Button` component](/components/button), you can set th
 Variants can use any name you choose, and deeply nested objects can be accessed with dot notation.
 
 </Note>
+
+## Combining Multiple Variants
+
+If you want to combine multiple variants on the same element, pass an array to `variants` in the `sx` prop.
+If styles conflict, those of the variant passed later will render.
+
+```jsx
+<button sx={{ variants: ['buttons.primary', 'buttons.small'] }} />
+```
 
 ## Color Modes
 


packages/components/package.json
```diff
@@ -13,7 +13,8 @@
     "@styled-system/color": "^5.1.2",
     "@styled-system/css": "^5.0.23",
```
jxnblk: Looks like this hasn't been updated in the `css-package` branch yet...

packages/components/package.json
```diff
@@ -13,7 +13,8 @@
...
     "@styled-system/should-forward-prop": "^5.1.2",
-    "@styled-system/space": "^5.1.2"
+    "@styled-system/space": "^5.1.2",
+    "deepmerge": "^4.2.2"
```
jxnblk: You can use the `merge` utility from `@theme-ui/core` for this instead

packages/components/test/index.js
```diff
@@ -48,6 +48,10 @@ const theme = {
       p: 4,
       bg: 'highlight',
     },
+    boop: {
```
jxnblk: 👍

packages/css/test/index.js
```diff
@@ -230,6 +230,21 @@ test('handles responsive variants', () => {
   })
 })
 
+test('returns multiple variants from theme', () => {
```
jxnblk: I'd suggest adding a few more tests for when the value of `variants` *isn't* an array and maybe one to demonstrate how using both `variant` and `variants` together behaves

packages/components/package.json
lachlanjc: I tried this out but it broke the code. I've switched back to deepmerge for the time being.

packages/components/package.json
lachlanjc: Rebased!

medelman17:
This would be amazing. But maybe even better if we could tackle this issue through a ‘styled-components-modifiers’-like ‘modifiers’ property to keep the API clean/distinct? 

On one hand, having to repeatedly deep merge in your mind while developing is NOT fun. On the other, I’d imagine that 99% of the time the override-variant will be making a minor change to the styling—otherwise one would just have another ‘no-help-needed’ variant defined. 

So, to me, it feels like a we need a little door to reach through (should the need arise) and make small tweaks in an otherwise manageable way. In my mind, it would look something like this: 

<Text sx={{ variants: “text.h1” modifiers=[‘bold’, ‘allCaps’, ‘primaryColor’, ‘etc.’] }} > ... 

I dunno; just spit ballin’. But would be very happy to see current PR land

lachlanjc:
Sorry, could you clarify exactly what the difference you're proposing with the `modifers` property is? A bit confused how that's different.

medelman17:
I guess, in my mind, it’s not different in the result. Which is why I think this PR is great! I just think the ergonomics are tougher when everything goes through one property. So, it’s different in execution, I guess—for my mental model, thinking of the ‘variant’ as the dominant style and the ‘modifiers’ as tweaks makes a lot of sense. I feel like I can see clearer lines of demarcation. Theming is hard; especially for beginners. I’m just throwin’ it out there.

lachlanjc:
Hmm, that makes a little more sense. The example I use in the docs is title + caps, where they’re not overlapping, just combining. In practice styles will definitely overlap but I think it’s kind of the same thing as adding several classes in traditional CSS—but actually simpler because you don’t have a specificity battle after that.

CanRau:
hum I'd still vote for `variants`.
I think I  get the idea of `modifiers` yet it feels kinda limiting and unnecessarily adds a new terminology 🤔

dakebl:
> hum I'd still vote for `variants`.
> I think I get the idea of `modifiers` yet it feels kinda limiting and unnecessarily adds a new terminology 🤔

Yeah, I agree. I like that pluralising `variant` makes it simple to understand what is expected, multiple `variants`.

brunnolou:
This looks cool at first glance, but I think it is conceptually wrong 😞

## 1. Invalid variants

For example, this is proposed here:

```js
  buttons: {
    // This alone is a button variant
    primary: {
      color: 'white',
      bg: 'primary',
    },
    // This alone is a button variant
    secondary: {
      color: 'white',
      bg: 'secondary',
    },
    // This alone is NOT a button variant! 😞
    small: {
      fontSize: 1
    }
  },
```

The `buttons.small` alone **is NOT** a button variant.

Imagine in the future a design tool reading all buttons to display them, and then we have **an unstyled `button.small`**...

I agree with @medelman17 when he said:

> thinking of the ‘variant’ as the dominant style and the ‘modifiers’ as tweaks

## 2. Array API and breakpoints

Using arrays is "kinda reserved" to breakpoints, @jxnblk mentioned here at https://github.com/system-ui/theme-ui/issues/403#issuecomment-539575087

Could be confused with the future `variant={['text.body', 'text.heading']}`
Having nested arrays to solve this could be even more confusing.

## 3. Similar but competing APIs

What to expect when combining both `variant` ad `variants` 🤷‍♂.

## 4. Allows breaking the design system

This very similar to class names. Meaning that everything is now possible, even invalid combinations:

```js
sx={{ variants: ['buttons.primary', 'cards.compact'] }}`
```
To be honest, I don't think we should allow multiple variants on the `sx` prop.
If we are to do this, we might want only to allow in the theme spec to guarantee the correct later usage.

# My proposal

1. Object notation instead.
2. Add Modifiers (the name could be something different like: `variantModifiers`, `variantTypes`, `addons`, or even `variants`).
3. Group them inside the variants.

```js
buttons: {
  primary: {
    color: 'white',
    bg: 'primary',
  },
  secondary: {
    color: 'white',
    bg: 'secondary',
  },
  modifiers: {
    size: {
      small: { fontSize: 1 },
    },
    shape: {
      rounded: { borderRadius: 4 },
      pill: { borderRadius: 100 },
    },
  },
},
```

### Usage

```js
sx={{
  variant: 'buttons.primary'
  modifiers: {
    size: 'small',
    shape: 'pill'
  }
}}
```

### Creating variants with modifiers

We could abstract even more things to modifiers like colors and create a variant out of modifiers:

```js
  primary: {
    modifiers: { color: 'primary', shape: 'rounded' },
  },
  secondary: {
    modifiers: { color: 'secondary', shape: 'rounded' },
  },
  pill: {
    modifiers: { color: 'orange', shape: 'pill', size: 'small' },
  }
  modifiers: {
    color: {...},
    size: {...},
    shape: {...},
  }
```

By doing that we can explicitly declare named variants and have the ability to modify then in an acceptable way.
Not by allowing creating invalid variant combinations like `variants: ['primary', 'secondary']`.

## Responsiveness

```js
sx={{
  variant: 'buttons.primary'
  modifiers: [{ size: 'small' }, { size: 'large' }]
}}
```

## Advantages
- Scoped to the current variant
- Keep the design system consistent
- Do not allow invalid combinations
- Better responsiveness

cour64:
Although I like your proposed solution @brunnolou. I'm still in favour of the multiple variants prop as it's much simpler to use.

Not sure where the distinction of modifiers/tweaks and variants is but to me anything that changes a style on a component creates a stylistic variation of that component and can thus be called a variant. If you have a ```small``` variant and a ```primary``` variant no matter the styles each applies they still create style variations of the button. Anyway that's my understanding of the nomenclature.

This syntax, for me at least, is easiest to grok:

```javascript
sx={{
  variants: ['buttons.color.primary', 'buttons.size.small', 'buttons.shape.pill']
}}
```
I do agree with @brunnolou that there could be some confusion with what to expect when the ```variant``` and ```variants``` props are both used concurrently.

brunnolou:
> ```js
> sx={{
>   variants: ['buttons.color.primary', 'buttons.size.small', 'buttons.shape.pill']
> }}
> ```

### Theme as a _stylesheet_ and variants as _classNames_ 🤯

@cour64 one of my biggest issues with this is the power. it would easily be misused. 
Nothing will prevent anyone using the **theme as a stylesheet** and `sx={{ variants: […] }}` as **classNames**! 
All the design system constraints could be lost.

We could see something like this:
> ```js
> sx={{
>   variants: ['buttons.primary', 'utilClasses.myCustomStyle', 'utilClasses.becauseICan']
> }}
> ```

### Difference between tweak and variants

> Not sure where the distinction of modifiers/tweaks and variants is but to me anything that changes a style on a component creates a stylistic variation of that component and can thus be called a variant. 

Following the BEM methodology, a "tweak" is the modifier:
> _"defines the appearance, state, or behavior of a block or element."_

A block or element in our context it's our components. IMO each variant alone defines the whole style of the component.
- `sx={{ variants:"buttons.primary" }}` and `sx={{ variants:"buttons.secondary" }}` should render a button.
- In your example if you define just the size variant, for example, then `variants: ['buttons.size.small']` will render an **unstyled** small button.

### Reinforcing the design system

So this is how I believe we could reinforce the design system constraints:
- A **variant** defines the main style and can be used alone
- A **variant** can be extended or modified with it owns **modifiers**
- A **modifier** shouldn’t be used alone without a variant

jonavila:
Since emotion supports composition by passing an array to the `css` prop, I opened a PR to revive the work from #174 to allow the same composition in the `sx` prop. The new PR is #704. If we allow arrays in `sx`, we can compose the variants that way. (https://emotion.sh/docs/composition).

At the moment, I have to use the `css` prop and wrap the array of objects with the `css` utility from theme-ui.


In the app I'm working on, I also have a one-off case where I need to compose various variants into a single variant. The workaround I've been using at the moment is by leveraging Emotion's `&` selector for the current class and combining variants like:

```
{
    colors: {
        primary: 'blue'
    },
    button: {
        primary: {
            bg: 'primary'
        }
    },
    text: {
        caps: {
            textTransform: 'uppercase',
        },
        muted: {
            color: 'gray1',
        },
    },
   {
    someSuperVariant: {
        variant: 'button.primary',
        '&': {
            variant: 'text.caps',
        },
    },       
   }
}
```

medelman17:
For what it’s worth, I published a ‘plugin’ of sorts that enables the above, BEM-like functionality by overloading either the sx or css prop. Find it 👉🏻 https://github.com/fabulascode/theme-ui-modifiers 

It’s early stages yet—edge cases haven’t been considered, etc.—but I like it’s functionality, the ergonomics, imho, are a better fit for my mental model—despite its obvious hacky-ness due to not being officially supported. 🤷🏻‍♂️

Here’s how it works generally (you can also use the css prop):

```
/** @jsx jsx */
import { jsx, Text as ThemeUIText } from 'theme-ui';
import { useModifiers, getVariant } from '@fabulas/theme-ui-modifiers';

const TextModifiers = {
  bold: { fontWeight: 700 },
  orange: { color: 'orange' },
  bigger: ({ theme, variant }) => {
    const { fontSize } = getVariant(theme, variant);
    if (Array.isArray(fontSize)) {
      return { fontSize: fontSize.map(s => s + 1) };
    }
    return { fontSize: fontSize + 1 };
  },
};

function Text({ children, ...props }) {
  const modifiers = useModifiers(props);

  return (
    <ThemeUIText variant={props.variant} sx={modifiers}>
      {children}
    </ThemeUIText>
  );
}

function App() {
  return (
    <div>
      <Text variant="h1" modifiers={['bigger', 'bold', 'orange']}>
        Hello!
      </Text>
    </div>
  );
} 
```

Would love to hear your thoughts and or ideas.

brunnolou:
@medelman17 I that's interesting, I like the `useModifiers` idea. I've just found two issues here.

## Responsiveness
Like I mentioned before IMO using object notation would be better to be compatible with responsiveness. 
E.g.: `<Text variant="h1"  modifiers={[{ weight: 'bold' }, { weight: 'normal' }]}`.

## Theming 
Keeping the modifiers at the component level we loose "themeability". Meaning that even if we change the theme, the modifiers will always remain the same.

## Suggestion

Using your idea maybe something like this could tackle these two issues:

1. Add `modifiers` to the theme
```js
// theme.js
buttons: {
  primary: { ...  },
  secondary: { ...  },
  modifiers: {
    size: {
      small: { fontSize: 1 },
      large: { fontSize: 3 },
    },
    shape: {
      rounded: { borderRadius: 4 },
      pill: { borderRadius: 100 },
    },
  },
},
```

2. Get modifiers from the theme by variant and modifiers names.
```jsx
function Button({ variant = 'primary', children, ...props }) {
  // By variant and modifiers names.
  const modifiers = useModifiers(`buttonns.${variant}`, props.modifiers);

  return (
    <ThemeUIButton variant={props.variant} sx={modifiers}>
      {children}
    </ThemeUIButton>
  );
}
```

```tsx
<Button modifiers={{ shape: 'pill' }} />
<Button variant="secondary" modifiers={[{ size: 'small' }, { size: 'large' }]} />
```
That would be the general usage, but if we want a better API for the componensts, we could create a different Buttons component implementation to map props to modifiers (before passing to `useModifiers`), to have sth liked this:

```tsx
<Button  variant="secondary" shape="pill" size="small" />
```

medelman17:
So, if you take a bit of a closer look, I think you’ll see that the approach I propose use object notation and solves for both responsiveness and theming—see the ‘bigger’ modifier.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

brunnolou:
@jonavila I like the idea of having arrays in the `sx` prop.

> The new PR is #704. If we allow arrays in `sx`, we can compose the variants that way.

Are you saying that the variants should also be merged?
What we would expect from something like these?
```js
<Box sx={[{ variant: 'buttons.primary' }, { variant: 'buttons.seconday' }]}
```
Replace the `primary` variant with `secondary` or merge them? 🤔 

If merged, then this would be possible (even though looks a bit weird):

```js
<Box sx={[{ variant: 'buttons.primary' }, { variant: 'buttons.sizes.large' }, { variant: 'text.caps' }]}
```

Regarding the need for extending multiple variants, I can see the benefit, especially while defining the `theme`.
But like I mentioned https://github.com/system-ui/theme-ui/pull/586#issuecomment-590825213 it's too powerful and can break the design system.

## Conclusion

I think both ideas are valid and coexist.

1. Being able to extend *different* variants (theme only 🤔?):
```js
// theme.js
buttons: {
    primary: { ... },
    attention: {
        variants: ['button.primary', 'text.caps'],
        // Or
        extends: ['button.primary', 'text.caps'],
    },
}
```

2. Add modifiers for the same variant
```js
// theme.js
buttons: {
  primary: {
    modifiers: { color: 'primary', shape: 'rounded' },
  },
  attention: {
    extends: ['button.primary', 'text.caps'],
    modifiers: { size: 'large' },
  },
  modifiers: {
    color: { ... },
    size: { ... },
    shape: { ... },
  },
}
```

I think we should make clear this distinction: "multiple variants" vs "modifiers".
I'd love to know @jxnblk opinion about all this.

cour64:
@lachlanjc I think the problem with the implementation in the css package has to do with the handling of responsive styles here:

https://github.com/system-ui/theme-ui/blob/cf5b00ad01a59b215b1ff6ef0cbbbf9893025a9e/packages/css/src/index.ts#L206

where the variants array is being treated as a responsive style and only the first element of the ```variants``` prop is being treated as the base style and then all subsequent elements get nested in media queries. 

Perhaps checking the key within the ```responsive``` function so as to properly handle the ```variants``` prop could be a solution, albeit a bit naive? Something a long the lines of:

```javascript
const responsive = styles => theme => {
  const next = {}
  const breakpoints =
    (theme && (theme.breakpoints as string[])) || defaultBreakpoints
  const mediaQueries = [
    null,
    ...breakpoints.map(n => `@media screen and (min-width: ${n})`),
  ]

  for (const key in styles) {
    const value =
      typeof styles[key] === 'function' ? styles[key](theme) : styles[key]

    if (value == null) continue

    if (!Array.isArray(value)) {
      next[key] = value
      continue
    }

    // check if the key is 'variants' and handle appropriately
    if (key === 'variants') {
      for (let i = 0; i < value.length; i++) {
        const variant = flatten([
          typeof value[i] === 'function' ? value[i](theme) : value[i],
        ])

        if (variant == null) continue

        if (!Array.isArray(variant)) {
          next[key] = [...(next[key] || []), variant]
          continue
        }

       // this handles responsive variants of with a 2d array so ['primary', ['small', 'medium']]
        for (let i = 0; i < variant.slice(0, mediaQueries.length).length; i++) {
          const media = mediaQueries[i]
          if (!media) {
            next[key] = [...(next[key] || []), variant[i]]
            continue
          }
          next[media] = next[media] || {}
          if (variant[i] == null) continue
          next[media][key] = [...(next[media][key] || []), variant[i]]
        }
      }
      continue
    }

    for (let i = 0; i < value.slice(0, mediaQueries.length).length; i++) {
      const media = mediaQueries[i]
      if (!media) {
        next[key] = value[i]
        continue
      }
      next[media] = next[media] || {}
      if (value[i] == null) continue
      next[media][key] = value[i]
    }
  }

  return next
}
```

and then applying the variants prop in the ```css``` function like so:

```javascript
if (key === 'variants') {
  // wrap val in an array and then flatten in case a single string is passed
  const variants = css(
    deepmerge.all(flatten([val]).map(v => get(theme, v)))
  )(theme)
  result = { ...result, ...variants }
  continue
}
```

The ```flatten``` function here would take a deeply nested array and flatten it out indefinitely to be a 1d array, this would allow defining responsive variants by using a 2d array as well as handle the case where a single string is passed instead of an array of variants.

There might be a couple specificity issues with this implementation though.

jonavila:
> @jonavila I like the idea of having arrays in the `sx` prop.
> 
> > The new PR is #704. If we allow arrays in `sx`, we can compose the variants that way.
> 
> Are you saying that the variants should also be merged?
> What we would expect from something like these?
> 
> ```js
> <Box sx={[{ variant: 'buttons.primary' }, { variant: 'buttons.seconday' }]}
> ```
> 
> Replace the `primary` variant with `secondary` or merge them? 🤔
> 
> If merged, then this would be possible (even though looks a bit weird):
> 
> ```js
> <Box sx={[{ variant: 'buttons.primary' }, { variant: 'buttons.sizes.large' }, { variant: 'text.caps' }]}
> ```
> 
> Regarding the need for extending multiple variants, I can see the benefit, especially while defining the `theme`.
> But like I mentioned [#586 (comment)](https://github.com/system-ui/theme-ui/pull/586#issuecomment-590825213) it's too powerful and can break the design system.

@brunnolou Yes, they will be merged because behind the scenes the sx prop is using emotion's css prop. So it will work exactly like the last example on this page: https://emotion.sh/docs/composition

In that example, the composition overrides just the color property, but it leaves the background-color untouched. Per their docs:

>Using Emotion’s composition, you don’t have to think about the order that styles were created because the styles are merged in the order that you use them.

jxnblk:
Thanks for bringing up some of these ideas, all, but this level of discussion should probably happen in an issue rather than this PR. The proposal here is to add an additional `variants` property that can deeply merge multiple variants. The notion of *modifiers* from other CSS methodologies is identical to *variants*, and I do not think it's worth adding another name for this. The proposal for using arrays in the `sx` prop would *not* merge variants, and should be considered a separate issue. We'd like to ship this current proposal, but if you have additional thoughts outside the scope of this PR, please create a new issue

cour64:
The solution I proposed here https://github.com/system-ui/theme-ui/pull/586#issuecomment-591462005 works for the most part but there's one issue I noticed. 

If a variant is used within a variant e.g.

```js
primary: {
  p: 3,
  fontWeight: 'bold',
  color: 'white',
  bg: 'primary',
  borderRadius: 2,
},
disabled: {
  variant: 'buttons.primary',
  bg: 'muted',
}
```

Then some style props could incorrectly be overwritten e.g. the output above merged would result in:

```js
{
   p: 3,
   fontWeight: 'bold',
   color: 'white',
   bg: 'muted',
   borderRadius: 2,
   variant: 'buttons.primary'
}
```

the `bg` prop defined in the `disabled` variant would be overwritten by the styles props from `primary` variant despite it being specified before the `bg` prop. This is due to how the `deepmerge` package merges two objects and updates props in-place instead of keeping the order of props from the source object. There's actually an issue about it here https://github.com/TehShrike/deepmerge/issues/169 although I don't think `deepmerge` is the culprit with regards to the style props being overwritten.

Hope this all makes sense.

lachlanjc:
> Thanks for bringing up some of these ideas, all, but this level of discussion should probably happen in an issue rather than this PR. The proposal here is to add an additional `variants` property that can deeply merge multiple variants. The notion of _modifiers_ from other CSS methodologies is identical to _variants_, and I do not think it's worth adding another name for this. The proposal for using arrays in the `sx` prop would _not_ merge variants, and should be considered a separate issue. We'd like to ship this current proposal, but if you have additional thoughts outside the scope of this PR, please create a new issue

Thanks for the input @jxnblk, I stopped working on this since it seemed like it might go to waste. Will try to look back at getting this toward the finish line sometime soon. Even deciding that we don’t want to move forward with some of the other proposals here, glad we discussed them!

dburles:
I like some of the ideas with this but I definitely think it needs to be fleshed out a bit! It's a big conceptual change to the meaning of a variant, and I think it's overloading the variant API a bit unnecessarily. I don't think there is any reason why these kinds of variants couldn't just be handled in the component itself via props.

I quite like the modifiers suggestion by @brunnolou https://github.com/system-ui/theme-ui/pull/586#issuecomment-591406029 and funnily enough the `useModifiers` hook looks *exactly* the same as what I implemented for variant prop idea over here: PR #823 (and proposal #800). So I think these changes together could potentially work well to support this idea.

The interesting thing is you can already implement most of this already outside of theme-ui and basically ignore the `variant` CSS property. The only part missing there is the ability to easily apply variant styles to *any* element, e.g. 

```jsx
<button sx={{ variant: 'buttons.primary' }}>...</button>
```

In this case the core library requires a change to support the `variant` prop instead.

```jsx
<button variant="buttons.primary">...</button>
```
@brunnolou 
> That would be the general usage, but if we want a better API for the componensts, we could create a different Buttons component implementation to map props to modifiers (before passing to useModifiers), to have sth liked this:

```jsx
<Button variant="secondary" shape="pill" size="small" />
```

Perhaps the combination of these hooks would be perfect:
```jsx
const Button = ({
  variant,
  shape,
  size,
  ...props
}) => {
  const variantStyle = useVariant('buttons', variant);
  const modifierStyle = useModifiers('buttons', { shape, size });

  return ...;
});
```

Also this would allow you to easily style multiple elements with the `modifierStyle` too! 🎉. (See #823 for example with the `variant` prop and `useVariant` hook).

dburles:
Here's an example implementation of the idea above to play around with: https://codesandbox.io/s/theme-ui-hook-api-example-9nnmu?file=/src/index.js

lachlanjc:
Closing since there are many other PRs for this functionality now, & the files have changed since the PR was made

Issue #388, opened 2019-09-25
Component level themes and container query support


junyper:
Thoughts on supporting an API something like:

```
/** @jsx jsx */
import { jsx, useTheme } from 'theme-ui'

function SomeText (props) {
  const defaultTheme = {
    breakpoints: ['40em', '56em'],
    fontWeight: 'bold',
    fontSize: [ 4, 3 ],
    color: 'primary'
    backgroundColor: 'lightgray',
    margin: [ 1, 3 ]
  }
   // this would allow setting a component theme via a theme prop or via context
  const theme = useTheme(defaultTheme, props.theme)
  return (
    <div
      css={{
        fontWeight: theme.fontWeight,
        fontSize: theme.fontSize,
        color: theme.color,
        backgroundColor: theme.backgroundColor,
        margin: theme.margin
      }}
      {...props}
    />
  )
}
```

Where the breakpoint theme values are computed using a container query using the component root element instead of a media query? This would better support themes for reusable components that adjust based on container size. FWIW ResizeObserver has pretty decent support now.

jxnblk:
Sorry for the slow response! Using ResizeObserver for "responsive" values is an interesting idea! I think to keep inline with the current API, the responsive values should probably be defined at the component level rather than the theme level though. If you'd like to add an optional package to support something like this, feel free to open a PR

alexpermyakov:
It would be really great to have breakpoints based on the component's width. Then we will be able to do cool things like this: https://philipwalton.github.io/responsive-components/#calendar 

ResizeObserver is probably the best option, but we need JS - so, it won't be a part of styled-system css() function.
Should Box component handle this?

Another question, if we define a component this way:

`<Text width={['100%', '50%']} breakpoints={['40em', '52em']}>Hello</Text>`

Does prop-breakpoints override breakpoints for the theme?

junyper:
@alexpermyakov if container queries are an option, I would suggest not setting global/theme level breakpoints and instead always define them at the component level.

alexpermyakov:
@junyper Agree 👍
Actually, we can slightly extend css() function and provide a dynamically calculated breakpoint value (via ResizeObserver):

```
test('returns styles for specified breakpoint', () => {
  const result = css({
    width: ['100%', '50%', '25%'],
    breakpoints: ['40em', '52em'],
  })({ breakpoint: 0 })

  expect(result).toEqual({
    width: '50%',
  })
})
```

jxnblk:
[...]

PR #587, opened 2020-01-20
Responsive components

alexpermyakov:
This is an idea of how to solve #388.


packages/responsive-hooks/README.md
```diff
@@ -0,0 +1,39 @@
+# @theme-ui/responsive-hooks
```
jxnblk: Thanks for putting this together! Since the word "responsive" typically means media queries, I think this package should explicitly be named something relating to "container queries" or "resizeObserver", especially to distinguish itself from the `@theme-ui/match-media` package

packages/responsive-hooks/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Creation
+
+```js
+import { useBreakpointIndex } from '@theme-ui/responsive-hooks'
+
+const ResponsiveText = props => {
+  const ref = useRef(null)
+  const index = useResponsiveIndex(props.breakpoints, ref.current)
```
jxnblk: It'd be great if the hook here used `useThemeUI` internally to get the breakpoints from the theming context, but allow for bespoke overrides with an argument. I think the hook could also call `useRef` internally, so maybe the API could look something like: `const [ref, index] = useContainerQuery()` or `const [ref, index] = useContainerQuery([ '32em', '48em' ])`

alexpermyakov: I've implemented this API, thanks for the suggestion, it's much cleaner. 

There are two things we have to clarify:

1. How to convert 'em' and 'rem' to px (ResizeObserver returns a width in pixels), it wasn't a case with CSS media-queries, but now we implement it with JS. AFAIK there are no decent solutions except some w3c drafts on how to convert em to px. I've provided my own code for it (requires to access DOM once).

2.  It works with all properties except **width** (a problem with a circular dependency):

If we provide a width based on a breakpoint index, then ResizeObserver calls a handler (because the width has been changed), then it will return a new index based on a new width, which leads to another ResizeObserver call, and again.

I can focus on tests and after that work on the circular dependency problem. What do you think @jxnblk

alexpermyakov: I think _width_ cannot depend on the breakpoints. Example:
```
<ResponsiveText breakpoints={['480px']} widths={['100%', '25%']}>
  Hello
</ResponsiveText>
```
We use the next algorithm for getting a breakpoint index:
1. Take breakpoint index 0 as an initial value
2. Apply width as widths[0], it means '100%'
3. If the width more than 480px now, then apply a breakpoint width index 1
4. Apply width as widths[1], it means '25%'
5. If the width less than 480px now, then apply a breakpoint width index 0
6. Go to point 2. Cycle

All styles depend on the breakpoints prop and the breakpoints prop doesn't depend on them. Except for _width_ prop: breakpoints also depend on them.

packages/container-query/src/utils.js
```diff
@@ -0,0 +1,31 @@
...
+export const em2px = (el, v) =>
+  parseFloat(getComputedStyle(el.parentNode).fontSize || 16) * parseFloat(v)
+
+export const rem2px = v =>
+  parseFloat(getComputedStyle(document.documentElement).fontSize || 16) *
+  parseFloat(v)
+
+const supportedUnits = new Set(['em', 'rem', 'px'])
+
+export const unit2px = (el, val) => {
```
jxnblk: I'm not sure I understand the reason for all these utilities and why they would need access to the document/element

alexpermyakov: ResizeObserver returns the width of the element in 'px', but we can define breakpoints using 'em', in order to compare these two values we need to convert em to px. The 'em' size depends on the parent's font-size. Is there any other way to do so?

alexpermyakov: We can use 1em -> 16px, but it might be wrong and the only right way to do so is to use document/element properties.

alexpermyakov: Well, I think I know the reason for the confusion.

useContainerQuery accepts an array of breakpoints based on its own width, not on viewport's width( where we can use css media query).

I thought @junyper meant container query support. There is no CSS support for it and it won't be there any time soon. Let me know is this clear @jxnblk

packages/responsive-hooks/README.md
alexpermyakov: More explanation here:
https://twitter.com/philwalton/status/1220790253956395008

Maybe be can just implement custom breakpoints for the component based on the viewport width? API remains exactly the same.

packages/container-query/src/utils.js
alexpermyakov: @jxnblk any thoughts on this one? I would like to finish this PR.

jxnblk:
I think this is still worth pursuing -- I don't have an immediate "solution" for how `em` units should be handled in a utility like this so maybe the way this is handled in this PR is fine, but I wanted to keep it open for discussion. If you don't have time to work on this, no worries though

alexpermyakov:
Ok, I’ll add some examples. Let’s keep it here for a while.

alexpermyakov:
[CSS Typed Object Model](https://drafts.css-houdini.org/css-typed-om-1/#stylepropertymap) can help us with units converting, see [demo](https://developers.google.com/web/updates/2018/03/cssom). It's still need the context - a parent node dom-element, but we can avoid a _string-number-string_ convertion.
Now:
`const em2px = (el, v) => parseFloat(getComputedStyle(el.parentNode).fontSize) * v`

With upcomming API:
`const em2px = (el, v) => el.parentNode.computedStyleMap().get('font-size').value * v`

packages/container-query/src/utils.js
alexpermyakov: We can get rid of all these utilities function using CSS Custom Properties:
```
CSS.registerProperty({
  name: '--size',
  syntax: '<length>',
  inherits: false,
  initialValue: 0
});
```
el.style.setProperty('--size', '10em');
el.computedStyleMap().get('--size').value; // returns a correct value in pixels. 

I'll update this PR.

alexpermyakov:
We can get rid of all these utilities function using CSS Custom Properties:

CSS.registerProperty({
  name: '--size',
  syntax: '<length>',
  inherits: false,
  initialValue: 0
});

el.style.setProperty('--size', '10em');
el.computedStyleMap().get('--size').value; // returns a correct value in pixels. 
I'll update this PR.

jxnblk:
Sorry, I haven't had much time to look at the implementation in detail, but generally I think this should try to only use APIs that are available on all the major/modern browsers -- i.e. it's better if it can *just work* in most cases

alexpermyakov:
Why not to use this new API and leave unit2px methods for the rest?  It might be something like:
const unit2px = () = > {
  if (el.computedStyleMap) {
    el.style.setProperty('--size', '10em');
    el.computedStyleMap().get('--size').value;
  } else {
    // use current unit2px method.
  } 
}

Both Safari and Firefox are working on computedStyleMap API, Edge is the chromium based now, this `else` swich might be removed sooner rather than later.

Issue #583, opened 2020-01-16
[@theme-ui/components] Add variants for Spinner component 


alexpermiakov:
**Is your feature request related to a problem? Please describe.**
For now, theme-ui Spinner component is quite hardcoded - keyframe animation is baked into it.

Some years ago I created [Styled HTML5 Progress element demo](https://codepen.io/alexdevp/pen/JXRWrJ)

**Describe the solution you'd like**
It would be great to pass keyframe animation into the theme. For example:

```
const theme = {
  buttons: {
    primary: {
      color: 'background',
      bg: 'primary',
    },
    secondary: {
      color: 'background',
      bg: 'secondary',
    },
  },
  spinners: {
    primary: {
      title = 'Loading...',
      color: 'secondary',
      animation: {
        name: spin.toString(),
        timingFunction: 'linear',
        duration: 500,
        iterationCount: 'infinite',
      },
    },
  },
}
```

While it's fine for the spinners above, it's not enough for others: for example, [these nice ones](https://loading.io/css/) have a different number of animated items. One solution might be to pass an argument - the number of animated items (usually it's between 1-12).

**Describe alternatives you've considered**
Create named Spinners, something like HeartSpinner.

jxnblk:
Haven't looked at the related code very closely yet, but think it totally makes sense to support Spinner variants in ways that it might not already handle -- IIRC you should be able to target the inner `<circle>` element from a variant, but YMMV

lachlanjc:
Hey! I think themes are already diverse enough in their scope & functionality (writing CSS & theme-based CSS in variants/components, styling document content in `styles`, the scales, plus Theme UI configuration options), I think it’d make it pretty confusing to also define component props inside the theme. I think these spinners would be a great fit either for the Recipes on the docs site or as an external package linked to on our Resources page (#1275).

@hasparus I’m cool to close

PR #599, opened 2020-01-22
Add more Spinner components

alexpermyakov:
This is a test implementation for #583. There are so many great spinners, it would be great to be able to create new ones.

Demo:
spinners


alexpermyakov:
I've added itemX/itemY prop functions along with the number of items and the animation props which allows us to implement pretty much any spinner.

**Things to clarify:**
1. What is better `<BarSpinner />` or <Spinner variant='bar' />? For the first one we still can apply the theme for changing colors/opacity: <BarSpinner variant='primary' />
2. Props vs Theme: while some props like _color_  should be a part of the theme, this is not really clear about _width/height_.
3. Where should we define animations? If we have a separate file for each spinner, it can be inside.
4. Do we need props like itemHeight/itemWidth? Even though it gives us some flexibility, it looks a bit excess. Instead the size of each spinner item might be relative to the size of the spinner itself. 
Probably we need width, height, duration props. Not sure about others.

alexpermyakov:
Do we even need all these spinners @jxnblk? It's a bit more than just variants.

alexpermyakov:
If these spinners make sense, I can polish them, apply variants, add a few more and update the doc, @jxnblk FYI.

alexpermyakov:
Well, I'm not sure this implementation is fully aligned with theme-ui philosophy.
I can close this PR after getting some feedback.

lachlanjc:
Your work here looks absolutely awesome, but yeah, I don't know that it makes sense to bundle them all directly. Either an extension package inside theme-ui or of course you could publish your own. Thoughts @jxnblk?

alexpermyakov:
Maybe instead of actually creating all these spinners - we can update the doc and put these examples there.

alexpermyakov:
Maybe it's a bit early days, should I close it @jxnblk?

RishikeshDarandale:
Are you planning to add it back?

lachlanjc:
@alexpermiakov It’d be awesome if you’re still interested in adding these to the docs!

Issue #577, opened 2020-01-13
Order of generated media queries is wrong


454de6e:
**Describe the bug**
I came across a case when the order of media queries generated by Theme UI is wrong and incorrect styles are applied. The root cause of the bug might be in `styled-system`, not `theme-ui`.

**To Reproduce**
1. Clone https://github.com/UNDataForum/gatsby-themes and switch to branch `theme-ui-issue`
1. Run `yarn` and `yarn start`.
1. Make sure your viewport width in Chrome or Firefox is more than 960px (60em.)
1. Load http://localhost:8000/ and inspect the page with DevTools.
1. Choose the parent element of the "United Nations" text in the hero image. See this comment for a screenshot that shows the element you need to choose https://github.com/UNDataForum/gatsby-themes/issues/291#issuecomment-573878753
1. Look at the CSS rules that are applied to the chosen element (see second screenshot in issue comment from previous step).
1. You should see that the `min-width` media queries have the wrong order and the `min-width: 30em` rule overrides the `min-width: 60em` rule leading to the wrong left and right padding being applied.

**Expected behavior**
The `min-width: 60em` media query should be at the top and override the `min-width: 30em` query.

**Additional context**
The CSS is generated from the `sx` prop here: https://github.com/UNDataForum/gatsby-themes/blob/theme-ui-issue/site/src/components/hero.js#L50-L64:

```
sx={{
  flexDirection: 'column',
  justifyContent: ['flex-start', null, 'flex-end'],
  color: 'background',
  backgroundImage: [
    theme =>
     `linear-gradient(to bottom, ${theme.colors.primary}, transparent 75%  )`,
    null,
    theme =>
      `linear-gradient(to top, ${theme.colors.text}, transparent 75%)`,
    ],
  height: '100%',
  px: [2, 3, 4],
  py: 4,
}}
```
Note the 3 CSS pros that define responsive behavior with an array. 2 of them skip one breakpoint with a `null` value as described in the [styled-system docs](https://styled-system.com/guides/array-props/#origins).

I noticed that if I replace **all** `null` values in the responsive arrays with preceding array value (creating more verbose CSS) like this

```
sx={{
  flexDirection: 'column',
  justifyContent: ['flex-start', 'flex-start', 'flex-end'],
  color: 'background',
  backgroundImage: [
    theme =>
     `linear-gradient(to bottom, ${theme.colors.primary}, transparent 75%  )`,
    theme =>
[...]

PR #600, opened 2020-01-22
Adjust media query sort logic 

jxnblk:
Related to #577


anguy95:
Glad this problem is fixed. I am running into this problem as well right now. I hope it gets merged in soon. 🙏🏼

PR #605, opened 2020-01-23
Add additional docs for child selectors

johno:
It's not always immediately clear that you can use
Emotion selectors which also access the theme, so
this adds an explicit section showing some usage.

### DIFF
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -146,6 +146,32 @@ export default props => (
 )
 ```
 
+## Child Selectors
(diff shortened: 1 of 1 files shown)


lachlanjc:
This is great 👍 Critical that all users know this is possible.

packages/docs/src/pages/sx-prop.mdx
```diff
@@ -146,6 +146,32 @@ export default props => (
 )
 ```
 
+## Child Selectors
```
jxnblk: This should come with a warning IMO :)

PR #607, opened 2020-01-24
Add section to docs about skipping breakpoints

keirwilliams:
Took me a while to figure out I could do this. So, I thought it would be valuable information in the docs. Feel free to reword it.


jxnblk:
Thanks!

Issue #609, opened 2020-01-26
non-pixel units (not mentioned in docs)


yairEO:
Hi, good job with this, been using for almost a year.

No where in the docs units are mentions, and all examples uses "pixels" and the user reading is expecting to figure it out without explanation.

I want to define `spaces`, `fontSizes` and pretty much anything in other units. sometimes `em`, sometimes `em`, sometimes `%`..  it would be incredibly helpful to show it in the example pages, [this](https://system-ui.com/theme/) & [this](https://theme-ui.com/theming/)

I have **no idea** why there are two pages explaining the same thing.. it's confusing which is better.

---

Also i'm reporting a dead link in the README:

![image](https://user-images.githubusercontent.com/845031/73139935-f8f23780-407b-11ea-87f2-a13c1bbd6a14.png)

jxnblk:
Thanks! That's a good point, and we should definitely mention that more explicitly in the docs

flo-sch:
I think we might be able to close this?

On theme-ui.com, the [Theme Specifications](https://theme-ui.com/theme-spec#font-sizes-typographic-scale) and [Theming](https://theme-ui.com/theming#typography) pages mention both:

> Values defined as numbers are converted to pixel units where appropriate. For other units, use string values, e.g. '1em'.

And system-ui.com is potentially out of scope for this issue tracker...?

lachlanjc:
Yep, agreed on all counts. @hasparus Can you close?

PR #610, opened 2020-01-26
Fix Theme Specification link in README

lachlanjc:
As mentioned in #609

### DIFF
--- a/packages/theme-ui/README.md
+++ b/packages/theme-ui/README.md
@@ -54,7 +54,7 @@ help you build UI rooted in constraint-based design principles.
 [mdx]: https://mdxjs.com
 [styled system]: https://styled-system.com
 [gatsby]: https://gatsbyjs.org
-[theme specification]: https://system-ui.com/theme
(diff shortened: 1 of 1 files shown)


packages/theme-ui/README.md
```diff
@@ -54,7 +54,7 @@ help you build UI rooted in constraint-based design principles.
 [mdx]: https://mdxjs.com
 [styled system]: https://styled-system.com
 [gatsby]: https://gatsbyjs.org
-[theme specification]: https://system-ui.com/theme
+[theme specification]: https://system-ui.com/theme-spec
```
jxnblk: Thanks! I think *this* link was correct, but the one on line 81 looks like it's a relative link, which is probably causing the issue

lachlanjc: Lol I totally messed this up, sorry!

lachlanjc:
Running into some weird edge case with GitHub here…commit shows as empty but the file shows changes. Closing.

PR #623, opened 2020-01-28
[docs] Fix components variants example

mxstbr:
The displayed example component is a secondary button, yet the specified example theme doesn't have a secondary variant for buttons defined, which is incongruent.

This fixes it by renaming the variant defined in the example theme


jxnblk:
Thanks! I wonder if we should show both a `primary` and `secondary` variant in the theme to make it more clear as to how it's intended to work...

lachlanjc:
This seems worth merging even if we should improve clarity here in the future?

PR #625, opened 2020-01-28
Add position relative to Checkbox and Radio component

isBatak:
This should fix issues like the one in this example https://f79d6.sse.codesandbox.io/ where the input tag is rendered outside the scroll container and extending body content below the fold.
Also, it seems like it is only an issue in the Chrome browser.

### DIFF
--- a/packages/components/src/Checkbox.js
+++ b/packages/components/src/Checkbox.js
@@ -39,7 +39,7 @@ const CheckboxIcon = props => (
 
 export const Checkbox = React.forwardRef(
   ({ className, sx, variant = 'checkbox', ...props }, ref) => (
-    <Box>
+    <Box sx={{ position: 'relative' }}>
       <Box
         ref={ref}
         as="input"
--- a/packages/components/src/Radio.js
+++ b/packages/components/src/Radio.js
@@ -39,7 +39,7 @@ const RadioIcon = props => (
 
 export const Radio = React.forwardRef(
   ({ className, sx, variant = 'radio', ...props }, ref) => (
-    <Box>
+    <Box sx={{ position: 'relative' }}>
       <Box
         ref={ref}
         as="input"


packages/components/src/Checkbox.js
```diff
@@ -39,7 +39,7 @@ const CheckboxIcon = props => (
 
 export const Checkbox = React.forwardRef(
   ({ className, sx, variant = 'checkbox', ...props }, ref) => (
-    <Box>
+    <Box sx={{ position: 'relative' }}>
```
jxnblk, line 42: Thanks! But generally we shouldn't introduce any positioning to the default styles if possible since it will create a new positioning context and can cause unintended style bugs -- can you link to a reproduction of the issue you're seeing along with a screenshot for context here?

isBatak: @jxnblk sorry for the delay. 
Here is the example https://f79d6.sse.codesandbox.io/
This is kinda related to form behavior. If you focus on an input element form will be automatically scrolled to that position.
I'll demonstrate it in a video:
https://monosnap.com/file/cCJMMlZvXtBXzIR61DO0nDCirTG0a8

isBatak: Sorry, I added `position: relative` to `label` element in the video, but it's the same fix if you add it on checkbox input wrapper `div` element.

jxnblk: Based on the video, that *looks* like normal browser behavior -- not sure if I'm misunderstanding this/don't see a link to the code, but maybe you can clarify the issue?

isBatak: @jxnblk the problem is in an extra gap/area at the bottom of the screen which makes the whole body jump (and scrollable), that should not happen, instead only light blue sidebar with red border should be scrollable.
The fix is to add `position: relative` to checkbox wrapping div.

lachlanjc:
Thanks a bunch for this PR—however, we’re not intending to merge it since this should be handled by users on a per-case basis. We also have an open issue for allowing props on the container components for doing this more easily.

devatrox:
@lachlanjc Can you link to the issue with the container props? I would really love this feature implemented

Issue #416, opened 2019-10-21
Support top, right, left, and bottom border variants


TomerAberbach:
Hi! First just wanted to thank you for this awesome package!

Would you be able to add support for the `border-top`, `border-right`, `border-left`, and `border-bottom` variants of `border-style`, `border-width`, `border-color`, and `border-radius`?

Currently only the [`borders` theme key supports `border-top`, `border-right`, `border-left`, and `border-bottom`](https://theme-ui.com/theme-spec/#theme-scales). After this change:
* The `borderWidths` theme key would also apply to `border-top-width`, `border-right-width`, `border-bottom-width`, and `border-left-width`.
* The `borderStyles` theme key would also apply `border-top-style`, `border-right-style`, `border-bottom-style`, and `border-left-style`.
* The `colors` theme key would also apply to `border-top-color`, `border-right-color`, `border-bottom-color`, and `border-left-color`.
* The `radii` theme key would also apply to `border-top-right-radius`, `border-bottom-right-radius`, `border-bottom-left-radius`, and `border-top-left-radius`.

I'd be happy to implement this change myself if you could give me some code pointers.

dburles:
Hey @TomerAberbach are you on the latest version of theme-ui? they were added them a little while back so they should all function! https://github.com/styled-system/styled-system/pull/671/files

TomerAberbach:
I think I must have mistyped `borderBottomWidth` as `borderWidthBottom` and that's why it didn't work for me! I had myself convinced that the properties I mentioned must not be supported by the package because [the page I linked](https://theme-ui.com/theme-spec/#theme-scales) does not list them in the table. But I'm glad to learn that they are supported!

Maybe I'll write make pull requests for the [Theme UI Theme Specification](https://github.com/system-ui/theme-ui/blob/master/packages/docs/src/pages/theme-spec.mdx) and [Styled System Theme Specification](https://github.com/styled-system/styled-system/blob/master/docs/theme-specification.md) pages so that the tables remain in sync with the supported properties!

dburles:
Yep they definitely should be in the table. It would be worth going over them to see if anything else is missing too. https://github.com/styled-system/styled-system/blob/master/packages/css/src/index.js I see `flex-basis`  in `sizes` is one.

TomerAberbach:
[...]

PR #626, opened 2020-01-29
Update theme scales in documentation

dburles:
As discussed in #416, this PR adds a `ThemeScales` component which generates a table directly from the scales defined in the css package.


jxnblk:
This is great! Thanks for putting this together!

Issue #620, opened 2020-01-28
sidenav css outline issue


t2ca:
There appears to be a CSS bug in Safari that shows an outline on the top link "Getting Started"

Steps to reproduce:
1. Open Safari on desktop
2. Make window smaller until menu shows up
3. Click on hamburger menu
4. The top link show an outline

lachlanjc:
Not sure what to do here…isn't that expected behavior? The top link is outlined, when you Opt-Tab, the outline goes to the next link. What do you think it should do instead?

t2ca:
In my experience the outline, which is a purple border, stays on the top link "Getting Started" even after clicking on different menu links. Im not getting the same result in chrome.

jxnblk:
Closing this because this is the intended behavior: opening the menu focuses the first element in the navigation

PR #627, opened 2020-01-29
Fixes a minor issue with Safari in docs

dburles:
See #620

### DIFF
--- a/packages/docs/src/gatsby-plugin-theme-ui/index.js
+++ b/packages/docs/src/gatsby-plugin-theme-ui/index.js
@@ -132,6 +132,7 @@ export default {
       transitionTimingFunction: 'ease-out',
       transitionDuration: '.2s',
       borderRadius: 2,
+      outline: 'none',
       '&:hover': {
         bg: 'highlight',
       },


packages/docs/src/gatsby-plugin-theme-ui/index.js
```diff
@@ -132,6 +132,7 @@ export default {
       transitionTimingFunction: 'ease-out',
       transitionDuration: '.2s',
       borderRadius: 2,
+      outline: 'none',
```
jxnblk: Haven't looked at this in context yet, but how would this work for keyboard users? This looks like it breaks accessibility at first glance

dburles: Yeah you're right, I've realised that aside from this bug, accessibility seems to be broken in Safari. I'll open a new issue

dburles: Right so it *is* working, though it needs to manually be enabled in preferences. That just leaves the original glitch to be fixed by some other means :)

PR #637, opened 2020-02-01
Clarify Gatsby link imports in docs

### LINKED ISSUE #597: Clarify <Link/> import in docs
The Theme UI Docs specify two packages from which to import a `<Link/>`.

1.
https://theme-ui.com/components/link
`import { Link } from '@theme-ui/components'`

2.
https://theme-ui.com/recipes/gatsby-link
```
/** @jsx jsx */
import { jsx } from 'theme-ui'
import { Link } from 'gatsby'
```

The first example in the Gatsby recipe does not specify which import is being used. Propose that docs should clarify.

lachlanjc:
Not super ideal that (I don’t think) we can include the imports directly in the live example, but I think this is a bit clearer about what to import & then giving a playground for changing the design.

Closes #597

### DIFF
--- a/packages/docs/src/recipes/gatsby-link.mdx
+++ b/packages/docs/src/recipes/gatsby-link.mdx
@@ -4,8 +4,25 @@ name: 'Gatsby Link'
 
 # Gatsby Link
 
-Add an active style to Gatsby Link or other React router link components.
-Import the [custom JSX pragma](/guides/jsx-pragma) for styling the router link components without wrappers.
+Add an active style to [Gatsby Link] or other React router link components.
+Import the Theme UI [custom JSX pragma](/guides/jsx-pragma) for styling
+the router link components directly, without using wrapper components.
+
+```jsx live
+<Link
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! I *think* (but might be wrong) that the recipes theme won't create the thumbnail correctly here if the live example isn't the first code fence: https://deploy-preview-637--theme-ui.netlify.com/recipes -- I wonder if there's another way to clarify that this is showing the Gatsby Link component. Maybe an import statement that's commented out??

lachlanjc:
Ah, you’re right. This is kind of a separate issue, but comments in live blocks always go invisible for me, on all themes:

image

jxnblk:
Ah yup, forgot about that...

lachlanjc:
Ok, this isn't the best/clearest thing ever, but I think it's nonetheless an improvement

jxnblk:
Agreed. This is definitely clearer than before -- thanks for looking into this!

PR #646, opened 2020-02-04
Accept object fontSize values for style guide

kripod:
As outlined [in the docs](https://theme-ui.com/theme-spec#font-sizes-typographic-scale), the `fontSizes` scale may be defined as an object instead of an array. To avoid an error being thrown, `Object.values(fontSizes)` should be used instead of `fontSizes` when mapping through values.


kripod:
While this fix helps a bit, it still doesn't solve the issue of too small texts rendered on the style guide page:

image

It turns out that the usage of `<TypeStyle fontSize={5 /* 5 or larger */}>` makes Theme UI choose `5px` when `fontSizes[5]` is undefined. I would suggest using arbitrarily given values for demonstrations in the style guide.

jxnblk:
Thanks! This LGTM

PR #662, opened 2020-02-10
Spike out potential TypeScript conversion

mxstbr:
This "reworks" `@theme-ui/core` to be written in TypeScript and export its own type definitions.  Thankfully, `microbundle` has built-in support for TypeScript so we barely need to do any build system changes. I also checked in the generated `index.d.ts` file (normally we would not check this into git) to show what it looks like.

cc @hasparus the DefinitelyTyped maintainer for the theme-ui types. No action items right now, just wanted you to be aware of this effort :wink:


mxstbr:
Here's where I am at: the `core` package is now written entirely in TypeScript, and generates a typedef file in the build process that (roughly) looks like the one that is in DefinitelyTyped right now, mostly because I copied all the custom types from there 😅

https://github.com/system-ui/theme-ui/blob/92e8090d397de19590a715c3e9b4f7bbaae58295/packages/core/index.d.ts#L1-L23

https://github.com/system-ui/theme-ui/blob/92e8090d397de19590a715c3e9b4f7bbaae58295/packages/core/theme.d.ts#L1-L134

*These files are currently committed to the root of the package so y'all can see what they look like. I will delete them again before we ship this.*

@hasparus does this look okay-ish to you, both the types and the general build process of how this works (automatically)? Any thoughts/feedback?

Once we are happy with where the types & the build process are at, then we ship this conversion and start tackling the other packages from there. I will be opening a planning issue with a clear outline what needs to be converted so all the contributors can help with this :pray:

packages/core/src/theme.d.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+export interface Theme extends Omit<StyledSystemTheme, 'colors'> {
+  /**
+   * Enable/disable custom CSS properties/variables if lower browser
+   * support is required (for eg. IE 11).
+   *
+   * References: https://theme-ui.com/color-modes/#turn-off-custom-properties
+   */
+  useCustomProperties?: boolean
```
jxnblk: Looks like there's a few newer configuration options not in here yet: https://theme-ui.com/theming#configuration-flags

packages/core/src/theme.d.ts
```diff
@@ -0,0 +1,143 @@
+import * as CSS from 'csstype'
+import { SystemStyleObject } from '@styled-system/css'
+import { Theme as StyledSystemTheme } from 'styled-system'
```
jxnblk: I'm guessing the plan is to eventually have the entire Theme interface defined in this repo, right?

packages/core/src/index.ts
```diff
@@ -28,10 +31,13 @@ const parseProps = props => {
   return next
 }
 
-export const jsx = (type, props, ...children) =>
+export const jsx: typeof React.createElement = (type, props, ...children) =>
```
hasparus: Although it isn't 100% correct (jsx handles `sx` and `css` props, createElement doesn't), focusing on this now would be bikeshedding.

packages/core/src/index.ts
```diff
@@ -1,11 +1,14 @@
 import {
   jsx as emotion,
   ThemeContext as EmotionContext,
+  InterpolationWithTheme,
 } from '@emotion/core'
-import { css, get } from '@theme-ui/css'
+import { css } from '@theme-ui/css'
 import React from 'react'
 import deepmerge from 'deepmerge'
+// @ts-ignore
```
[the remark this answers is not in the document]
hasparus: It is a detail tho. I added it in https://github.com/system-ui/theme-ui/pull/665.

packages/core/src/index.ts
mxstbr: Feel free to start submitting PRs to clean this up :wink:

Issue #651, opened 2020-02-05
When using `useThemeUI`, the `theme` does not return default theme values


jordanoverbye:
I was speaking with @johno  about this as it's an issue with Blocks UI.

Lets say you have a super basic theme like this and you use that with `ThemeProvider`.

```js
const theme = {
  colors: {
    primary: 'red',
    secondary: 'blue',
  },
}

export const App = props => <ThemeProvider theme={theme}>{props.children}</Theme>
```

When you use the `useThemeUI` hook, it only returns the original theme. 

```js
export const ExampleComponent = props => {
  const { theme } = useThemeUI()
  // `theme` is the same object as above
}
```
This is kind of confusing, because it does not show the default values that get applied when using the `sx` prop like `spacing`, `fontSizes`, `fontWeights` etc.

## Expected behaviour:
The `theme` object returned from `useThemeUI` has all the default theme values merged.

jxnblk:
This makes a ton of sense. I think only the `space` and `fontSizes` scales use default values and those are defined in `@theme-ui/css` if anyone wants to look into this -- potentially could be refactored to behave more like default values at the theme level

hasparus:
Is this up for grabs? I just encountered this problem and it seems pretty straightforward so I'd rather contribute than workaround on my side.

jxnblk:
@hasparus feel free to take a look into this -- if it ends up being a cross-package refactor, we'll probably want to make sure there we don't introduce any unintentional breaking changes FWIW

PR #663, opened 2020-02-11
Return default theme values from useThemeUI

hasparus:
Closes https://github.com/system-ui/theme-ui/issues/651

Key part of the PR:
https://github.com/system-ui/theme-ui/pull/663/files#diff-d1c352d7f9f3a715b84c656a6704289aR82-R88

- I think values of the default theme are not mentioned in the docs, so I just test if the default theme is about the same shape. A change that passes the tests shouldn't crash user apps. It might change styles though.
I thought a bit about checking for actual equality with arrays in default theme but this would be probably too much.
- I'm not sure why lint-staged made such a big diff, because neither husky config nor prettier config didn't change for months (I used save without formatting to touch as little as possible :D)
- **Problems**
  - _@theme-ui/css_ feels like a little odd place to keep a default for `useThemeUI`, but I have no idea how do do this without
    - either a breaking change: `css` function used standalone (like in `packages/css/test`) gets no defaults
    - or introducing weird circular dependency between packages, what would kill the point of having them separate
  - I just introduced exactly the same private API export I talked with @isBatak about earlier today.
  I should probably add `@internal` JSDoc comment for clarity.

### DIFF
--- a/packages/color/src/index.ts
+++ b/packages/color/src/index.ts
@@ -1,8 +1,11 @@
 import * as P from 'polished'
 import { get, Theme } from '@theme-ui/css'
 
-const g = (t: Theme, c: string) =>
-  get(t, `colors.${c}`, c)
+/**
+ * Get color from theme.colors
+ */
+export const getColor = (theme: Theme, color: string) =>
+  get(theme, `colors.${color}`, color)
     .replace(/^var\(--(\w+)(.*?), /, '')
     .replace(/\)/, '')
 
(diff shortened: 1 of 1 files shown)


hasparus:
I broke this PR accidentally. This should be commit https://github.com/system-ui/theme-ui/commit/af45e8412a3f894e8a1ac90bad2564087d402b53 instead of e932b4f.

jxnblk:
Thanks for looking into this! I don't know if I have any great direction advice at the moment, but this makes me wonder if the "default theme" (font sizes and space) should be removed in the next breaking change and supplemented with a preset that could handle this. Even if we keep the defaults, maybe having a separate preset package could help with the cross-dependency piece of this

PR #665, opened 2020-02-13
 Add resolveJsonModule to @theme-ui/core tsconfig 

hasparus:
I wondered if my comment https://github.com/system-ui/theme-ui/pull/662#discussion_r379112712 sounds rude. I figured a small PR would make it nicer.


mxstbr:
The comment did not sound rude at all! However, I love the PR, keep 'em coming!! :heart_eyes:

Can you resolve the build failure? Then I'll merge this :ok_hand:

```
Error: /home/circleci/repo/packages/core/src/index.ts(104,30): semantic error TS2769 No overload matches this call.
  The last overload gave the following error.
    Type '{}' is missing the following properties from type 'ContextValue': __EMOTION_VERSION__, theme
```

hasparus:
rollup-plugin-typescript-2 used in microbundle [enforces few compiler options](https://www.npmjs.com/package/rollup-plugin-typescript2#some-compiler-options-are-forced). This is the reason the error happened in build, but not in text editors :c

I have added `moduleResolution: "node"` to `compilerOptions`, because it's enforced in microbundle, and exactly what we'd expect anyway. Since it made deepmerge  to typecheck `deepmerge` I took the liberty to annotate `merge` exactly like deepmerge without last argument.

Screenshot from 2020-02-14 09-40-56
image

Then, `context` type inferred properly and gave me a nicer error.

```
Property '__EMOTION_VERSION__' is missing in type '{ theme: Partial<Theme>; }' but required in type 'ContextValue'.
```

`__EMOTION_VERSION__` was missing from context value. But hey! It shouldn't be, as it is present in `outer`. I have added a generic parameter here
```ts
merge.all<ContextValue>({}, outer, { theme })
```
to let TypeScript know that we are indeed merging not only objects of shape `{ theme: Partial<Theme> }`, but entire `ContextValue`s with the `__EMOTION_VERSION__` included.

Issue #668, opened 2020-02-14
Convert to TypeScript


mxstbr:
Based on the discussions around #121 and the test we just merged in #662, we are going to convert theme-ui to TypeScript! :tada: The plan is to go package-by-package and gradually move all `.js` files to `.ts` files. 

As you can see there is a fair amount of packages to convert, so **we would love your help**! :muscle:

Here is how to convert a package (take a look at #662 for an example):

1. Add the [`--tsconfig tsconfig.json` option to the "prepare" and "watch" commands](https://github.com/system-ui/theme-ui/blob/3525f07098a466baf8708005ba706343b4a0fe4d/packages/core/package.json#L10-L11) in the `package.json` and [copy the `tsconfig.json` from the core package](https://github.com/system-ui/theme-ui/blob/master/packages/core/tsconfig.json).
1. Add `"types": "dist/index.d.ts"` and `"source": "src/index.ts"` to the `package.json`
1. Go file-by-file, rename them from `.js` to `.ts` and fix all type errors and `any` types in the generated typedef (`dist/index.d.ts`). 
1. Submit a PR and tag me as a reviewer!

To avoid duplicate work please comment which package you want to work on (as long as nobody else is also working on it) so we can mark it as taken.

**Packages that still need to be converted:**

- [x] chrome (#909)
- [x] color (#672)
- [x] color-modes (claimed)
- [ ] components (https://github.com/system-ui/theme-ui/pull/692)
- [x] core (#662)
- [x] css (#672)
- [x] custom-properties (#671) 
- [x] editor (#900)
- [x] match-media (#696)
- [x] mdx (#703) 
- [x] preset-base (#675)
- [x] preset-bootstrap (#694)
- [x] preset-bulma (#694)
- [x] preset-dark (#694)
- [x] preset-deep (#694)
- [x] preset-funk (#694)
- [x] preset-future (#694)
- [x] preset-polaris (#694)
- [x] preset-roboto (#694)
- [x] preset-swiss (#694)
- [x] preset-system (#694)
- [x] preset-tailwind (#694)
- [x] preset-tosh (#694)
- [x] presets (#694)
- [x] prism (#805)
- [x] sidenav (#917)
- [x] style-guide
- [x] tachyons (https://github.com/system-ui/theme-ui/pull/691)
- [x] tailwind (https://github.com/system-ui/theme-ui/pull/686)
- [x] theme-provider (#708)
- [x] theme-ui (#724) 
- [x] typography (#890)

[...]

PR #671, opened 2020-02-15
fix(@theme-ui/custom-properties): TS conversion

LekoArts:
Converts custom-properties over to TS.

`.idea` was added to gitignore
`strict` was used for tsconfig.json

Part of https://github.com/system-ui/theme-ui/issues/668

PR #672, opened 2020-02-15
Adopt TS in theme-ui/color

hasparus:
Part of #668

I noticed @LekoArts is doing conventional commits and I'm stealing the idea.

### Adopted TypeScript

I commited this in a weird way to preserve file history (two first commits on this branch).

### Added jsdoc comments

I've added jsdocs with descriptions from README,
because they're visible in tooltips when you hover a function.
(accessible by ts.getLeadingCommentRanges in Compiler API)

### Added grayscale to README

This is a separate commit, so we can revert it easily.
I noticed it's missing and it doesn't look like it's internal API,
so I took to add it.

### TSDoc comments parameter descriptions

```
/**
 * Darken a color by an amount 0–1
 * @param c key from theme.colors
 * @param n number between 0 and 1
 */
export const darken = (c: string, n: number) => (t: Theme) =>
  P.darken(n, g(t, c))
```

### Autocomplete color names dependent on theme type?s correctly.

My colleague has mentioned the fact that we get editor support for
color names in `sx` prop and functions from `@theme-ui/color`
during development as a biggest DX limitation of theme-ui.

We can write a typo (like "primry") and it still typechecks
```
import { ColorMode } from '@theme-ui/core'

// "background" | "text" | "primary" | "secondary" | "muted" | "accent" 
type ColorName = keyof ColorMode;

export const darken = (c: ColorName, n: number) => (t: Theme) =>
  P.darken(n, t.colors[c])
```

It could be implemented by removing [string index signature](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks)
from ColorMode and turning ColorMode into an interface so more
properties can be added with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces)
and [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).

After such action, instead of `string` color parameters in theme-ui/color would
be typed as `ColorName` (or something similar).

I am not sure if it is legal to pass color representation (e.g. "#fff")
to the functions here. I'm looking at `g` function and I think it isn't, it is
also undocumented, so I'll assume that it's not.


hasparus:
I’m gonna fix build promptly, wanted to push sth fast to claim the task.

packages/color/package.json
```diff
@@ -11,6 +13,9 @@
     "@theme-ui/css": "^0.3.1",
     "polished": "^3.4.1"
   },
+  "peerDependencies": {
```
LekoArts: I think we should come to a consensus here whether to put the dependencies for types in peerDep, devDep or Dep. I opted for devDep: https://github.com/system-ui/theme-ui/pull/671/files#diff-03713682702149e044f8824251410926R22

hasparus: yeah, previously the packages depended on _common concepts_.
These concepts are now described as types and we keep them in `@theme-ui/core`, so the dependency needs to be noted in package.json.

I think there are two possibilities:

- `@theme-ui/color` has `@theme-ui/core` both in peer dependencies and a dev dependencies.
  - Dev dep only would be a TypeScript build error for users without any previous warning. This is also how I got _"semantic error TS2307 Cannot find module '@theme-ui/core'."_ on Circle CI :D
- `@theme-ui/color` depends on `@theme-ui/core`. It should be okay in monorepo with the same versions. I think I'll go this way since `color` already depends on `css`.

After the build there is no import to `@theme-ui/core` so it won't get bundled for people who use only color utils.

hasparus:
I'm working on tests in TS right now. It's probably good to have them converted, because types were tested in DefinitelyTyped.

hasparus:
I've added ts-jest, mostly because I didn't notice it's not here before changing `color/test/index.js` to TypeScript 🤦 Feel free to revert it if you'd rather test theme-ui with the same tools it's built with.

`ts-jest` compiles tests with `tsc` rather than Babel, so they're typechecked. No need for an additional "typecheck" step in CI. I think this is a good idea for a library, because types in DT were not only typechecked but additionally tested with [`dtslint`](https://github.com/microsoft/dtslint)'s _ExpectType_ and _ExpectError_.
A small thing to help to ensure that we're not trading off UX for ease of maintenance.

jxnblk:
Thanks for diving into this! Just a couple quick thoughts:

> We can write a typo (like "primry") and it still typechecks

Yes, we want to have better checks for typos from keys you've defined in your theme interface. Not sure what that ultimately will look like, but I'm guessing it's okay to punt on this for now with `@theme-ui/color`

> I am not sure if it is legal to pass color representation (e.g. "#fff")

All `sx` style objects and these functions should be able to accept either a key from the `theme.colors` scale *or* a raw color value. The leniency here is a core part of the API, but we have discussed having an opt-in "strict mode" to allow tighter control over how that works in the future

hasparus:
Sure, I just wanted to drop some notes.

> All sx style objects and these functions should be able to accept either a key from the theme.colors scale or a raw color value. The leniency here is a core part of the API, but we have discussed having an opt-in "strict mode" to allow tighter control over how that works in the future

So `string` here is the only correct. Thanks.

The dependency on core here could be dropped when @theme-ui/css exports `Theme`.

mxstbr:
@hasparus let's use ts-node for all the tests! :100:

hasparus:
Moving Theme turned out to be a pretty big change. This PR consumed https://github.com/system-ui/theme-ui/pull/673.

packages/color/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "strict": true,
+
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "moduleResolution": "node"
+  },
+  "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.d.ts"]
```
LekoArts: Did you run into issues without that explicit `include` ?

packages/custom-properties/package.json
```diff
@@ -19,7 +19,7 @@
     "pluralize": "^8.0.0"
   },
   "devDependencies": {
-    "@theme-ui/core": "^0.3.1",
-    "@types/pluralize": "^0.0.29"
+    "@theme-ui/css": "^0.3.1",
+    "@types/pluralize": "0.0.29"
```
LekoArts: You accidentally pinned the dependency here

packages/color/tsconfig.json
hasparus: Yup, tests get built.
```
dist
  src
  test
    index.test.d.ts
```

packages/custom-properties/package.json
hasparus: damn, thanks! pinning `types/pluralize` makes totally no sense here.

packages/color/tsconfig.json
LekoArts: Makes sense! Might want to add that to other files in a follow-up PR then

hasparus: Since this is a monorepo, we could add `tsconfig.json` to the root with all _base_ compiler options (like `moduleResolution node` and no `resolveJsonModule`) and then use [`"extends"`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) to inherit them. The tsconfig in root would be useful when you open the entire repo in an IDE.

I'm not sure for 100%, but I think "include" is inherited literally (relative to the new config, not to the parent), so most tsconfigs in packages could look like
```
{ 
  "extends": "../core/tsconfig.json",
  "compilerOptions": {
    "strict": true // if possible since core is unstrict
  }
}
```

mxstbr: Yes we should totally do this! 👍

PR #673, opened 2020-02-16
(@theme-ui/css) adopt TypeScript

hasparus:
Part of https://github.com/system-ui/theme-ui/issues/668

This is by no means a finished PR. I just wanted to push work on it a bit forward.
This may be the one of the trickiest packages in the repository to type.
`get` is the hardest part. Its namesake's types in @types/lodash don't look pretty.

I arrived at this `index.d.ts`
```ts
import { Theme } from '@theme-ui/core';
import * as CSS from 'csstype';
import { SystemStyleObject } from './types';
export * from './types';
export declare const get: (obj: object, key: string | number, def?: unknown, p?: number, undef?: unknown) => unknown;
export declare const multiples: {
    marginX: string[];
    marginY: string[];
    paddingX: string[];
    paddingY: string[];
    size: string[];
};
export declare const scales: {
    readonly color: "colors";
    // redacted
    readonly fill: "colors";
    readonly stroke: "colors";
};
interface CssPropsArg {
    theme?: Theme;
    [key: string]: unknown;
}
export declare const css: (args: SystemStyleObject) => (props?: CssPropsArg) => CSS.Properties<string | 0>;
```

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -12,5 +12,14 @@
   "license": "MIT",
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "csstype": "^2.5.7"
+  },
+  "devDependencies": {
+    "@theme-ui/core": "0.3.1"
+  },
+  "peerDependencies": {
+    "@theme-ui/core": "0.3.1"
   }
 }
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -1,7 +1,13 @@
-export const get = (obj, key, def, p, undef) => {
-  key = key && key.split ? key.split('.') : [key]
-  for (p = 0; p < key.length; p++) {
-    obj = obj ? obj[key[p]] : undef
+export const get = (
+  obj: object,
+  key: string | number,
+  def?: unknown,
+  p: unknown,
+  undef: unknown
+) => {
+  const path = key && typeof key === 'string' ? key.split('.') : [key]
+  for (p = 0; p < path.length; p++) {
+    obj = obj ? obj[path[p]] : undef
   }
   return obj === undef ? def : obj
 }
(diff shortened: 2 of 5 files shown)


hasparus:
I've encountered a problem with circular dependency between `css` and `core`. The Theme could be moved to `css` but then the "core" starts to be a bit confusing.

jxnblk:
WRT dependencies, `@theme-ui/css` is probably the lowest level package and I don't think it should depend on `core` or the other packages. You could argue that `css` should be included in the `core` package, but there are some bespoke use-cases for the `css` package as a standalone one.

For the `get` function, would that not be the sort of thing that should be typed with `any`?

hasparus:
> WRT dependencies, @theme-ui/css is probably the lowest level package and I don't think it should depend on core or the other packages. You could argue that css should be included in the core package, but there are some bespoke use-cases for the css package as a standalone one.

This clarified a lot for me. Thank you.

Similarly to @emotion/core, @theme-ui/core adds context and jsx function to the styling logic which is defined in css package. I understand that one could use @theme-ui/css in projects without React, maybe even augment other framework's jsx pragma with it.

I'm gonna move the `Theme` here.

> For the get function, would that not be the sort of thing that should be typed with any?

`any` sort of turns off type checking and TypeScript gives up all around it, but it probably makes sense in this case. If `get` was a main feature of the package we could add the several overloads like [in lodash](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/ts3.1/common/object.d.ts#L1023-L1032) (which has 18), but it may be not worth it here.

mxstbr:
Let's move Theme to css and depend on it from core (and all other packages that use the Theme type)!

mxstbr:
Closing as this was done in #672! Thanks @hasparus!

PR #684, opened 2020-02-17
@theme-ui/color - remove number arg as it's not used

allanpope:
Removed number argument as it's not being used in `grayscale` function.
As I understand it, the value has to be 1 for the color to be greyscaled.


jxnblk:
Thanks!

PR #695, opened 2020-02-20
fix readme for preset-tailwind

axe312ger:
As styles is part of `theme` already, I guess the README is wrong. This should correct the docs :)

See: https://github.com/system-ui/theme-ui/blob/master/packages/preset-tailwind/src/index.ts#L552

### DIFF
--- a/packages/preset-tailwind/README.md
+++ b/packages/preset-tailwind/README.md
@@ -13,9 +13,10 @@ import theme from '@theme-ui/preset-tailwind'
 
 export default {
   ...theme,
-  styles: {
-    ...theme,
-  },
+  // You can extend the default theme like:
+  // styles: {
+  //   ...theme.presets,
+  //}
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/preset-tailwind/README.md
```diff
@@ -13,9 +13,10 @@ import theme from '@theme-ui/preset-tailwind'
 
 export default {
   ...theme,
-  styles: {
-    ...theme,
-  },
+  // You can extend the default theme like:
+  // styles: {
+  //   ...theme.presets,
```
LekoArts:
```suggestion
  //   ...theme.styles,
```

You meant that? :)

axe312ger: yes 😅

axe312ger:
I see this pattern in the readme in all presets. Is it wrong everywhere or do I miss something?

https://github.com/system-ui/theme-ui/tree/master/packages/preset-future
https://github.com/system-ui/theme-ui/tree/master/packages/preset-polaris
https://github.com/system-ui/theme-ui/tree/master/packages/preset-roboto
...

jxnblk:
@axe312ger thanks! Yeah, that looks like a typo that was copypasta'd around

PR #696, opened 2020-02-20
@theme-ui/match-media : TypeScript conversion

LekoArts:
Changing `match-media` package to TS. Didn't touch the tests yet.

Part of #668

### DIFF
--- a/packages/css/src/types.ts
+++ b/packages/css/src/types.ts
@@ -511,7 +511,7 @@ export type ColorMode = {
 }
 
 export interface Theme {
-  breakpoints?: ObjectOrArray<number | string | symbol>
+  breakpoints?: Array<string>
   mediaQueries?: { [size: string]: string }
   space?: ObjectOrArray<CSS.MarginProperty<number | string>>
   fontSizes?: ObjectOrArray<CSS.FontSizeProperty<number>>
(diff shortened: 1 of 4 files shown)


packages/css/src/types.ts
```diff
@@ -511,7 +511,7 @@ export type ColorMode = {
 }
 
 export interface Theme {
-  breakpoints?: ObjectOrArray<number | string | symbol>
+  breakpoints?: Array<string>
```
LekoArts: According to the documentation this is only ever a `string[]`. Also was necessary to fix type issues in match-media

packages/match-media/package.json
```diff
@@ -2,21 +2,24 @@
...
+    "prepare": "microbundle --no-compress --tsconfig tsconfig.json",
+    "watch": "microbundle watch --no-compress --tsconfig tsconfig.json"
   },
   "publishConfig": {
     "access": "public"
   },
   "devDependencies": {
     "react": "^16.9.0",
-    "theme-ui": "^0.3.1"
+    "@theme-ui/core": "^0.3.1",
```
LekoArts: Switch out `theme-ui` for `/core` (to access `Theme`)

packages/css/src/types.ts
jxnblk: I'd guess the types for this might have originated in older versions of Styled System FWIW -- might be some other holdovers from that theme interface to keep an eye out for

PR #701, opened 2020-02-21
@theme-ui/match-media : Type variable

bldng:
This is an addition to https://github.com/system-ui/theme-ui/pull/696

**Motivation**
Initially I wanted to work on adding typescript support for `useResponsiveValue` but saw today that I was too late 😄. This change allows to either pass or in most cases infer the type of the array. The uses cases range from a straight forward condition
```ts
 const isNarrow = useResponsiveValue([true, false]); // -> boolean
```
to more complex configurations
```ts
const layoutOptions = useResponsiveValue<LayoutOptions>( // inferred in most cases
[
  {columns: 1, ...},
  {columns: 2, ...},
]); // -> LayoutOptions 
```
all within one call.

### DIFF
--- a/packages/match-media/src/index.ts
+++ b/packages/match-media/src/index.ts
@@ -52,9 +52,12 @@ export const useBreakpointIndex = (options: defaultOptions = {}) => {
   return value
 }
 
-type Values = ((theme: Theme | null) => string[]) | string[]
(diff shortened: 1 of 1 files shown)


LekoArts:
Since the README only showed a `string` and I never used that package, I went with that. But a generic is better, yeah 👍

PR #704, opened 2020-02-22
Compose styles with arrays in sx prop

jonavila:
This PR attempts to revive the work done @phelma (https://github.com/phelma) in #174. I took his work and put in the core package and update the `SxStyleProp` type. 

Sorry about the noise added by Prettier in the test file.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -13,10 +13,15 @@ import './react-jsx'
 
 export * from './types'
 
+const flattenDeep = arr =>
+  Array.isArray(arr)
+    ? arr.reduce((a, b) => a.concat(flattenDeep(b)), [])
+    : [arr]
+
 const getCSS = props => {
   if (!props.sx && !props.css) return undefined
   return theme => {
-    const sx = Array.isArray(props.sx) ? props.sx : [props.sx]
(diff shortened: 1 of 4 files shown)


packages/core/src/index.ts
```diff
@@ -16,9 +16,10 @@ export * from './types'
 const getCSS = props => {
   if (!props.sx && !props.css) return undefined
   return theme => {
-    const styles = css(props.sx)(theme)
+    const sx = Array.isArray(props.sx) ? props.sx : [props.sx]
+    const styles = sx.map(s => css(s)(theme))
```
jonavila: @jxnblk Do you have a preferred method for recursively flattening arrays? I realized after I opened the PR that Emotion recursively flattens arrays of styles and this doesn't account for that scenario.

lachlanjc: There are merge utilities in this file (see line 66)—is that what you’re looking for?

jonavila: @lachlanjc Thanks for the suggestion. Looks like the merge utility didn't have what i was looking for. I update a PR to do the flattening though.

jorisw:
Please make this PR part of 0.4.0.

PR #707, opened 2020-02-23
Fix useBorderBox

epilande:
Fix `useBoxSizing` -> `useBorderBox`.

https://github.com/system-ui/theme-ui/blob/cf5b00ad01a59b215b1ff6ef0cbbbf9893025a9e/packages/theme-provider/src/index.js#L12


jxnblk:
Thanks!

PR #708, opened 2020-02-23
Convert @theme-ui/theme-provider to TypeScript

epilande:
Convert `@theme-ui/theme-provider` to TypeScript.

This PR needs https://github.com/system-ui/theme-ui/pull/707.

Related to #668.


packages/css/src/index.ts
```diff
@@ -237,5 +238,5 @@ export const css = (args: SystemStyleObject = {}) => (
     }
   }
 
-  return result
+  return result as SerializedStyles
```
epilande: We probably don't want to do this..but I couldn't figure out how to get this function to play nicely with the styles prop in the `Global` component (`@emotion/core`).

Emotion's css type definition: 
https://github.com/emotion-js/emotion/blob/2f608a2e038d9a3c21dc506fc19c6786b24a4bec/packages/css/types/index.d.ts#L19

Would love some suggestions 🙏

8eecf0d2: I don't have a local copy to play around with but I noticed `SerializedStyles` is originally from [@emotion/utils](https://github.com/emotion-js/emotion/blob/master/packages/utils/types/index.d.ts#L29-L34) and looks like:

```typescript
export interface SerializedStyles {
  name: string
  styles: string
  map?: string
  next?: SerializedStyles
}
```

so it looks like the initial `result` object on L204 would need `name` and `styles` properties?

epilande: Yeah it's incorrect. Thanks for pointing that out. I looked at what `result` contains and it's a css object.
I found a better type to use here and it keeps the `Global` component happy

hasparus:
LGTM.

I have two minor issues but I don't think we should address them here.

- There's a missing dependency on `theme-provider -> typography` which, I think, will fail microbundle after babel packages update.
- The `colorMode` and `setColorMode` could be merged to ContextValue in `@theme-ui/color-modes` instead of being defined in `core` ahead of time.

---

**Okay, how do I unnaprove?** There's a small issue I need to investigate.

packages/theme-provider/test/index.tsx
```diff
@@ -15,7 +14,7 @@ const renderJSON = el => renderer.create(el).toJSON()
 
 test('renders', () => {
   const json = renderJSON(
-    <ThemeProvider>
+    <ThemeProvider theme={{}}>
```
hasparus, line 17: Okay, the tests in other packages would also need this fix.
There are currently 7 failing on my machine before rebasing this branch on top of fresh master and 4 after rebase.

@jxnblk This is public API change. Does it make sense to use ThemeProvider without passing theme?

packages/theme-provider/src/index.ts
```diff
@@ -1,12 +1,17 @@
...
+} from '@theme-ui/core'
+import { css, Theme } from '@theme-ui/css'
 import { ColorModeProvider } from '@theme-ui/color-modes'
 import { MDXProvider } from '@theme-ui/mdx'
 import { Global } from '@emotion/core'
 
 const BodyStyles = () =>
   jsx(Global, {
-    styles: theme => {
+    styles: (theme: Theme) => {
```
hasparus, line 14: This fails in tests, because Global.styles argument is `unknown` and it's not assignable to `theme`.

We'd either need to do
```tsx
<Global<Theme> styles={correctlyTypedTheme => ({})}>
```
or 
```tsx
jsx(Global, {
  styles: emotionTheme => {
    const theme = emotionTheme as Theme
    //...
  },
});
```

Both are in fact assertions. The second one is just explicit. I'll leave it here as a suggestion.

```suggestion
    styles: emotionTheme => {
      const theme = emotionTheme as Theme
```

packages/theme-provider/test/index.tsx
jxnblk: That's a good question. There *is* the notion of a default theme, which is currently hard-coded (in various places knowing me) that includes `theme.breakpoints` and `theme.fontSizes`, but that currently doesn't require the `ThemeProvider` the way it's implemented. I think it's worth considering changing how that works – in #832 I suggested removing that altogether, in which case making the `theme` prop required, probably makes sense here

hasparus: I'll fix that unit test then and we'll make it required.

hasparus:
Hey @epilande, I took the liberty of fixing merge conflicts and few minor test problems.

@jxnblk @mxstbr
I followed up on https://github.com/system-ui/theme-ui/pull/708#discussion_r422386087, and made JS tests and examples reflect that the `theme` prop is now expected.

hasparus:
Sidenote:

https://github.com/system-ui/theme-ui/blob/076cd7b5cab2a8a58f96da9f148a33590588f169/packages/theme-provider/src/index.ts#L34-L58

I'm not sure if `keyof IntrinsicSxElements` here isn't too little. i.e. it's passed to MDXProvider so I think it can actually be any string. Whatever we pass will get used in MDX.

It should actually be okay now if `components` are defined separately and then passed,
but if someone defines them inline in JSX, she'll get an "extra keys" error.

@epilande Do you think it makes sense?

PR #709, opened 2020-02-23
refactor(gatsby-theme-code-recipes): convert to typescript

velein:
Part of #668

Converts `gatsby-theme-code-recipes` to Typescript. I've double checked if this running correctly on `docs` (docs are using this package as dependency)

Might be a good idea to make additional tweaks by adding `strict: true` and type `any` props. I can try add that if you don't have anything against it.


mxstbr:
> Might be a good idea to make additional tweaks by adding strict: true and type any props. I can try add that if you don't have anything against it.

Yes please!! :pray:

velein:
Done @mxstbr 
BTW sorry I've mistakenly set Brent as a reviewer and cannot add you now 🤦‍♂

packages/gatsby-theme-code-recipes/gatsby-config.js
```diff
@@ -7,5 +7,12 @@ module.exports = ({ path = 'src/recipes' } = {}) => ({
         path,
       },
     },
+    {
+      resolve: `gatsby-plugin-typescript`,
+      options: {
+        isTSX: true,
```
hasparus, line 13: Hey @velein, why do we need `isTSX` and `allExtensions` here?

packages/gatsby-theme-code-recipes/src/recipe.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { RecipeQuery } from '../types'
+
+type Recipe = Omit<RecipeQuery, 'body'> & {
+  children: React.ReactNode
+}
+
+/** Props such as a `name` etc. are passed down and used by other packages */
```
hasparus, line 8: This can also be shadowed, right?

packages/gatsby-theme-code-recipes/gatsby-config.js
hasparus: The more I read about these two options, the more I'm sure I'll actually turn them on in my personal projects, but I'm just not sure if we want to force them for `gatsby-theme-code-recipes` users.

velein: Fair point, just removed those

packages/gatsby-theme-code-recipes/src/recipe.tsx
velein: Looks like it is already shadowed by other package `packages/docs/src/gatsby-theme-code-recipes/recipe.js`. Anyway I will remove this comment because it doesn't do its job (TBH have no clue what I was trying to tell here).

lachlanjc:
Thank you for your contribution here! This package has since been removed from Theme UI—I’m sorry this waited so long with such an unfulfilling finish, but we’d absolutely welcome other contributions. <3

PR #710, opened 2020-02-24
Convert gatsby-theme-ui-layout to TypeScript

epilande:
Convert `gatsby-theme-ui-layout` to TypeScript.

Related to #668.

~Since this gets consumed as a Gatsby theme, I don't think it needs to be compiled.
I tested this out with `gatsby-theme-ui-blog` (which is still in JavaScript) and it works just like before 😃~


mxstbr:
> Since this gets consumed as a Gatsby theme, I don't think it needs to be compiled.

Are Gatsby themes able to compile TypeScript? What if they use TypeScript options the package doesn't support? 🤔

epilande:
> > Since this gets consumed as a Gatsby theme, I don't think it needs to be compiled.
> 
> Are Gatsby themes able to compile TypeScript? What if they use TypeScript options the package doesn't support? 🤔

epilande:
@mxstbr oops I lied. I tried it out with a clean gatsby install and it didn’t work.. I thought it was magic :P

jxnblk:
Sounds like it might make sense to leave the Gatsby plugins & themes in JS to avoid needing to compile them for the time being, right?

epilande:
@jxnblk yeah, another option could be to set up gatsby-config with gatsby-plugin-typescript, though this package is bare bones and  that it might not be worth it

jxnblk:
My guess is that it might not be worth it for the time being. One (small) nice thing about shipping the source code is that users can view more human-readable code in `node_modules`, which also is nice for some potential tooling we could build for shadowing (if that makes sense).

If Gatsby supports `.ts` in plugins/themes in the future, then I think it could make sense to convert these over.

epilande:
Yeah 💯 agree. Closing for now

PR #719, opened 2020-02-24
Experimental WIP: strict-ui

mxstbr:
Click [here](https://gist.github.com/mxstbr/f24105fe48a5263b2bc42699216f55d4) for background info. This is just an experimental prototype for me to test these ideas in practice.

This is what the _entire_ API currently looks like (essentially it's theme-ui with the restrictions outlined in the doc above and way less API surface area):

```JSX
/* @jsx jsx */
import { Grid, Flex, jsx, useStrictUI, ThemeProvider } from 'strict-ui';

<ThemeProvider theme={myTheme}>
  <Grid gap={2} columns={1}>
    <div sx={{ padding: 1 }} />
    <div sx={{ color: 'primary' }} />
  </Grid>
  <Flex flexDirection="column" gap={2}>
    <div sx={{ background: 'secondary' }} />
  </Flex>
</ThemeProvider>
```

**Known issues**

- [x] ~~Responsive values don't work~~
- [x] ~~`red: 'red'` in the theme breaks the "only theme values" detection~~
- [x] ~~Doesn't block shorthands (`mx`, `ml`) or x-y properties (`marginX`)~~
- [x] ~~Need to set `* { margin: 0; }` in reset~~
- [ ] There is no way to float images
- [ ] There is no way to specify `display: none` or `display: inline-block`?
- [ ] What to do about `margin: auto`? Can't center anything
- [ ] Fix undefined/null values
- [ ] `display: list-item` and padding-inline-start accepts non-theme values


packages/strict-ui/src/index.ts
```diff
@@ -4,7 +4,7 @@ import { Global } from '@emotion/core'
 import { scales, aliases, transforms } from './css'
 import sanitize from './sanitize.css'
 
-export { Grid, ThemeProvider, useThemeUI as useStrictUI } from 'theme-ui'
+export { Grid, Flex, ThemeProvider, useThemeUI as useStrictUI } from 'theme-ui'
```
jxnblk: 😎

packages/strict-ui/src/sanitize.css.ts
```diff
@@ -0,0 +1,214 @@
...
+  * Custom stuff
+  */
+  *,
+  *:before,
+  *:after {
+    box-sizing: border-box;
+  }
+  img {
+    display: inline-block;
+  }
```
chasemccoy: I'd definitely be careful with the CSS reset that you choose, because almost any change to this down the road could be breaking. CSS resets are a tricky subject and tough to get right. I've been experimenting with a modern one lately: https://github.com/chasemccoy/resetti

packages/strict-ui/src/sanitize.css.ts
```diff
@@ -0,0 +1,211 @@
+import { injectGlobal } from 'emotion'
+
+/*!
+ * Sanitize 2.0.0 (http://git.io/sanitize)
+ * Licensed under the MIT License.
+ */
+// Copied from https://cdnjs.cloudflare.com/ajax/libs/sanitize.css/2.0.0/sanitize.min.css
+injectGlobal`
```
joe-bell: Maybe there's a way to approach this reset more dynamically, rather than a blanket global style? This way you'll be able to use `strict-ui` components alongside others.

So for instance in my `Stack` and `Inline` component, when it's rendered as an  `ol` or `ul` I make use of a shared reset:
```tsx
export const Stack: React.FC<StackProps> = ({ gap = 3, ...props }) => (
  <div
    {...props}
    sx={{
      /* …stack styles */
      /* …reset returns the following: */
      /* Reset: ensure direct child list-items render without bullets */
      "& > li": {
        listStyleType: "none",
        "&:before": {
          position: "absolute",
          content:
            '"\\200B"' /* Add zero-width space to prevent VoiceOver disable */,
        },
      },
    }}
  />
);
```

Might be worth at taking a look at how @markdalgleish tackled this in Braid!
https://twitter.com/markdalgleish/status/1223161650246602752

joe-bell:
Just a variant idea for if `flexWrap: wrap` is defined:

```ts
const inline = (gap: number = 16) => ({
  /* styles to add to the parent */
  overflow: 'hidden',
  alignItems: 'center',
  margin: (gap / 2) * -1,
  /* styles for the children */
  '& > *': {
    margin: gap / 2,
  },
});
```

Then we can render elements nicely **inline** like so:

Mar-20-2020 11-32-45

joe-bell:
Just an update: I've actually added that `Inline` component above (and some other flex-based layout primitives) to a package called [**raam**](https://github.com/joe-bell/raam).

- Repo: [github.com/joe-bell/raam](https://github.com/joe-bell/raam)
- Docs: [raam.joebell.co.uk](https://raam.joebell.co.uk)

Please feel free to extend upon this with your stricter type strategy! I'd be happy to help out

lachlanjc:
@mxstbr Hey! What are you thinking about this PR now?

@hasparus can follow up on what’s happening with our TypeScript support right now.

mxstbr:
Going to close it for now as it was a nice experiment but would need a lot more work to be usable and I don't have the time to work on this!

PR #721, opened 2020-02-25
fix(@theme-ui/prism): Fix code block line height bug in Safari

dcastil:
Safari renders a small space between inline-block elements, resulting in a variable height of lines which cause a scroll overflow depending on their width.

On iOS this results in the font size of each token being scaled up individually depending on the width of the token element. On desktop Safari the font size stays consistent and the text causes a line break.

This PR fixes that issue by removing `display: inline-block;` from all tokens except single empty elements in a line in order to preserve empty lines.

Screenshot desktop Safari:

Screenshot 2020-02-25 at 04 20 15

Screenshot iPhone:

IMG_A6C3791072A9-1

### DIFF
--- a/packages/prism/src/index.js
+++ b/packages/prism/src/index.js
@@ -24,15 +24,12 @@ export default ({ children, className: outerClassName, title, ...props }) => {
         <Styled.pre className={`${outerClassName} ${className}`} style={style}>
           {tokens.map((line, i) => (
             <div {...getLineProps({ line, key: i })}>
-              {line.map((token, key) => {
-                const tokenProps = getTokenProps({ token, key })
-
-                if (line.length === 1 && !tokenProps.children) {
-                  tokenProps.sx = { display: 'inline-block' }
-                }
-
-                return <span {...tokenProps} />
-              })}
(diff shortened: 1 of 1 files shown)


lachlanjc:
Whatttt I’ve always wondered about this issue!! Thank you for the solution. Very useful.

mcs07:
I also encountered this issue. I have used an alternative fix that is slightly more concise - just change line 30 to:

```js
sx={token.empty ? { display: 'inline-block' } : undefined}
```

dcastil:
@mcs07 `token.empty` is always `undefined` in my case. Which version of `prism-react-renderer` do you use?

mcs07:
I believe v1.0.2. Looks like `token.empty` was added here https://github.com/FormidableLabs/prism-react-renderer/pull/6. Not sure what release.

dcastil:
Ah sorry, I was wrong. I accidentally tested it on the wrong code block which didn't contain any empty lines. I'll change the code to your approach. 😊

jxnblk:
Thanks!

PR #722, opened 2020-02-25
feat: Update Tailwind preset with new transforms/transitions

LekoArts:
Tailwind recently added transitions/transforms, so I updated the preset.
Reference: https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js

I also updated some other parts to align it more with the defaultConfig.
Also pulled out the values in `sizes` for easier reference (to the original) and re-usability in other instances.

The additional values AFAIK need to be used in a functional way then, e.g.

```jsx
<div
  sx={{
    transition: theme => theme.transitionProperty.default
  }}
</div>
```

### DIFF
--- a/packages/preset-tailwind/src/index.ts
+++ b/packages/preset-tailwind/src/index.ts
@@ -257,8 +257,8 @@ export const colors = {
 
 export const baseFonts = {
   sans:
-    '-apple-system, BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,"Noto Sans",sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji"',
-  serif: 'Georgia, Cambria, "Times New Roman", Times, serif',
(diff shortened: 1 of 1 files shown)


packages/preset-tailwind/src/index.ts
```diff
@@ -535,6 +550,85 @@ export const styles = {
...
+    '-3': '-3deg',
+    '0': '0',
+    '3': '3deg',
+    '6': '6deg',
+    '12': '12deg',
+  },
+}
+
+export const transitions = {
+  transitionProperty: {
```
jxnblk: We don't have a really solid specification for transition properties in the theme spec, but wondering if these should be named e.g. `transitions.property` to avoid redundant naming conventions??

LekoArts: Yeah, that's a good idea :)

LekoArts: Will do that in a week

LekoArts:
@jxnblk Done :)

jxnblk:
Thanks! This looks great!

PR #726, opened 2020-02-26
Add missing peer dependencies to support pnp

### LINKED ISSUE #725: Add support for pnp and yarn v2 (berry)
Currently, theme-ui doesn't work with pnp/yarn berry because it doesn't list peer dependencies properly in multiple packages such as theme-provider and core. It also depends on emotion and mdx-js which have similar issues.

Fix the peer dependencies by adding them to the suitable packages and ask dependencies to fix it as well.

[packageExtensions in .yamlrc.yml ](https://www.reddit.com/r/javascript/comments/f4r213/askjs_is_someone_using_yarn_v2_in_business/fht7emv/) but it gets out of hand and would be more helpful to just have the project fix it.

It would be good if someone can take it otherwise I will try to fix it and PR myself in a few days.

searchableguy:
Fixes #725

### DIFF
--- a/packages/theme-provider/package.json
+++ b/packages/theme-provider/package.json
@@ -12,7 +12,9 @@
     "@emotion/core": "^10.0.0",
     "@theme-ui/color-modes": "^0.3.1",
     "@theme-ui/core": "^0.3.1",
-    "@theme-ui/mdx": "^0.3.0"
+    "@theme-ui/css": "^0.3.1",
+    "@theme-ui/mdx": "^0.3.0",
+    "react": "^16.12.0"
   },
   "repository": "system-ui/theme-ui",
   "author": "Brent Jackson",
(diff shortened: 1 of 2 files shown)


packages/theme-provider/package.json
```diff
@@ -12,7 +12,9 @@
     "@emotion/core": "^10.0.0",
     "@theme-ui/color-modes": "^0.3.1",
     "@theme-ui/core": "^0.3.1",
-    "@theme-ui/mdx": "^0.3.0"
+    "@theme-ui/css": "^0.3.1",
```
jxnblk: Thanks for looking into this, but since, for example, `@theme-ui/core` depends on `@theme-ui/css`, I don't think these should be added as dependencies to this package

Iwillaccept: but you are importing that package inside theme-provider hence yarn complaining about you not listing it it dependencies.
https://github.com/system-ui/theme-ui/blob/e3a24cee136553ee62ea777ca99037612b5f5ff1/packages/theme-provider/src/index.js#L2

PR #727, opened 2020-02-27
Link fix in docs.

MarkosKon:
Fixed the link to `@theme-ui/prism` from [Styling MDX Content](https://theme-ui.com/styling-mdx#syntax-highlighting). The redirect works only if you open it in a new tab.


jxnblk:
Thanks!

PR #741, opened 2020-03-08
Update theming docs

jxnblk:


packages/docs/src/pages/index.mdx
```diff
@@ -7,9 +7,12 @@ import Hexo from '../components/hexo'
 
 <Hexo />
 
-Theme UI
+# Theme UI: The Design Graph Framework
```
mxstbr: Interesting re-brand, I like it 👍 Let's see if "Design Graph" sticks

jxnblk:
I'm not sure what I did to break the build, but can try to dig in later

PR #748, opened 2020-03-09
Add Gatsby doc to side bar

laurieontech:
In the long term, we should probably add more specifics to this doc so we can link from .org. However, first step is to make it more discoverable.

### DIFF
--- a/packages/docs/src/sidebar.mdx
+++ b/packages/docs/src/sidebar.mdx
@@ -1,5 +1,4 @@
-- Getting Started
-  - [Getting Started](/getting-started)
+- [Getting Started](/getting-started)
   - [Getting Started with Gatsby](/getting-started/gatsby)
 - [Theming](/theming)
 - [The `sx` Prop](/sx-prop)
--- a/packages/docs/src/sidebar.mdx
+++ b/packages/docs/src/sidebar.mdx
@@ -1,4 +1,6 @@
-- [Getting Started](/getting-started)
+- Getting Started
+  - [Getting Started](/getting-started)
+  - [Getting Started with Gatsby](/getting-started/gatsby)
 - [Theming](/theming)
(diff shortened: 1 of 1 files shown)


packages/docs/src/sidebar.mdx
```diff
@@ -9,6 +9,7 @@
 - [API](/api)
 - [Theme Specification](/theme-spec)
 - [Customize](/customize)
+- [Theme UI and Gatsby](/getting-started/gatsby)
```
jxnblk: Thanks! I was originally thinking this could be nested under *Getting Started*, similar to what https://mdxjs.com does. We could add similar pages for CRA, Nextjs, etc. What do you think of that approach?

laurieontech: Seems perfectly reasonable to me. I wasn't sure how prominent you wanted it to be and was fine with anything. Nested under getting started is ideal! Though I'm hopeful this doc can expand beyond getting started content?

jxnblk: I'm guessing we'd eventually want more Gatsby-related content in the docs -- maybe there could be an entire section dedicated to usage in Gatsby -- but seems like there'd still be value in having a getting started page for Gatsby, since using the plugin is a bit different than using the `ThemeProvider` directly

laurieontech: Sounds good. We can start there and then look at additional docs later.

laurieontech: Took a stab at an idea.

jxnblk: Love it! Thanks! Feel free to open some more issues/PRs if you have any direction for how to start fleshing out more Gatsby-specific docs (i.e. if we want guides/sections related to usage in Gatsby themes, etc.)

PR #749, opened 2020-03-10
+ example filename to mdx-layout-components.mdx

tomByrer:
Your documentation is very good, but I personally find it hard to link 2 examples together if there is no file name mentioned in the first.
My poor brain was looking for `export const MyLayout ...`.

### DIFF
--- a/packages/docs/src/pages/guides/mdx-layout-components.mdx
+++ b/packages/docs/src/pages/guides/mdx-layout-components.mdx
@@ -27,6 +27,7 @@ export default props => (
 To add styles to this component, add a `theme` prop to the `ThemeProvider` and an `sx` prop to the `div`.
 
 ```jsx
+// nile name: MyLayout.js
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/guides/mdx-layout-components.mdx
```diff
@@ -27,6 +27,7 @@ export default props => (
 To add styles to this component, add a `theme` prop to the `ThemeProvider` and an `sx` prop to the `div`.
 
 ```jsx
+// nile name: MyLayout.js
```
jxnblk: Thanks for the suggestion, but I think we should solve this for all code blocks using a [code block meta string](https://mdxjs.com/guides/live-code#code-block-meta-string)

```suggestion
```jsx filename=Layout.js
```

lachlanjc:
@hasparus Close in favor of #1120?

PR #757, opened 2020-03-12
Discussion: Add `typecheck` scripts

hasparus:
This PR moves typechecking out of test to a separate package.json script.
This is to ensure package tsconfigs are respected instead of typechecking everything with one config.

---

Hey @jxnblk, this is what I meant in https://github.com/system-ui/theme-ui/issues/668#issuecomment-598341799 and https://github.com/system-ui/theme-ui/issues/668#issuecomment-598352395.

Feel free to close and totally disregard this.
I'm totally cool with splitting this into separate PRs / commits.

In the meantime, I encountered two problems, which might be worth merging.
I'll elaborate in comments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "serve": "yarn workspace docs serve --port 8000",
     "clean": "lerna run clean",
     "format": "prettier --write \"**/*.js{,on}\" \"**/*.md\"  \"**/*.mdx\"",
-    "test": "lerna run typecheck && jest",
+    "typecheck": "lerna run typecheck",
+    "test": "yarn typecheck && jest",
     "logo": "yarn workspace docs logo",
     "dev:chrome": "yarn workspace @theme-ui/chrome dev",
     "dev:editor": "yarn workspace @theme-ui/editor dev"
@@ -39,7 +40,9 @@
     "react-test-renderer": "^16.8.6",
     "ts-jest": "^25.2.0"
   },
-  "resolutions": {},
+  "resolutions": {
+    "@types/vfile-message": "1.0.1"
+  },
   "jest": {
     "testMatch": [
       "**/packages/**/test/*.{js,ts,tsx}"
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "serve": "yarn workspace docs serve --port 8000",
     "clean": "lerna run clean",
     "format": "prettier --write \"**/*.js{,on}\" \"**/*.md\"  \"**/*.mdx\"",
-    "test": "jest",
+    "test": "lerna run typecheck && jest",
     "logo": "yarn workspace docs logo",
     "dev:chrome": "yarn workspace @theme-ui/chrome dev",
     "dev:editor": "yarn workspace @theme-ui/editor dev"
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -39,22 +40,10 @@
     "react-test-renderer": "^16.8.6",
     "ts-jest": "^25.2.0"
   },
-  "resolutions": {},
+  "resolutions": {
+    "@types/vfile-message": "1.0.1"
```
hasparus: New @types/vfile-message (v2) are stub package.
We end up with old (non-TS) vfile in node_modules,
what results in error on `skipLibCheck: false`.

We have `"strict": true` in root tsconfig.json, what includes `skipLibCheck: false`.
It means we're interested if the types of libs we use. I'm a big fan of monorepos, but it's pretty easy to end up with different versions of the same package :/ 

I'm not sure if this happens to anyone else though. This part probably needs a closer look.

The interesting part of yarn.lock.
```
remark@^10.0.1:
  version "10.0.1"
  resolved "https://registry.yarnpkg.com/remark/-/remark-10.0.1.tgz#3058076dc41781bf505d8978c291485fe47667df"
  integrity sha512-E6lMuoLIy2TyiokHprMjcWNJ5UxfGQjaMSMhV+f4idM625UjjK4j798+gPs5mfjzDE6vL0oFKVeZM6gZVSVrzQ==
  dependencies:
    remark-parse "^6.0.0"
    remark-stringify "^6.0.0"
    unified "^7.0.0"

unified@^7.0.0:
  version "7.1.0"
  resolved "https://registry.yarnpkg.com/unified/-/unified-7.1.0.tgz#5032f1c1ee3364bd09da12e27fdd4a7553c7be13"
  integrity sha512-lbk82UOIGuCEsZhPj8rNAkXSDXd6p0QLzIuSsCdxrqnqU56St4eyOB+AlXsVgVeRmetPTYydIuvFfpDIed8mqw==
  dependencies:
    "@types/unist" "^2.0.0"
    "@types/vfile" "^3.0.0"
    bail "^1.0.0"
    extend "^3.0.0"
    is-plain-obj "^1.1.0"
    trough "^1.0.0"
    vfile "^3.0.0"
    x-is-string "^0.1.0"

"@types/vfile@^3.0.0":
  version "3.0.2"
  resolved "https://registry.yarnpkg.com/@types/vfile/-/vfile-3.0.2.tgz#19c18cd232df11ce6fa6ad80259bc86c366b09b9"
  integrity sha512-b3nLFGaGkJ9rzOcuXRfHkZMdjsawuDD0ENL9fzTophtBg8FJHSGbH7daXkEpcwy3v7Xol3pAvsmlYyFhR4pqJw==
  dependencies:
    "@types/node" "*"
    "@types/unist" "*"
    "@types/vfile-message" "*"

"@types/vfile-message@*":
  version "2.0.0"
  resolved "https://registry.yarnpkg.com/@types/vfile-message/-/vfile-message-2.0.0.tgz#690e46af0fdfc1f9faae00cd049cc888957927d5"
  integrity sha512-GpTIuDpb9u4zIO165fUy9+fXcULdD8HFRNli04GehoMVbeNq7D6OBnqSmg3lxZnC+UvgUhEWKxdKiwYUkGltIw==
  dependencies:
    vfile-message "*"
```

packages/components/index.d.ts
```diff
@@ -0,0 +1,335 @@
+import * as React from 'react';
```
hasparus: image

:point_up: this is why people start using ReasonML. `SxProps` and `SxProps` are not identical. Cool.

The error is on my side.
1. I've added `@types/theme-ui__components` to `theme-ui` deps to export types from the root package, because adopting TS in `@theme-ui/components` is a lot more work.
2. `@types/theme-ui__components` imported `SxStyleProp` from `@types/theme-ui`.
3. We end up with four lines with `extends SxProps` instead of two, as we want to.

image

---

Hey @kripod! I copied `@types/theme-ui_components` here.
When v0.4 is released with TypeScript support we can get rid of the package from DefinitelyTyped (i.e.change it to a stub logging "@theme-ui/components provides its own definitions").

hasparus:
Similar PR was already merged.

PR #765, opened 2020-03-16
fix(@theme-ui/css): add `size` to AliasesCSSProperties

hasparus:
Defined in https://github.com/system-ui/theme-ui/blob/ca9c2261d8a1168233933682769b038f59396896/packages/css/src/index.ts#L52

Missing from https://github.com/system-ui/theme-ui/blob/6b93a303bb622e01044e692c8ec845fe35d89359/packages/css/src/types.ts#L57

---

### DIFF
--- a/packages/css/src/types.ts
+++ b/packages/css/src/types.ts
@@ -299,6 +299,16 @@ interface AliasesCSSProperties {
    * @see https://developer.mozilla.org/docs/Web/CSS/padding-bottom
    */
   paddingY?: StandardCSSProperties['paddingTop']
+  // TODO: Move me to `MultiplesCSSProperties type and colocate it with the
+  // multiples object possibly.
+  /**
+   * The **`size`** is a shorthand property for CSS properties **`width`** and **`height`**.
+   *
+   * @see https://theme-ui.com/sx-prop#theme-aware-properties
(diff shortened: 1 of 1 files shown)


hasparus:
hey @jxnblk, sorry pinging you, but I just encountered this problem again, totally forgot I created this PR and started making the same PR again :D

I can get rid of the TODO comment, if you'd like.
Let me explain what I meant by it, and we won't need it anymore.

The type `AliasesCSSProperties` currently describes the the props originating from both `aliases` and `multiples` objects. This originates from DefinitelyTyped, and it might be tad confusing.

I think creating `MultiplesCSSProperties` could help; possibly colocating it with `multiples` object could also be helpful and make inconsistencies easier to spot. `MultiplesCSSProperties` could even be derived from `multiples`, but we'd lose the helpful comments in user tooltips and the type to do this wouldn't be the most maintainable thing, so I wouldn't go that far.

Let me know what you think :)

jxnblk:
I'm probably not the best person to answer this, but the changes LGTM. I'd defer to someone with more TS experience on this, cc @mxstbr

jxnblk:
Let's go ahead and get this in to keep things rolling; the TODO can be addressed in a follow-up PR

PR #778, opened 2020-03-17
Remove non-existent export

dburles:
Looks like this was removed at some point already but had accidentally returned.

### DIFF
--- a/packages/theme-ui/src/index.ts
+++ b/packages/theme-ui/src/index.ts
@@ -17,7 +17,6 @@ export * from '@theme-ui/components'
 export {
   css,
   get,
-  ColorMode,
   CSSObject,
   CSSProperties,
   CSSPseudoSelectorProps,


dburles:
@jxnblk just a quick ping on this minor update

packages/theme-ui/src/index.ts
```diff
@@ -17,7 +17,6 @@ export * from '@theme-ui/components'
 export {
   css,
   get,
-  ColorMode,
```
jxnblk, line 20: I *think* this might be a TypeScript type, but not sure. Can anyone clarify? cc @mxstbr

hasparus: yup, if you hover it in a text editor, this ugly thing shows up

image

and if you scroll down, there's a comment which is a bit too long and outdated. It actually references the old ColorMode component

image

I'll give it some more attention and update it.

Issue #702, opened 2020-02-21
Option to disable LocalStorage usage


8eecf0d2:
**Is your feature request related to a problem? Please describe.**
The [colorModes](https://theme-ui.com/color-modes) API writes `theme-ui-color-mode: "default"` to localStorage when not being used, this seems to be unnecessary and _slightly_ aggressive behaviour.

**Describe the solution you'd like**
When not in use colorModes should not write to localStorage.

Additionally provide an option to disable localStorage usage for colorModes.

**Describe alternatives you've considered**
None.

**Additional context**
Based on issues https://github.com/system-ui/theme-ui/issues/512 and https://github.com/system-ui/theme-ui/issues/553 it seems like localStorage is not a requirement for Theme UI to function correctly.

If this is acceptable I'm happy to make a PR.

hew:
This would be a nice feature to be able to turn off. I spun up my app on a new domain and it took me a while to figure out why all my colors were all wonky.

eschaefer:
+1 for this since `localStorage` keeping track of a color mode is really problematic for our use-case of migrating slices of our app to theme-ui.

When you have a nested `<ThemeProvider>`, it will override the `theme-ui-color-mode` value, and make for some pretty wonky stuff happening if you end up on another page that has a different color mode.

8eecf0d2:
I completely forgot about this but just saw it pop up as an autocomplete option with Typescript haha 😄 

Thanks a bunch @eschaefer for getting this in!

PR #782, opened 2020-03-18
Option to not use localStorage to persist the color mode

eschaefer:
Addresses the concerns brought up in https://github.com/system-ui/theme-ui/issues/702.

By default, `localStorage` is still used to persist the color mode. 

Adding a `useLocalStorage` option to the theme config could disable this persistence. Very useful for folks migrating disconnected pieces of an app to `theme-ui`, and stops the notorious color mode switching.

I have also added a test + updated the docs.

### DIFF
--- a/packages/docs/src/pages/color-modes.mdx
+++ b/packages/docs/src/pages/color-modes.mdx
@@ -156,7 +156,7 @@ modes on the same domain.
 
 ```js
 {
-  noLocalStorageForColorMode: true,
+  useLocalStorage: false,
   colors: {
     text: '#000',
     background: '#fff',
--- a/packages/docs/src/pages/color-modes.mdx
+++ b/packages/docs/src/pages/color-modes.mdx
@@ -148,7 +148,7 @@ If you do not have a color mode named `dark` or `light`, this will have no effec
 }
 ```
 
-### Disable setting color mode on `localStorage`
+### Disable persisting color mode on `localStorage`
 
 In some cases, it is not desirable to have the color mode set as a `localStorage` value. 
 The common use-case being that disconnected slices of your app use different color 
--- a/packages/docs/src/pages/color-modes.mdx
+++ b/packages/docs/src/pages/color-modes.mdx
@@ -148,3 +148,25 @@ If you do not have a color mode named `dark` or `light`, this will have no effec
 }
 ```
 
+### Disable setting color mode on `localStorage`
+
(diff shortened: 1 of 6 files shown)


packages/docs/src/pages/theming.mdx
```diff
@@ -181,6 +181,7 @@ Flag | Default | Description
 `initialColorModeName` | `'default'` | The key used for the top-level color palette in `theme.colors`
 `useColorSchemeMediaQuery` | `false` | Initializes the color mode based on the `prefers-color-scheme` media query
 `useBorderBox` | `true` | Adds a global `box-sizing: border-box` style
+`noLocalStorageForColorMode` | `false` | Disables the color mode being loaded from `localStorage`
```
jxnblk: Thanks for looking into this. I think to remain grammatically consistent with the other options, this could be `useLocalStorage`, with a default to `true`

eschaefer: Totally makes sense. Will update to match the other options' grammar.

eschaefer: Alright, updated!

packages/docs/src/pages/color-modes.mdx
```diff
@@ -148,3 +148,25 @@ If you do not have a color mode named `dark` or `light`, this will have no effec
 }
 ```
 
+### Disable persisting color mode on `localStorage`
+
+In some cases, it is not desirable to have the color mode set as a `localStorage` value. 
+The common use-case being that disconnected slices of your app use different color 
+modes on the same domain.
```
jxnblk: I'd prefer not to get into the *why* someone would want to disable `localStorage` to avoid any sort of judgemental sounding language, if that makes sense
```suggestion
To disable `localStorage` add the `useLocalStorage: false` flag to your theme.
```

cainvommars:
Would be great to get this bug fix moving!

eschaefer:
@jxnblk is there anything else you would like to see addressed or fixed? Is this good to go?

jxnblk:
Sorry for the delay! This LGTM -- thanks!

jonsherrard:
Was this ever published?

clairefritz:
@jonsherrard This was released in `v0.4.x` if I'm not mistaken. We were actually using the release candidate.

jonsherrard:
Thanks @clairefritz! Managed to get it working with `npm i @theme-ui@next` which installs a `v0.4.x` version :)

PR #797, opened 2020-03-24
Add support for CSS Logical Properties

### LINKED ISSUE #794: CSS Logical Properties should be theme-aware
**Describe the bug**

We use [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) to help support RTL and vertical languages. They are not theme-aware, so it's cumbersome to use theme values with them. This indirectly encourages developers to use physical direction instead, which makes it harder to support different writing modes.

For example, our styles may use `marginInlineStart` instead of `marginLeft`, which would be the same in a LTR language like English, but would instead mean `marginRight` when the language reads right-to-left like Hebrew.

**To Reproduce**
Use sx like so:

```js
<Box sx={{
  marginInlineStart: 'medium'
}} />
```

**Expected behavior**
```css
margin-inline-start: '8px'; /* or whatever `theme.space.medium maps` to */
```

**Actual behavior**
```css
margin-inline-start: 'medium'; /* the literal string 'medium' */
```

**Additional context**

There are a lot of logical properties. I'm happy to do the work of adding them to theme-ui if you feel this is appropriate to include.

lachlanjc:
Pretty sure I got them all :)

Closes #794


packages/css/test/index.ts
```diff
@@ -160,6 +160,25 @@ test('handles all core styled system props', () => {
   })
 })
 
+test('handles css logical properties', () => {
```
folz: This feels a bit nitpicky, but I think it'd be worthwhile to include all the logical properties in a test since any one of them could regress.

peduarte:
Nice, good to see this 😍 

Just today I created this issue in Styled System, since I use `@styled-system/css` instead of Theme UI's.

👉  https://github.com/styled-system/styled-system/issues/1268

How do you feel about adding aliases to the margin and padding logical properties? I think since we have aliases for margin and padding it could make sense here too.

- `mbs`: `marginBlockStart`
- `mbe`: `marginBlockEnd`
- `mis`: `marginInlineStart`
- `mie`: `marginInlineEnd`
- `pbs`: `paddingBlockStart`
- `pbe`: `paddingBlockEnd`
- `pis`: `paddingInlineStart`
- `pie`: `paddingInlineEnd`

lachlanjc:
@jxnblk Thoughts on this? Should we add aliases for spacial logical properties?

folz:
A point of confusion I could see is that `marginBlock` is a valid rule, but `marginBottom` has already taken the `mb` abbreviation. `mbk` maybe?

`marginInline` could also be abbreviated `mi` if we go forward with logical abbreviations.

jxnblk:
I don't think we should worry about shorthands for this PR. @peduarte note that `@styled-system/css` is deprecated in favor of `@theme-ui/css` so this feature probably won't land in the old package

peduarte:
Ah, I had a feeling that this would be the case 🙂 

SGTM 👍

lachlanjc:
@jxnblk Are you looking for any changes here or is this good to merge?

jxnblk:
Sorry for the delay! This looks great, thanks for getting this together!

PR #802, opened 2020-03-26
Fix dropdown appearing under other combo boxes.

nick-at-hawkfish:
Because all comboboxes are set to `z-index: 2`, the dropdown from the top combobox (font family) appears under the other two comboboxes (font weight and line height).


jxnblk:
I haven't looked at this code recently, but ideally the positioned element shouldn't have any position properties when closed, which should prevent this issue and other potential style bugs related to CSS positions

lachlanjc:
Thanks for your contribution! This strikes me as a slightly heavy-handed fix though, that might causes various conflicts in people’s setups (what if this component is inside a modal, for example?). It’d be great to fix this issue but I’m in support of closing this @hasparus.

PR #803, opened 2020-03-26
docs: add advanced use case

jlengstorf:
This is something I ran into today, so I figured I'd add it to the docs for future googlers.

Love,
DenverCoder9


jxnblk:
Awesome, thanks!

PR #807, opened 2020-03-27
feat(@theme-ui/style-guide): convert to typescript

beerose:
This PR migrates `style-guide` package to TypeScript.

Part of #668.

### Known issues

In `TypeScale` component `fontSizes` prop is assumed to be an array. However, in `Theme` typings are saying that it can be either array or object. 
It will result in a runtime error if `TypeScale` gets an object instead of an array for `fontSizes`.


beerose:
Ping @mxstbr

hasparus:
@jxnblk I'm leaving a TODO comment here as part of the "defined Public API / strict schema" issue.

https://github.com/system-ui/theme-ui/pull/807/files#diff-f2c5834d0c95b77c1c950a5445e6376fR26

jxnblk:
@beerose thanks for the help with this! Looks great!

@hasparus thanks! TBH this particular package probably doesn't handle all the possible shapes a theme object could take, so I think leaving a TODO comment is fine for this PR

Issue #556, opened 2020-01-08
Add gap prop to Flex component


TheMightyPenguin:
**Is your feature request related to a problem? Please describe.**
I was looking into how to make a Stack layout, and found this recipe: https://theme-ui.com/recipes/stack.
This recipe won't work if the user needs/wants to support IE11, as that browser doesn't have support for `grid-gap`. Also, I think having to be aware of using either `display: flex` or `grid` to implement a Stack is not the ideal development experience. I think that things like `display: flex;` or `display: grid;` should be an implementation detail in most cases.

**Describe the solution you'd like**
Having a Stack component that handles those implementation details and just lets you specify a Space that belongs to your space scale, will be a flexible option that still behaves under your theme constraints. Something that looks like:

```jsx
// Items are spaced out using the element with index 3 on your spaces scale
<Stack space={3}>
  <div>One</div>
  <div>Two</div>
  <div>Three</div>
  <div>Four</div>
</Stack>
```

Not sure if that component fits into the principles of this library, but I saw an opportunity, and if it's wanted here I'd love to implement it! 😄 

**Describe alternatives you've considered**
We could do something similar to what's described here: https://css-tricks.com/css-grid-in-ie-css-grid-and-the-new-autoprefixer/, just to have the `Grid` component translate the `gridGap` prop as Autoprefixer does.

**Additional context**
I've been using the Stack component for a while and found it incredibly useful and reusable. I was heavily inspired by the Stack component on this design system by the team at Seek (https://seek-oss.github.io/braid-design-system/components/Stack)

lachlanjc:
I'm not sure if/how to implement this, but I'd totally use it if it was available. I oftentimes have a row of buttons, a vertical list of cards, etc & sometimes Grid isn't the right choice, so this could be great. Interested in @jxnblk's thoughts though.

TheMightyPenguin:
Yeah! I've been using this component for a while now, and it's really versatile. Adapts to a lot of use cases, and turns out stacking things with a space between is a pretty common task! (in my case at least)

jxnblk:
I might be missing something obvious, but what would this provide that the `Grid` component doesn’t?

TheMightyPenguin:
* support for IE11 without much hassle (probably implemented as `flex` under the hood)

TheMightyPenguin:
@jxnblk from what I see:
* Support for IE11 without much hassle (probably implemented as `flex` under the hood)
[...]

PR #809, opened 2020-03-29
feat(@theme-ui/components) Add a Stack component

janhoogeveen:
This PR is an attempt at solving https://github.com/system-ui/theme-ui/issues/556

- [X] Working component
- [ ] Finish documentation
- [ ] Add tests

**Prior art**
I've taken inspiration from both https://every-layout.dev/layouts/stack/ and https://github.com/seek-oss/braid-design-system.

**Technical approach**
I've started out using the `Grid` component directly, based on the [Stack recipe](https://theme-ui.com/recipes/stack). However, because a `Stack` component only flows in one direction and I'd like to support IE11 (which doesn't support grid-gap) I thought it would be better to use more traditional methods of building a Stack layout.

The [lobotomized owl selector](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/) by Heydon sounded like the right approach, but I wanted to avoid using brittle child css selectors within `__css` or `sx`, so I used `React.children.map` to pass extra props to nested items.

By passing a `marginTop` property to every child element except the first one, we realize a cross-browser stack component.

**No horizontal layout**
In theory it should be possible to make column based stacks by using `Flex` instead of `Box`. However, I feel you're better off using `Grid` directly if you need that. This component does one thing and it does one thing well, maintaining a vertical rhythm in your page by enforcing the same margins between components.

**Tradeoffs**
If the child component does not support the `marginTop` prop, the `Stack` component won't work at all. Not sure yet how to solve that, but I'm sure someone here can give advice on that if the rest of the approach is solid.

**Alternative approach**
Styling the child items directly from the Stack component with (pseudocode)
```
__css={{
  '> *+*': {
    mt: gap
  }
}}
```

Similar to how ReactUI solves this.
[Example](https://react-ui.dev/components/Stack)
[Source](https://github.com/siddharthkp/react-ui/blob/master/packages/react-ui/src/components/stack/index.js
)

**Future ideas**
- By using flexbox, you can introduce more advanced patterns such as using `margin: auto` on one of the children. I suspect that's too advanced for the scope of this component.
- Support a `direction` prop. It would mean all components need to get a display prop of inline-block or flex?

### DIFF
--- a/packages/components/src/Stack.js
+++ b/packages/components/src/Stack.js
@@ -1,19 +1,25 @@
 import React from 'react'
-import Box from './Box'
+import { Box } from './Box'
 
 export const Stack = React.forwardRef(
-  ({ gap = 2, flow = 'row', ...props }, ref) => {
+  ({ gap = 2, children, ...props }, ref) => {
+    /** Make a copy of the children so we can pass the styles as props instead of using brittle CSS selectors */
+    const elements = React.Children.map(children, (element, index) => {
+      if (index === 0) {
+        return element
+      }
+
+      const newProps = {
+        marginTop: gap,
+      }
+
+      return React.cloneElement(element, newProps)
+    })
+
     return (
-      <Box
-        ref={ref}
-        {...props}
-        __themeKey="stacks"
-        __css={{
-          display: 'grid',
-          gridGap: gap,
-          gridAutoFlow: flow,
-        }}
-      />
+      <Box ref={ref} {...props} __themeKey="stacks">
+        {elements}
+      </Box>
     )
   }
 )
--- a/packages/components/src/Stack.js
+++ b/packages/components/src/Stack.js
@@ -2,24 +2,16 @@ import React from 'react'
 import { Box } from './Box'
 
 export const Stack = React.forwardRef(
-  ({ gap = 2, children, ...props }, ref) => {
-    /** Make a copy of the children so we can pass the styles as props instead of using brittle CSS selectors */
-    const elements = React.Children.map(children, (element, index) => {
-      if (index === 0) {
-        return element
-      }
-
-      const newProps = {
-        marginTop: gap,
-      }
-
-      return React.cloneElement(element, newProps)
-    })
-
-    return (
-      <Box ref={ref} {...props} __themeKey="stacks">
-        {elements}
-      </Box>
-    )
-  }
+  ({ gap = 2, children, ...props }, ref) => (
+    <Box
+      ref={ref}
+      {...props}
+      __themeKey="stacks"
+      __css={{
+        display: 'grid',
+        gridGap: gap,
+      }}>
+      {children}
+    </Box>
+  )
 )
--- a/packages/components/src/Stack.js
+++ b/packages/components/src/Stack.js
@@ -1,0 +1,19 @@
+import React from 'react'
+import Box from './Box'
+
+export const Stack = React.forwardRef(
+  ({ gap = 2, flow = 'row', ...props }, ref) => {
+    return (
+      <Box
+        ref={ref}
+        {...props}
+        __themeKey="stacks"
+        __css={{
+          display: 'grid',
+          gridGap: gap,
+          gridAutoFlow: flow,
+        }}
+      />
+    )
+  }
+)
--- a/packages/components/src/index.js
+++ b/packages/components/src/index.js
@@ -1,6 +1,7 @@
 export { Box } from './Box'
 export { Flex } from './Flex'
 export { Grid } from './Grid'
+export { Stack } from './Stack'
 export { Button } from './Button'
 export { Link } from './Link'
 export { Text } from './Text'
--- a/packages/components/test/index.js
+++ b/packages/components/test/index.js
@@ -8,6 +8,7 @@ import {
   Card,
   Flex,
   Grid,
+  Stack,
   Heading,
   Image,
   Link,
--- a/packages/docs/src/pages/components/stack.mdx
+++ b/packages/docs/src/pages/components/stack.mdx
@@ -1,0 +1,71 @@
+---
+title: Stack
+---
+
+# Stack
+
+Use the `Stack` component as a layout primitive to add spaces between components. It allows you to define
+spacing in one direction, either `row` or `column`. This is very useful when stacking components within a bigger grid system. 
+
+```js
+import { Stack, Box, Grid } from 'theme-ui'
+```
+
+```jsx live=true
+
+<Grid
+  gap={3}
+  columns={2}>
+  <Box bg="muted">
+    <Stack gap={3} flow="row">
+      <Box bg="primary" color="white">Stack Item</Box>
+      <Box bg="primary" color="white">Stack Item</Box>
+      <Box bg="primary" color="white">Stack Item</Box>
+    </Stack>
+  </Box>
+  <Box bg='muted'>Box</Box>
+</Grid>
+```
+
+Change the spacing by using the gap property
+
+```js
+import { Stack, Box } from 'theme-ui'
+```
+
+```jsx live=true
+<Stack gap="3">
+  <Box bg="primary" color="white">Stack Item</Box>
+  <Box bg="muted">Stack Item</Box>
+  <Box bg="primary" color="white">Stack Item</Box>
+</Stack>
+```
+
+Want to stack items horizontally? Use the flow property
+
+```js
+import { Stack, Box } from 'theme-ui'
+```
+
+```jsx live=true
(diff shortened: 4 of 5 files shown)


robphoenix:
I'm really interested in this, and have tackled something similar in a work component library. I ended up here while looking into replacing my custom components with a library such as `@theme-ui/components`, and have run into the same trade-offs as you have here; namely adding the margin to native elements as well as elements that support the `marginTop` prop.

I initially used the same approach as [Braid's](https://github.com/seek-oss/braid-design-system/blob/master/lib/components/Stack/Stack.tsx), wrapping each child element in a div that has the margin value added to it, but have run into issues where I have needed to access the CSS of that containing div.

I see [Chakra UI](https://chakra-ui.com/stack) currently do the same as your approach here, with the addition of a `shouldWrapChildren` prop that can be used when elements that are unaffected by the `marginTop` prop are used, and which then does what the Braid Stack component does. Chakra UI however looks to be changing this approach to using the owl selector: 

https://github.com/chakra-ui/chakra-ui/issues/207
https://github.com/chakra-ui/chakra-ui/pull/277
https://github.com/chakra-ui/chakra-ui/blob/dev-ts/packages/layout/src/Stack.tsx#L110

I'm currently using an approach with emotion which works but hasn't been battle tested yet so is probably fragile (see below). I did try this with the theme-ui `sx` prop, but didn't get far. 

I do think it would be a real shame if the Stack component didn't work on elements that didn't understand the `marginTop` prop. 

```typescript
/**@jsx jsx */
import { jsx } from '@emotion/core'
import React from 'react'

import Flex, { FlexProps, FlexAlignment } from '../Flex/Flex'
import Box from '../Box/Box'
import { spacing } from '../../theme'

export type StackAlignment = `left` | `right` | `center`
type StackProps = FlexProps & {
  space?: number
  align?: StackAlignment
  shouldWrapChildren?: boolean
}

// Originally built off https://github.com/seek-oss/braid-design-system/blob/master/lib/components/Stack/Stack.tsx
// and subsequently inspired by https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Stack/index.js
const Stack: React.FC<StackProps> = React.forwardRef<HTMLElement, StackProps>(
  (props, ref) => {
    const {
      children,
      space = 0,
      align,
      as,
      shouldWrapChildren = false,
      ...rest
    } = props

    // https://github.com/emotion-js/emotion/issues/1102#issuecomment-446887725
    const cloneElement = (element: any, props: any) => {
      return jsx(element.type, {
        key: element.key,
        ref: element.ref,
        ...element.props,
        ...props,
      })
    }

    const validStackItems = React.Children.toArray(children).filter(
      React.isValidElement
    )

    const isList = as === 'ol' || as === 'ul'
    const stackItemComponent = isList ? 'li' : 'div'

    const alignOptions = {
      left: `flex-start`,
      right: `flex-end`,
      center: `center`,
    }

    return (
      <Flex
        ref={ref}
        as={as}
        direction="column"
        alignItems={align ? (alignOptions[align] as FlexAlignment) : `stretch`}
        {...rest}
      >
        {validStackItems.map((child: React.ReactElement, i) => {
          const isLastChild = i === validStackItems.length - 1
          const marginBottom = !isLastChild ? spacing[space] : ``

          const {
            props: { css, ...rest },
          } = child

          const childProps = {
            css: css ? [css, { marginBottom }] : { marginBottom },
            ...rest,
          }

          if (shouldWrapChildren || isList) {
            return (
              <Box
                key={i}
                as={stackItemComponent}
                marginBottom={!isLastChild ? space : `none`}
              >
                {child}
              </Box>
            )
          }

          // https://github.com/emotion-js/emotion/issues/1404
          if (css) return React.cloneElement(child, childProps)
          return cloneElement(child, childProps)
        })}
      </Flex>
    )
  }
)

export default Stack

```

jxnblk:
> If the child component does not support the marginTop prop, the Stack component won't work at all. Not sure yet how to solve that, but I'm sure someone here can give advice on that if the rest of the approach is solid.

IMO, it might make sense to use Theme UI's `jsx` and pass the `marginTop` to the `sx` prop (I *think* that would still work, but haven't tried it/may be thinking about this wrong).

The only huge issue I can imagine with using a child selector internally is increasing the specificity of the rule and not being able to easily override the margin value directly on a child element. Generally, I suspect that should be avoided, but would be useful in certain one-off cases and potentially lead to unexpected behavior...

janhoogeveen:
Before I pick this up again, @jxnblk what is your stance on browser support? If we consider IE11 dead I'm actually thinking pivoting back to css-grid is a much better idea.

jxnblk:
@janhoogeveen I think as long as we document the browser support along with the component (I don't think we're currently doing this with the Grid component...), I think it should be fine to use CSS grid layout for this component. The core Theme UI works with *any* CSS, and a component like this could be seen as an _opt-in_ feature

janhoogeveen:
> @janhoogeveen I think as long as we document the browser support along with the component (I don't think we're currently doing this with the Grid component...), I think it should be fine to use CSS grid layout for this component. The core Theme UI works with _any_ CSS, and a component like this could be seen as an _opt-in_ feature

Got it! I'll refactor to CSS grid to save us a lot of headache.

janhoogeveen:
I've pushed an update that uses css-grid. It's much simpler, but also much less powerful.

In the meantime, I've been looking into alternatives. A dirty solution but one that seems to work really well is wrapping all children in a `<Box>` and passing `mt` to the wrapper. The end-result is the same as using css-grid, but with a bit of hacking you could transform it into a horizontal layout by setting the top level container to Flex.

robphoenix:
I've recently been experimenting with a grid based Stack component.

```typescript
export const Stack = React.forwardRef<HTMLDivElement, StackProps>(
  (props, ref) => {
    const {
      gap = 0,
      align = `stretch`,
      row = false,
      sx,
      ref: _ref,
      ...rest
    } = props

    return (
      <Grid
        ref={ref}
        gap={gap}
        sx={{
          alignContent: row ? align : 'start',
          justifyContent: row ? 'start' : align,
          justifyItems: align,
          alignItems: align,
          gridAutoFlow: row ? 'column' : 'row',
          ...sx,
        }}
        {...rest}
      />
    )
  }
)
```

So far I like it more than my previous approach of wrapping children in a div with added margin; the layout isn't complicated by an additional element I can't style. It also feels philosophically more inline with what a Stack component is; a parent that manages it's childrens layout, rather than adding properties to them.

jxnblk:
Sorry for the delay on this! Just coming back to this again and looking at #850 -- this approach certainly looks a bit simpler implementation-wise, but you mention this approach being "much less powerful" -- could you elaborate on that and what you're thinking?

janhoogeveen:
> Sorry for the delay on this! Just coming back to this again and looking at #850 -- this approach certainly looks a bit simpler implementation-wise, but you mention this approach being "much less powerful" -- could you elaborate on that and what you're thinking?

No worries.

I think it boils down to this.

**Grid** - Implementation is small, but limited to vertical stacking. You'll need to add margin-left/margin-right to your content to create horizontal spacing.

**Flex** - Implementation is more verbose, but supports horizontal and vertical stacking. All whitespace can be handled by one Stack component.

The flex component has the added bonus of supporting flex' margin behavior. I haven't tried it out yet, but for instance you could add `margin-left: auto` to the last item in the stack and it will be right aligned.

jxnblk:
Thanks! This is helpful!

> Grid - Implementation is small, but limited to vertical stacking. You'll need to add margin-left/margin-right to your content to create horizontal spacing.

It's been a while since I've tried to wrap my head around all the possibilities with the grid layout API, but horizontal "stacking" seems like something that's doable, is it not? Also should a "Stack" component even need to run horizontally if there's already the Grid component?

That said, based on browser support + the nature of how flexible box layout module works (e.g. auto margin, etc), I'd be more inclined to suggest going the flexbox route, if you agree that makes sense.

janhoogeveen:
> Thanks! This is helpful!
> 
> > Grid - Implementation is small, but limited to vertical stacking. You'll need to add margin-left/margin-right to your content to create horizontal spacing.
> 
> It's been a while since I've tried to wrap my head around all the possibilities with the grid layout API, but horizontal "stacking" seems like something that's doable, is it not? Also should a "Stack" component even need to run horizontally if there's already the Grid component?

I thought the same thing but I couldn't get it working.

> That said, based on browser support + the nature of how flexible box layout module works (e.g. auto margin, etc), I'd be more inclined to suggest going the flexbox route, if you agree that makes sense.

Fully agree. Should we close this PR and continue the discussion in the other one? I'll see to it that I'll finish it up.

Issue #811, opened 2020-03-30
Contextual Class Names for base components


yuriybelike:
**Is your feature request related to a problem? Please describe.**
When I inspect the elements in the browser, only the "wrapper" React component name is present in the `class` attributes. Theme UI's base component names such as Heading or Image do not show up.

**Describe the solution you'd like**
Base components such as Heading or Image should show its name in the rendered element's `class` attribute for easier debugging.

**Additional context**

Source code

```tsx
/** @jsx jsx */
import React, { FunctionComponent } from "react";
import { jsx, Flex, Heading, Text, Image } from "theme-ui";

const UsersImporter: FunctionComponent = () => {
  return (
    <Flex
      as="article"
      data-testid="info-clip"
      sx={{
        flexDirection: "column",
        px: "35%",
        py: "6em",
        bg: "lightgray"
      }}
    >
      <Heading>My users</Heading>
      <Heading
        mb={2}
        sx={{
          textAlign: "center"
        }}
      >
        IMPORT YOUR USER
      </Heading>
      <Image alt="import information" src="https://picsum.photos/300/100" />
      <Text as="p">Lorem ipsum</Text>
    </Flex>
  );
};
```

Rendered HTML in the browser

```html
<article data-testid="info-clip" class="css-nxzrfe-UsersImporter">
  <h2 class="css-1r1c8td">My users</h2>
  <h2 class="css-1cytxyy-UsersImporter">IMPORT YOUR USER</h2>
  <img
    alt="import information"
    src="https://picsum.photos/300/100"
    class="css-9taffg"
  />
  <p class="css-vurnku">Lorem ipsum</p>
</article>
```

Expected HTML

```html
<article data-testid="info-clip" class="css-nxzrfe-UsersImporter css-nxzrfe-Flex">
  <h2 class="css-1r1c8td-Heading">My users</h2>
  <h2 class="css-1cytxyy-UsersImporter css-1cytxyy-Heading">IMPORT YOUR USER</h2>
  <img
    alt="import information"
    src="https://picsum.photos/300/100"
    class="css-9taffg-Image"
  />
  <p class="css-vurnku-Text">Lorem ipsum</p>
</article>
```

lachlanjc:
Yep, huge agree here. Let’s close as a duplicate of #1080 though @hasparus

Issue #800, opened 2020-03-26
Proposal: variant prop


dburles:
Expanding on #294 (and related to #706), I've been thinking about an idea of replacing the `variant` faux CSS property instead with a special `variant` prop.

As mentioned in #294, the variant API no longer becomes useful when dealing with components composed of multiple elements, this is a problem for both userland components (#294) and `@theme-ui/components` i.e. the `Select` component mentioned in #706. 

Because of this, you are forced to abandon use of the `variant` API and instead implement something akin to what this `variant` prop proposal offers. The downside with that in the current implementation is then your component library will have two API's for using variants, (`sx` variant and `variant` prop), unless as I mentioned, you decide to implement a variant prop in your component library for all components that require it (as a workaround). With this proposal, that would not be required. e.g:

```js
// Contrived example of the issue above
const Button = ({
  variant,
}) => {
  return (
    <button
      sx={{
        // ...
        variant: `buttons.Button.${variant}`,
      }}
    />
  )
};
```

In implementation of the variant prop, the example API @jxnblk mentioned in #706 could be altered as follows:

```js
const theme = {
  ...
  forms: {
    select: {
      // styles for select element
      select: {
        borderRadius: 5,
      },
      container: {
        // styles for root element
      },
      arrow: {
        // styles for arrow icon go here
      },
    },
  },
}
```

The `Select` component from `@theme-ui/components` can now reference the `variant` prop (as now can any userland components too, and also easily pick from the theme object too if they wish):

Rough mockup:
```js
export const Select = React.forwardRef((props, ref) => {
  const { theme } = useTheme();
  return (
    <Box
      {...getMargin(props)}
      sx={{
        display: 'flex',
        ...theme.forms.select.container,
      }}
    >
      <Box
        ref={ref}
        as="select"
        {...omitMargin(props)}
        __css={{
          display: 'block',
          width: '100%',
          p: 2,
          appearance: 'none',
          fontSize: 'inherit',
          lineHeight: 'inherit',
          border: '1px solid',
          borderRadius: 4,
          color: 'inherit',
          bg: 'transparent',
          ...theme.forms.select.select,
        }}
      />
      <DownArrow
        sx={{
          ml: -28,
          alignSelf: 'center',
          pointerEvents: 'none',
          ...theme.forms.select.arrow,
        }}
      />
    </Box>
  );
});
```

~~With some additional changes to the `Select` component, it would also be possible to support inline overrides, e.g:~~
[...]

PR #817, opened 2020-03-31
add display names to components

### LINKED ISSUE #1080: Improve Components DisplayNames properties
Hi there,

sometimes I find it extremely difficult to properly debug base components in `React Dev Tools` due to lack of `Display Name` (I think this is it?). For smaller layouts it will not be an issue, however sometimes we can construct more complex ones using only components from theme-ui. Debugging those can be a little bit of pain as some sample React's component tree can look like:

image

A little bit cleaner react's component tree to debug.

dburles:
React devtools and Emotion will now render the correct component names, for https://github.com/system-ui/theme-ui/issues/811


lachlanjc:
This is awesome! Any reason we can’t do it with [the simpler syntax](https://github.com/rebassjs/grid/blob/master/src/index.js#L25) though?

dburles:
For that instance it is due to the emotion `styled` API not reading it for whatever reason

dburles:
The function name approach is needed when using `forwardRef` https://reactjs.org/docs/forwarding-refs.html

jxnblk:
Rad, thanks! Would changing the Box component to not use the `styled` HOC affect this at all? (See https://github.com/system-ui/theme-ui/pull/692#issuecomment-600131921)

dburles:
No it should be okay, if it's just a regular component, it will only need to pass a named function to `forwardRef`. These changes will definitely conflict with #692. I think if this were merged first, it shouldn't be very difficult to reconcile.

yuriybelike:
Hey! Thanks for this PR related to my issue.

I may not be approaching this the right way, but when I inspect the components in the [Deploy preview](https://deploy-preview-817--theme-ui.netlify.com/) I can only see `-Box` prefixes, instead of `-Image`, `-Heading` and such.

image

dburles:
> Hey! Thanks for this PR related to my issue.
> 
> I may not be approaching this the right way, but when I inspect the components in the [Deploy preview](https://deploy-preview-817--theme-ui.netlify.com/) all I see are `-Box` suffixes, instead of `-Image`, `-Heading` and such.
> 
> image

Yeah that is odd, I'll have a look into it

dburles:
Ok after digging into it a bit, I think the best solution to ensure emotion renders the correct names is to add the label property to all components. I'll push a fix soon.

dburles:
> Rad, thanks! Would changing the Box component to not use the `styled` HOC affect this at all? (See [#692 (comment)](https://github.com/system-ui/theme-ui/pull/692#issuecomment-600131921))

I'm kinda second guessing using the emotion `label` option as they're not stripped in production, though it is one solution. I've gone down a bit of a rabbit hole and the indirection through the `Box` component is causing problems in that emotion always resolves to `Box` unless we otherwise specify a `label` on every component.

This should be able to be greatly simplified by writing components without reference to `Box` such as:

```js
/** @jsx jsx */
import React from 'react'
import { jsx, useThemeUI } from 'theme-ui'

export const Alert = React.forwardRef(function Alert(props, ref) {
  return (
    <div
      ref={ref}
      {...props}
      sx={{
        display: 'flex',
        alignItems: 'center',
        px: 3,
        py: 2,
        fontWeight: 'bold',
        color: 'white',
        bg: 'primary',
        borderRadius: 4,
        variant: 'alerts.' + props.variant
      }}
    />
  )
})
``` 

Some of these components use `styled-system` style props, which is another reason why they all inherit from the `Box` component, e.g.:

```js
<Alert variant='secondary' mb={2}>Secondary</Alert>
```

IMO, there is no need for this now that we have the `sx` prop, it is odd to have two API's and only adds confusion. This might also simplify the TS refactor over at #692. Also, this definitely plays into the `variant` prop idea also: #800.

dburles:
I think I'll roll with the `label` option for now as that'll ensure that emotion renders the correct names. I'll try have that up soon.

dburles:
Alright so I think this is ready! Emotion will now render the component names, it's unfortunate but they'll all be named like `css-191ogd4-Box-Text` for example (always with `Box`). So to summarise this PR, it includes two things:

1. React devtools will now render the component names
2. Emotion will generate classnames containing the component names

It would be even better if we could get CSS sourcemaps working too, the babel plugin will require changes to support theme-ui. Something to look at in future :)

packages/components/src/Alert.js
```diff
@@ -1,20 +1,23 @@
...
-      px: 3,
-      py: 2,
-      fontWeight: 'bold',
-      color: 'white',
-      bg: 'primary',
-      borderRadius: 4,
-    }}
-  />
-))
+export const Alert = React.forwardRef(function Alert(props, ref) {
```
jxnblk, line 4: For my understanding: is the named `function` syntax still necessary with the Emotion `label` property added below?

jxnblk: Also, holy smokes! I don't think I realized the `label` property would work more generally in the `sx` prop – that's probably worth putting in a guide in the docs or mentioning somewhere 🤔

dburles: > For my understanding: is the named `function` syntax still necessary with the Emotion `label` property added below?

Just for React devtools

jxnblk: Does the `displayName` static property no longer work for React devtools?

dburles: It does but just not when you’re using `forwardRef` https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

dburles: Just thought I'd double check it, it does indeed work with `forwardRef`! Strange, I am sure that it wasn't working at the time, anyway, I'd probably stick with this approach as it's what's recommended and is more concise.

atanasster:
Hi @dburles - this is a great PR. 

Can you please upgrade it to resolve conflicts and we also have some new components
- Paragraph
- Switch

Also whats the final status on displayName? 

I hope we can merge this soon, as @lachlanjc is working on a rewrite to remove emotion/styled and it seems there is some overlap here.

dburles:
@atanasster Thanks. I've just fixed the merge conflicts. Named functions serve the same purpose as `displayName`, so either way it doesn't make any difference. 

I think we should probably remove the `label` property:

1. It isn't stripped in production, which means additional data over the wire.
2. It's tedious to correctly handle alongside the `__css` prop on `Box`. For instance `CheckBoxChecked` (in `Checkbox`) which renders the `SVG` component has custom styles defined using the `__css` prop. The means in each lower level component (`SVG` in this case), for those styles to function we would have to merge `label` with those styles so that the styles are correctly applied. It's possible to do, but unnecessarily tedious. This is only a problem for such instances (which I think are few, but still...).

I'm not sure `label` is worth the additional complexity, at least with the current implementation with everything using `Box`.

If we can make a decision on that I am happy to remove it and at least sort out the naming in devtools for now. Emotion `label` can always be revisited later.

atanasster:
Thanks @dburles thats great

I would say remove label. You are correct this PR should be only about the devTools and we can decide about label in some other PR if its requested by users.

Would you mind switching to static displayName, its a bit more standard and will be also easier for the rework that @lachlanjc is doing.

dburles:
I can't find why `displayName` appears to be working now ([it didn't used to](https://github.com/facebook/react/issues/13703)), the [React docs](https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools) still recommend naming the functions instead of using `displayName` when using `forwardRef`. So I'm not sure which is incorrect.

atanasster:
@dburles i think displaying the component name looks cleaner too, no?

dburles:
I mean like, technically you shouldn't be able to use `displayName` on `forwardRef` components. I do recall it not working some time in the past, now for some reason it seems to work. I'm not sure what that means because the docs still recommend providing a named function to forwardRef (and not displayName). Unless I can confirm that `displayName` is now supported on `forwardRef` components I wouldn't recommend it.

atanasster:
I see you are correct it was recommended by bvaughn in that thread. You make the call, seems both work so we should be good.

dburles:
I'll aim to have the label removed by early next week.

atanasster:
Fantastic, really looking forward to have your PR merged.

For displayName - any choice you make is fine by me, but until then (monday) if @hasparus @lachlanjc have any notes about it here

hasparus (approved):
Great PR, @dburles! Thank you!

@atanasster let's merge this after the `label` removal.

atanasster:
@dburles - would you mind updating Paragraph and Switch, just so we dont forget to bring them up to date as well.

dburles:
@atanasster Whoops, nice spot. Just updated those, should be good to go then.

atanasster:
@dburles - thanks a lot - its ready to merge. Let's just wait @hasparus to confirm if we are merging only bug fixes to release a 0.6.0 stable, or if we are also merging features.

atanasster (approved):
Thanks a bunch, great PR

hasparus:
> Let's just wait @hasparus to confirm if we are merging only bug fixes to release a 0.6.0 stable, or if we are also merging features.

You wrote it like I'm some sort of authority.

This PR is open since March, and IMHO it's a bugfix, as I'd expect proper names in React DevTools.

hasparus:
Sorry it took so long to merge it, @dburles 🙏

PR #818, opened 2020-04-01
docs: added custom CacheProvider guide

deckchairlabs:
I came across the need to specify a different container element (iframe) for the styles generated by emotion, so providing this guide for anyone needing to do the same.


jxnblk:
Awesome, thanks!

Issue #783, opened 2020-03-18
Reduce bundle size?


dburles:
Since `0.3.0` there's been a [jump in the bundle size](https://bundlephobia.com/result?p=theme-ui@0.3.1). Note that a chunk of it can be ignored due to the addition of the `@emotion/core` dependency. I feel like the library at its core could be kept minimal and fundamentally about the theme and `sx` prop relationship. `@theme-ui/components` seems like an obvious one to keep stand-alone, however lately it seems it's now part of `theme-ui` which has definitely also added to the size of the bundle.

jxnblk:
1. The `theme-ui` package is meant to be the "batteries-included" version
2. It *should* support tree-shaking -- i.e. if you're not using the components, they shouldn't be included in the final bundle (but knowing how tree-shaking works in practice YMMV)
3. The `@theme-ui/core` package can be used if you want the `sx` prop without components or MDX

dburles:
> 3. The `@theme-ui/core` package can be used if you want the `sx` prop without components or MDX
I didn't know that you could use it that way, that would solve the concern with the bigger install size of the main package. Do you think it would be worth documenting? Perhaps it can just be added to the Packages section along with a note and link somewhere in Getting Started.

jxnblk:
Yeah, this could definitely use better documentation -- I think starting with the packages README.md would be a good place

dburles:
If we recommend this officially, do you see any potential issues down the line in then having to support users who have chosen to install this way?

jxnblk:
I think any of the packages available will follow semver and should be safe to use their public APIs. The core package was split off based on dependencies, and I don't imagine there will be a ton of breaking changes there

lachlanjc:
We have several ideas for reducing bundle size & active plans in this area, so I’m going to close this for now since we don’t have a specific objective in this issue

PR #820, opened 2020-04-02
Update documentation for core installation

dburles:
As discussed in #783, this adds additional documentation around installing the core package. Also improved the clarity of the getting started section a little.

### DIFF
--- a/packages/docs/src/pages/getting-started/index.mdx
+++ b/packages/docs/src/pages/getting-started/index.mdx
@@ -4,20 +4,13 @@ title: Getting Started
 
 # Getting Started
 
-The `theme-ui` package includes a recommended set of packages out of the box, including [MDX support](/packages/mdx) and a set of [primitive UI components](/packages/components).
-However, if you would prefer a minimal installation, you can instead start with the `@theme-ui/core` package. For more information, see the [core package documention](/packages/core).
-
-## Setup
-
-### 1. Install Theme UI:
+Install Theme UI.
 
 ```sh
 npm i theme-ui
 ```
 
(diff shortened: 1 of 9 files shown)


packages/docs/src/pages/getting-started/index.mdx
```diff
@@ -4,13 +4,20 @@ title: Getting Started
 
 # Getting Started
 
-Install Theme UI.
+The `theme-ui` package includes a recommended set of packages out of the box, including [MDX support](/packages/mdx) and a set of [primitive UI components](/packages/components).
```
jxnblk, line 7: Thanks for updating the core package readme, but I don't think the general *Getting Started* page should branch off from the happy-path installation guide. The idea here is to guide someone new to the library without requiring too many decisions to be made, if that makes sense. I think an addition guide could be added as an advanced use-case, but not necessary for this PR if you don't have time for that

dburles: Yeah I was 50/50 on that myself, I originally had it as a footnote at the bottom of the first section, but moved and expanded on it as I figured it was a good place to also explain what you get with the main package itself.

For now should I remove that section entirely or move it? I'm happy to work on an advanced guide but I am unsure on how much it could be filled out.

jxnblk: I'd say for now, undo the changes in the main getting started page, but feel free to start a stub for a guide or something similar -- doesn't need to be linked to in the nav if it's a WIP

jxnblk:
Thanks!

PR #821, opened 2020-04-03
Add ColorKeys component to docs

dburles:
Adds a `ColorKeys` component as there were two very similar tables representing the color keys, and they had diverged slightly.


jxnblk:
Thanks!

Issue #294, opened 2019-08-13
CSS properties only with sx prop and variant API thoughts


dburles:
Bringing across a set of comments I had made against [this PR](https://github.com/styled-system/styled-system/pull/684).

There is a fair case that it would make sense for the `sx` prop to only support CSS properties, and that was @jxnblk's original thought too. Especially so if support is added to [reference theme values within CSS shorthand](https://github.com/system-ui/theme-ui/issues/81).

The other thought that came out of this was then how to handle variants. It would mean that `variant` could no longer be accessed within the sx prop.
However, perhaps that's not too much of a problem. One thing I've found is that it quickly breaks down once the variant needs to affect multiple elements at once.

So in aiming to define a consistent API across a component library, I opted instead to always provide a variant prop (when needed), that way I can handle both cases. The problem is then a source of confusion around having the two options available. I haven't yet found a nice, (ideally declarative) approach to handling these kinds of variants. What are your thoughts?

In the world of the sx prop only allowing CSS properties, the variant API could be built around the variant prop and also better accomodate the ability to cover multiple elements at once.

Here's an example of a basic idea:

```js
const variants = {
  primary: {
    wrapper: {
      bg: 'primary',
    },
    heading: {
      color: 'white',
    },
    subtitle: {
      color: 'textWhiteMediumEmphasis',
    },
  },
  secondary: {
    wrapper: {
      bg: 'none',
    },
    heading: {
      color: 'text',
    },
    subtitle: {
      color: 'textMediumEmphasis',
    },
  },
};

const CardHeader = ({
  title,
  subtitle,
  variant,
  ...props
}) => {
  return (
    <Flex
      {...props}
      sx={{
        borderTopLeftRadius: 'large',
        borderTopRightRadius: 'large',
        ...variants[variant].wrapper,
      }}
    >
      <Box sx={{ flex: 1 }}>
        <Heading
          as="h6"
          sx={variants[variant].heading}
        >
          {title}
        </Heading>
        {subtitle && (
          <Text
            sx={variants[variant].subtitle}
          >
            {subtitle}
          </Text>
        )}
      </Box>
    </Flex>
  );
};
```

lachlanjc:
Forgive me if I’m misunderstanding, I’ve read your message a few times & not sure I’m getting it. Does the behavior you’re looking for here go beyond having array support in the variant prop? The example you’re showing seems like it should work fine with how Theme UI already works.

dburles:
[...]

PR #823, opened 2020-04-04
New variant API

dburles:
This is a work in progress and a bit of a proof of concept refactor to replace the `variant` `sx` property with the `variant` prop. See #800. This would be a breaking change, but theme-ui is not v1.0 yet 🔨😄. I've also refactored the `Box` and `Flex` components and removed the dependencies on both `styled-system` and `@emotion/styled`, which also addresses issues with #817 and #690. We could probably drop the usage of `Box` from all the other components now as it's not doing all that much. I've also decided to drop the use of the few styled-system props that were previously part of the `Box` component in favour of using the `sx` prop instead.

There is also a new API for components with multiple elements: `Select`, `Checkbox` and `Radio`. It's now possible to supply styles to each part. Such as for `Select`:

```js
const theme = {
  ...
  forms: {
    select: {
      // styles for select element
      select: {
        borderRadius: 5,
      },
      container: {
        // styles for root element
      },
      arrow: {
        // styles for arrow icon go here
      },
    },
  },
}
```

Todo:
- [ ] MDX
- [ ] Tests
- [ ] Update docs examples
- [ ] Update docs theme object
- [ ] Update other themes
- [ ] Remove variant handling from `css` function
- [ ] Make `useVariant` hook public? – could be quite useful for custom components.

### DIFF
--- a/packages/components/src/AspectRatio.js
+++ b/packages/components/src/AspectRatio.js
@@ -1,3 +1,5 @@
+/** @jsx jsx */
+import { jsx } from '@theme-ui/core'
 import React from 'react'
 import Box from './Box'
 
@@ -21,7 +23,7 @@ export const AspectRatio = React.forwardRef(function AspectRatio(
       />
       <Box
         {...props}
-        __css={{
+        sx={{
           position: 'absolute',
           top: 0,
           right: 0,
--- a/packages/components/src/AspectRatio.js
+++ b/packages/components/src/AspectRatio.js
@@ -1,8 +1,11 @@
 import React from 'react'
 import Box from './Box'
 
-export const AspectRatio = React.forwardRef(
-  ({ ratio = 4 / 3, children, ...props }, ref) => (
+export const AspectRatio = React.forwardRef(function AspectRatio(
+  { ratio = 4 / 3, children, ...props },
+  ref
+) {
+  return (
     <Box
       ref={ref}
       sx={{
@@ -29,4 +32,4 @@ export const AspectRatio = React.forwardRef(
       </Box>
     </Box>
   )
-)
+})
--- a/packages/components/src/AspectRatio.js
+++ b/packages/components/src/AspectRatio.js
@@ -10,12 +10,12 @@ export const AspectRatio = React.forwardRef(function AspectRatio(
   return (
     <Box
       ref={ref}
-      sx={{
+      __css={{
         position: 'relative',
         overflow: 'hidden',
       }}>
       <Box
-        sx={{
+        __css={{
           width: '100%',
           height: 0,
           paddingBottom: 100 / ratio + '%',
@@ -23,7 +23,7 @@ export const AspectRatio = React.forwardRef(function AspectRatio(
       />
       <Box
         {...props}
-        sx={{
+        __css={{
           position: 'absolute',
           top: 0,
           right: 0,
--- a/packages/components/src/Badge.js
+++ b/packages/components/src/Badge.js
@@ -1,13 +1,19 @@
+/** @jsx jsx */
+import { jsx } from '@theme-ui/core'
 import React from 'react'
 import Box from './Box'
+import { useVariant } from './util'
 
-export const Badge = React.forwardRef(function Badge(props, ref) {
+export const Badge = React.forwardRef(function Badge(
+  { variant, ...props },
+  ref
+) {
+  const variation = useVariant('badges', variant)
   return (
     <Box
       ref={ref}
       {...props}
-      __themeKey="badges"
-      __css={{
+      sx={{
         display: 'inline-block',
         verticalAlign: 'baseline',
         fontSize: 0,
@@ -17,6 +23,7 @@ export const Badge = React.forwardRef(function Badge(props, ref) {
         borderRadius: 2,
         color: 'white',
         bg: 'primary',
+        ...variation,
       }}
     />
   )
--- a/packages/components/src/Badge.js
+++ b/packages/components/src/Badge.js
@@ -1,21 +1,23 @@
 import React from 'react'
 import Box from './Box'
 
-export const Badge = React.forwardRef((props, ref) => (
-  <Box
-    ref={ref}
-    {...props}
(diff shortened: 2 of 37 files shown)


jxnblk:
Thanks! There's some interesting ideas here -- I need to add the v1 roadmap issue I've been drafting, but some of this could potentially fit in line with some other parts of that

packages/components/src/AspectRatio.js
```diff
@@ -18,7 +23,7 @@ export const AspectRatio = React.forwardRef(
       />
       <Box
         {...props}
-        __css={{
+        sx={{
```
jxnblk, line 26: I'd be considering changing the private API a bit here, but does this mean that `props.sx` is no longer forwarded to the component here? This looks like it'd break in MDX usage

packages/components/src/Badge.js
```diff
@@ -1,21 +1,30 @@
...
-      color: 'white',
-      bg: 'primary',
-    }}
-  />
-))
+export const Badge = React.forwardRef(function Badge(
+  { variant, ...props },
+  ref
+) {
+  const variation = useVariant('badges', variant)
```
jxnblk, line 11: This word/meaning bothers me, sorry...

```suggestion
  const variantStyle = useVariant('badges', variant)
```

dburles: all good!

packages/components/src/AspectRatio.js
dburles: Yes you're right, sorry had overlooked that, I've re-added the `__css` API.

dburles:
Here's an example my latest thoughts on a unified method to this whole approach:

```js
const modifiers = {
  variant: {
    primary: {
      title: { ... },
      subtitle: { ... },
    }
  },
  size: {
    small: {
      title: { fontSize: 3 },
      subtitle: { fontSize: 0 },
    },
    large: {
      title: { fontSize: 5 },
      subtitle: { fontSize: 2 },
    },
  },
  // etc
};

const Component = ({ size = 'small', modifiers: customModifiers }) => {
  const modifierStyle = useModifiers(
    { variant, size },
    modifiers,
    customModifiers // Optional secondary modifiers object that will merge with `modifiers`.
  );

  return (
    <div>
      <div sx={[{ fontFamily: 'heading' }, modifierStyle.title]}>{title}</div>
      <div sx={modifierStyle.subtitle}>{subtitle}</div>
    </div>
  );
};
```

General ideas:

Much of these changes have come from ideas presented in #586. 
This introduces new API with the `useModifiers` hook, but also gives a prescribed approach to building library components. The core component library should also use this same API.

- One universal term for prop based variations, I'm using 'modifiers' here, it fits nicely when using the `variant` term as the main type of modifier, `type` would work too, but it's overloaded by the HTML `type` attribute
- The theme does not need to be overloaded with `modifiers` objects, I'm just defining them alongside the component as it's unlikely they'll be shared with other components, but if they need to be, you can also put them on the theme
- It's possible to easily extend components defined this way by passing in a `modifiers` prop with the parts that you wish to alter, e.g. `<Component modifiers={{ size: { small: { title: fontSize: 2 } } }} />`
- The `sx` prop will need to [support array syntax](https://github.com/system-ui/theme-ui/pull/704) to automatically deep merge styles
- I don't think MDX styles will need a `variant` prop

I'd really like to get some discussion going to fine tune it. It's been a great API to work with so far in my own experience. I don't know if I am stepping outside of the core ideas behind theme-ui, but this solves a bunch of problems inherent with the current approach.

LekoArts:
This is purely a note from a conversation I had in private with @jxnblk for documentation sake:

Coming from https://github.com/system-ui/theme-ui/issues/990 I tested out the changes to the `Box` component as the `shouldForwardProp` in https://github.com/system-ui/theme-ui/blob/master/packages/components/src/Box.js#L18 isn't passing e.g. `activeClassName`. Changing the behavior of that component would fix this.

jxnblk:
Hey @dburles sorry for the lack of activity/discussion here, but really appreciate the thought and effort that went into this! I think to help move our TS conversion along, we'd like to ditch the use of the `styled` API, which you've done here. In an effort to try to unblock some of that, I'd like to branch off from where you've started here and get a new branch in shape to go in, but just wanted to check to see if that sounds okay to you. I think we could continue some of the other parts of the discussion here around API changes in other issues or a new PR, if that makes sense

dburles:
@jxnblk No problem. There is definitely some more discussion needed with these changes.

There are a bunch of scattered conversations on the variant API in general, what are your thoughts on creating a new banner issue to bring the discussion together as it's a little scattered right now. #800 is the main issue for this PR, there's also PR's #294 and #1017.

Just to confirm, would you like me to create a new PR from these changes, but without the variant alterations? Let me know if there's anything I've missed.

I'm not sure where that leaves this PR for the future, but I can move some of my comments from here over to any new discussion on the variant API changes in general.

jxnblk:
> what are your thoughts on creating a new banner issue to bring the discussion together as it's a little scattered right now. 

Yeah, a new issue or a few new issues might make sense. I think it should be generally inline with #832 or a proposal for a new API that can be evaluated on its own. e.g. `useVariant` is an interesting idea, but some of the other ideas don't require that or can be decoupled 

> Just to confirm, would you like me to create a new PR from these changes, but without the variant alterations?

No, no, no. Sorry if I wasn't clear. I can create a new branch from this PR, descope a couple things and tidy it up for a new PR which would supersede this one, but keeping your commit history -- just wanted to make sure you were cool with that.

> I'm not sure where that leaves this PR for the future, but I can move some of my comments from here over to any new discussion on the variant API changes in general.

Starting from from a new issue is fine IMO, linking to prior art or past issues and PRs where needed.

dburles:
Sounds good to me

lachlanjc:
Closing because we’re not planning on implementing this, & https://github.com/dburles/mystical is a successor if you’re looking for this functionality :) Thank you so much for the thoughtful PR though—we really appreciate it.

PR #833, opened 2020-04-08
feat(docs): dynamic favicon

jletey:
The favicon of the site when viewed in a dark mode browser is hard to see:

Screenshot 2020-04-08 at 10 15 46

@jxnblk My fix is currently not working, which is weird because it has worked for me in the [past](https://github.com/learn-anything/learn-anything/blob/master/app/src/pages/_document.tsx#L9-L25).


packages/docs/src/components/head.js
```diff
@@ -19,14 +19,30 @@ export default props => {
...
-      <link rel="icon" type="image/png" href="/icon.png" />
+      <link
+        rel="icon"
+        media="(prefers-color-scheme:dark)"
+        href="/icon-dark.png"
+        type="image/png"
+      />
+      <link
+        rel="icon"
+        media="(prefers-color-scheme:light)"
```
lachlanjc, line 30: Many operating systems don’t support this media query yet, so maybe cut this line so it’s the default?

jletey: @lachlanjc That is the whole point of this PR, to support `prefers-color-scheme` - I wonder if there is another way of going about doing this 🤔

jxnblk: I think @lachlanjc is suggesting moving the default (e.g. light mode) `<link>` above the other one and only using the `media` attribute on the dark mode one – I would assume that would work, but have not tried it out myself

lachlanjc: Yep, exactly what @jxnblk said 👍

jletey:
I have made the suggested changes, but still does not work :thinking:

jxnblk:
It appears to be working in the deploy preview for latest MacOS & Chrome

jletey:
> It appears to be working in the deploy preview for latest MacOS & Chrome

Yep, for some reason my computer cached the favicon - but can now confirm it is working 👍

jxnblk:
Sorry for the delay! This looks great, thanks for working on this!

PR #836, opened 2020-04-08
Fix unlisted peer dependencies

sjmeverett:
Some of the peer dependencies are not listed (#725), which causes a problem for Yarn Berry. (It's also technically incorrect, either way.)

An existing PR (#726) attempts to fix this, but doesn't quite get it right.

The correct solution is to keep passing peer dependencies upstream. If A relies on B, and B expects C to be installed as a peer, then A must either install it, or also expect it to be installed as a peer.

Thanks for your time!


quinnturner:
Quite some time has passed since this PR was created. To confirm that this is still an issue, here's a current stack trace:

```
 ERROR #98123  WEBPACK

Generating SSR bundle failed

@theme-ui/mdx tried to access @theme-ui/css (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.

Required package: @theme-ui/css (via "@theme-ui/css")
Required by: @theme-ui/mdx@virtual:7c04d473ae13266e8b953c54c0ad898aa2f80ae2cf2e1f73a9b1285e502045234560bd1756785b7d058c71f247b08e5c45d187da687a952e97aba305e3c2d477#npm:0.3.0 (via /Users/quinn/Dev/Tirasius/Phoenix/.yarn/$$virtual/@theme-ui-mdx-virtual-d9f3cb1045/0/cache/@theme-ui-mdx-npm-0.3.0-faddcad8d2-da3f668691.zip/node_modules/@theme-ui/mdx/dist/index.esm.js)
Ancestor breaking the chain: @theme-ui/theme-provider@npm:0.3.1



File: ../../.yarn/$$virtual/@theme-ui-mdx-virtual-d9f3cb1045/0/cache/@theme-ui-mdx-npm-0.3.0-faddcad8d2-da3f668691.zip/node_modules/@theme-ui/mdx/dist/index.esm.js
```

hasparus:
Hey! Thanks for the PR.

hasparus:
@quinnturner, thanks for bumping this.
I'll try to release it tomorrow in both 0.3 and 0.4-rc versions.

PR #837, opened 2020-04-09
Fix global style docs code typo

gosseti:
A small typo in the docs where a raw `*` is an object key name for the example code. Have wrapped it with quotes.


jxnblk:
Good catch!

PR #850, opened 2020-04-16
feat(@theme-ui/components) Add a Stack component based on Flex layout

### LINKED ISSUE #556: Add gap prop to Flex component

janhoogeveen:
This PR is an attempt at solving https://github.com/system-ui/theme-ui/issues/556

image

- [X] Working component
- [X] Finish documentation
- [X] Add tests

**Prior art**
I've taken inspiration from https://github.com/siddharthkp/react-ui/blob/master/packages/react-ui/src/components/stack/index.js

**Technical approach**
This solution uses Flexbox to create stack layouts. I was afraid that by using css selectors you wouldn't be able to override margins anymore, but turns out you can quite easily work your way around it.

For example, you can use a negative value in your child component to negate the gap property:

image

**Tradeoffs**
Flex components are more complex than a simple grid component. 

**Alternative approach**
This is an alternative, more powerful version of https://github.com/system-ui/theme-ui/pull/809

### DIFF
--- a/packages/components/src/Stack.js
+++ b/packages/components/src/Stack.js
@@ -1,24 +1,61 @@
 import React from 'react'
 import { Box } from './Box'
 
+export const merge = (...objs) => {
+  return objs.reduce(function(merged, currentValue) {
+    return mergeTwo(merged, currentValue)
+  }, {})
+}
+
+const createGap = (direction, gap) => {
+  if (direction === 'vertical') {
+    return {
+      marginBottom: gap,
+      marginRight: 0,
+    }
+  }
+
+  return {
+    marginBottom: 0,
+    marginRight: gap,
+  }
+}
+
 export const Stack = React.forwardRef(
-  ({ gap = 2, children, ...props }, ref) => {
-    /** Make a copy of the children so we can pass the styles as props instead of using brittle CSS selectors */
-    const elements = React.Children.map(children, (element, index) => {
-      if (index === 0) {
-        return element
-      }
+  (
+    {
+      gap = 2,
+      direction = 'vertical',
+      inline = false,
+      justify = 'normal',
+      align = 'normal',
+      children,
+      ...props
+    },
+    ref
+  ) => {
+    const styles = {
+      display: inline ? 'inline-flex' : 'flex',
+      width: '100%',
+      justifyContent: justify,
+      alignItems: align,
+    }
 
-      const newProps = {
-        marginTop: gap,
-      }
+    if (Array.isArray(direction)) {
+      styles.flexDirection = direction.map(d =>
+        d === 'vertical' ? 'column' : 'row'
+      )
+      styles['> *:not(:last-child)'] = direction.map(d => createGap(d, gap))
+    } else {
+      styles.flexDirection = direction === 'vertical' ? 'column' : 'row'
+      styles['> *:not(:last-child)'] = createGap(direction, gap)
+    }
 
-      return React.cloneElement(element, newProps)
-    })
+    console.log(styles)
 
     return (
-      <Box ref={ref} {...props} __themeKey="stacks">
-        {elements}
+      <Box ref={ref} {...props} __themeKey="stacks" __css={styles}>
+        {children}
       </Box>
     )
   }
--- a/packages/components/src/Stack.js
+++ b/packages/components/src/Stack.js
@@ -51,8 +51,6 @@ export const Stack = React.forwardRef(
       styles['> *:not(:last-child)'] = createSpace(direction, space)
     }
 
-    console.log(styles)
-
     return (
       <Box ref={ref} {...props} __themeKey="stacks" __css={styles}>
         {children}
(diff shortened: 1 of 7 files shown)


janhoogeveen:
Documentation is done, the component seems to work as expected and I've added tests.

All yours @jxnblk!

jxnblk:
@janhoogeveen sweet! thanks for working on this! I'll try to give this a review over the next few days

packages/components/src/Stack.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { Box } from './Box'
+
+export const merge = (...objs) => {
+  return objs.reduce(function(merged, currentValue) {
+    return mergeTwo(merged, currentValue)
```
jparklev, line 6: possibly jumping in w/o enough context, but where is `mergeTwo` coming from in this case?

packages/components/src/Stack.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    if (Array.isArray(direction)) {
+      styles.flexDirection = direction.map(d =>
+        d === 'vertical' ? 'column' : 'row'
+      )
+      styles['> *:not(:last-child)'] = direction.map(d => createSpace(d, space))
+    } else {
+      styles.flexDirection = direction === 'vertical' ? 'column' : 'row'
+      styles['> *:not(:last-child)'] = createSpace(direction, space)
```
raunofreiberg, line 51: I think it would be better to use the lobotomized owl selector (`> * + *`) instead here so it would not override any margin defined on the children.

[Here's an example.](https://codesandbox.io/s/infallible-proskuriakova-5zwzg)

siddharthkp:
whoops, saw my repo in the description as inspiration

please don't take the stack implementation from react-ui, it's doesn't follow the spec proposal (yet)

further reading if you're curious: https://sid.st/unpolished/flex-gap-polyfill/

packages/components/src/Stack.js
joe-bell: I disagree, I think adding any complex selectors (either this current implementation or the lobotomized owl) here is dangerous to be honest.

A lot of Theme UI is based on the `Box` primitive and anything like this is going to lead to specificty issues

joe-bell:
Although I really appreciate the work put into this, using `> *` selectors feels out of place with Theme UI for me; it could cause specificity issues and may also [impact SSR too](https://github.com/emotion-js/emotion/issues/1178).

**Side note:** If anyone's coming here looking for a `Stack` for Theme UI in the meantime, [I built **`raam`** for this exact purpose](https://raam.joebell.co.uk/).

packages/components/src/Stack.js
TheMightyPenguin: Something simimlar to what Braid does could be done, they add a bottom spacing to each item (using padding), and add a negative margin using the same spacing to the container.

Reference:
* https://github.com/seek-oss/braid-design-system/blob/master/lib/components/Stack/Stack.tsx#L116
* https://github.com/seek-oss/braid-design-system/blob/master/lib/components/Stack/Stack.tsx#L31

And you can inspect an example on [their playroom](https://seek-oss.github.io/braid-design-system/playroom/#?code=N4Igxg9gJgpiBcIA8BlALgQzAawAQGcAHLGAXgB0R8BbDAGzsoD5yA7XXJABTpIAsIdWACdcfGAEsA5nzSlgAFgAMAX1wB6Fu049%2BgkWMky5i1Rq0duvMDAFCYo8dNnzlazWyTr0WbFpAqQA)

joe-bell: This is the approach I'm using in [`raam`](https://raam.joebell.co.uk/) too and works rather nicely (although it needs a tidy up)

PR #865, opened 2020-04-23
Allow negative theme lookup of space objects

kenny-f:
Currently, if you set your `space` to be an object in theme instead of an array then using for example `-xl` does not work.

Example:
```
// theme.js
const theme = { space : { s: '8px', m: '16px' } } 
```
```
// some-component.js
<div
  sx={{
    margin: '-m'
  }}
/>
```

The `-m` does not get replaced with `-16px`.

What we currently have to do is:
```
// some-component.js
<div
  sx={{
    margin: (theme) => `-${theme.space.m}`
  }}
/>
```
which is not as nice.

This pr will see if the value starts with a `-` and return the theme lookup value and add a `-`.

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -129,6 +129,11 @@ type Scales = typeof scales
 
 const positiveOrNegative = (scale: object, value: string | number) => {
   if (typeof value !== 'number' || value >= 0) {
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks! This seems like a helpful addition to the API :)

PR #866, opened 2020-04-24
Add `opacities` scale to CSS function

### LINKED ISSUE #863: Opacities Scale
In [Modulz](https://modulz.app), we're exposing Opacity Tokens, however:
- `theme-ui/css` and `styled-system/css` functions don't include an `opacities` scale in its scales mapping
- System UI Spec doesn't specify one (https://github.com/system-ui/theme-specification/issues/12)

I'd like to be able to use the CSS function with an `opacities` scale:
```js
css({
  opacity: 1 // should return 50%
})({
  theme: {
    opacities: ["0%", "50%", "100%"]
  }
})
```

The CSS function defaults to mapping to a scale named after the property. It's probably not intentional, @jxnblk would appreciate some context here.

Because of that, if we have a scale called `opacity`, it works:

```js
css({
  opacity: 1 // should return 50%
})({
  theme: {
    opacity: ["0%", "50%", "100%"]
  }
})
```

**Additional context**
Here's a [Codesandbox Demo](https://codesandbox.io/s/css-function-opacities-scale-42g9w?file=/src/App.js) with my expectation and workaround mentioned above

peduarte:
- Added `opacity` — `opacities` mapping in CSS function
- Updated Theme typing definition to accept an `opacities` scale
- Updated tests

Closes #863

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -56,6 +56,7 @@ export const scales = {
   color: 'colors',
   backgroundColor: 'colors',
   borderColor: 'colors',
+  opacity: 'opacities',
   margin: 'space',
   marginTop: 'space',
   marginRight: 'space',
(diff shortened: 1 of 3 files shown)


jxnblk:
Awesome, thanks!

PR #870, opened 2020-04-26
Add default theme key for Text

joe-bell:
Adds a default theme key of `body` to the `Text` component

**Why should `Text` have a theme key?**
  - We can keep `Text` agnostic from the body font-size (allowing for rem/em customisation where necessary)
  - Right now `Text` fully inherits from the `root`, this particularly becomes a problem in some problem when [transitioning properties](https://bugs.webkit.org/show_bug.cgi?id=46041). 
  - Considering everything else in Theme UI is so component focussed, I feel like it would make sense to avoid this `inherit`ance. Besides, we can also allow for other variants to extend upon `body` if necessary.

**Why `body`?**
  - I think this keeps it inline with `heading` and feels common; `body` is also used in other places 

**Example**

I've used this to avoid the aforementioned '`inherit` flicker' in my [personal site](https://github.com/joe-bell/joebell.co.uk/blob/master/src/theme-ui/themes/base.js#L89), however I still think it would be beneficial for other reasons I've outlined.

### DIFF
--- a/packages/components/src/Text.js
+++ b/packages/components/src/Text.js
@@ -2,5 +2,5 @@ import React from 'react'
 import Box from './Box'
 
 export const Text = React.forwardRef((props, ref) => (
-  <Box ref={ref} {...props} __themeKey="text" />
+  <Box ref={ref} variant="default" {...props} __themeKey="text" />
 ))
--- a/packages/components/test/__snapshots__/index.js.snap
+++ b/packages/components/test/__snapshots__/index.js.snap
@@ -492,12 +492,6 @@ exports[`Embed renders 1`] = `
 `;
 
 exports[`Field renders 1`] = `
-.emotion-2 {
-  box-sizing: border-box;
-  margin: 0;
-  min-width: 0;
-}
-
 .emotion-0 {
   box-sizing: border-box;
   margin: 0;
@@ -525,6 +519,12 @@ exports[`Field renders 1`] = `
   border-radius: 4px;
   color: inherit;
   background-color: transparent;
+}
+
+.emotion-2 {
+  box-sizing: border-box;
+  margin: 0;
+  min-width: 0;
 }
 
 <div
@@ -1313,6 +1313,7 @@ exports[`Text renders 1`] = `
   box-sizing: border-box;
   margin: 0;
   min-width: 0;
+  font-size: 20px;
 }
 
 <div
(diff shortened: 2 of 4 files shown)


lachlanjc:
YES! Making a marketing website with Theme UI right now & super want this — I want bigger text by default without changing the body styles. Huge +1 to this.

jxnblk:
Thanks! I think this is a logical addition. I'd only ask about the name: should the variant be something *other* than `body` since that is used as the `<body>` font elsewhere? Perhaps `theme.text.default` or something else?

joe-bell:
@jxnblk no worries! When it comes to personal work I always opt for `default` but it felt a little out of place in Theme UI for me; where keys tend to be semantic. I felt `body` made more sense here because `heading` is used as a key too, and any group of text could be a 'body of text'. Happy to change it if you don't feel it's right!

What about `content` or `paragraph` as some alternatives?

joe-bell:
@jxnblk on second thoughts I'm on board for `default` too. If we opt for `default` it would be great if other components followed the same key format in a later version of Theme UI (e.g. `button.default`)

Happy with either approach here

alexluong:
Literally about to open an issue with the same suggestion and found this instead.

I'd throw another name in the hat: `theme.text.text`. That would be in line with `theme.text.heading` as default for `<Heading />` component.

joe-bell:
I think to be honest I'm more leaning towards `.default` now, but thanks for the suggestion @alexluong!

joe-bell:
@jxnblk switched the key to `default`, anything else to do on my end?

jxnblk:
This looks great, thanks! This will probably go out with the next canary release -- I might need to roll back the public docs to the latest 0.3 release, but I'll figure out how to handle that

dabrady:
@yuyokk Same! [The docs](https://theme-ui.com/components/text) explicitly say this behavior already exists:
> Text style variants can be defined in the theme.text object. The Text component uses theme.text.default as its default variant style.

I don't know how the docs are versioned, or what processes are in place for releasing them, but I'm sure we're not the only ones puzzled by it. Is this what you were talking about, @jxnblk, when you said:
> I might need to roll back the public docs to the latest 0.3 release, but I'll figure out how to handle that

akonwi:
@jxnblk is there any update on when this will be released?

hasparus:
@akonwi We'll get to it this week / tomorrow. Gotta do some more tests of the entire changeset.

PR #878, opened 2020-04-28
Let gatsby-plugin-theme-ui take base preset options

laurieontech:
This code allows users to pass in their base themes through the plugin options. They can do it by path or by requiring the package directly. Any local shadowing is respected and merged with the base preset.

### DIFF
--- a/packages/gatsby-plugin-theme-ui/README.md
+++ b/packages/gatsby-plugin-theme-ui/README.md
@@ -34,7 +34,7 @@ module.exports = {
   plugins: [
     { resolve: 'gatsby-plugin-theme-ui',
       options: {
-        themePreset: '@theme-ui/preset-funk'
+        preset: '@theme-ui/preset-funk'
         // or require('@theme-ui/preset-funk')
       }
     }],
--- a/packages/gatsby-plugin-theme-ui/README.md
+++ b/packages/gatsby-plugin-theme-ui/README.md
@@ -20,7 +20,7 @@ prevent a flash of unstyled colors when using color modes.
 
 | Key                      | Default value    | Description                                                                      |
 | ------------------------ | ---------------- | -------------------------------------------------------------------------------- |
-| `themePreset`               | `null`            | This can be a JSON theme object or a string package name. Make sure the package you're requiring is installed in your dependencies.               |
+| `preset`               | `null`            | This can be a JSON theme object or a string package name. Make sure the package you're requiring is installed in your dependencies.               |
 
 > Note that this plugin assumes the theme object is exported as `default`.
 
--- a/packages/gatsby-plugin-theme-ui/README.md
+++ b/packages/gatsby-plugin-theme-ui/README.md
@@ -22,9 +22,9 @@ prevent a flash of unstyled colors when using color modes.
 | ------------------------ | ---------------- | -------------------------------------------------------------------------------- |
 | `themeModule`               | `null`            | JSON theme object, use the `require` syntax to include it in options. Make sure the package you're requiring is installed in your dependencies.               |
 | `themeModulePath`            | `null`  | A string package name that the plugin will require for you. Make sure the package you're requiring is installed in your dependencies.                                                             |
-| `moduleExportName`              | `default` | The name of the export from the theme module, applies to `themeModule` or `themeModulePath` resolution depending which one you're using    |
 
-> Note that if your theme is exported at the top level, the `moduleExportName` of `default` is bypassed. See [theme-ui/preset-deep](https://github.com/system-ui/theme-ui/blob/master/packages/preset-deep/src/index.ts).
+
+> Note that this plugin assumes the theme object is exported as `default`.
 
 The theme module you include in options is considered your base theme. Any further customization and shadowing will be merged with it. 
 
--- a/packages/gatsby-plugin-theme-ui/README.md
+++ b/packages/gatsby-plugin-theme-ui/README.md
@@ -16,6 +16,33 @@ module.exports = {
 In addition to providing context, this plugin will also
 prevent a flash of unstyled colors when using color modes.
 
+## Options
+
+| Key                      | Default value    | Description                                                                      |
+| ------------------------ | ---------------- | -------------------------------------------------------------------------------- |
+| `themeModule`               | `null`            | JSON theme object, use the `require` syntax to include it in options. Make sure the package you're requiring is installed in your dependencies.               |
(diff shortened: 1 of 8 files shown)


laurieontech:
Making this a draft in order to execute the following:
- Error/command for missing dependency
- Remove moduleExportName which is only possible if we align on default for all existing presets (will create PRs)

laurieontech:
This depends on #882 for some themes to work cleanly.

packages/gatsby-plugin-theme-ui/README.md
```diff
@@ -16,6 +16,31 @@ module.exports = {
 In addition to providing context, this plugin will also
 prevent a flash of unstyled colors when using color modes.
 
+## Options
+
+| Key                      | Default value    | Description                                                                      |
+| ------------------------ | ---------------- | -------------------------------------------------------------------------------- |
+| `themePreset`               | `null`            | This can be a JSON theme object or a string package name. Make sure the package you're requiring is installed in your dependencies.               |
```
jxnblk, line 23: NB (and not to bikeshed), but what are your thoughts on making this simply `preset`? Generally trying to avoid overuse of the word *theme* if possible :)

packages/gatsby-plugin-theme-ui/README.md
```diff
@@ -16,6 +16,31 @@ module.exports = {
...
+
+### Using options
+
+```js
+// gatsby-config.js
+module.exports = {
+  plugins: [
+    { resolve: 'gatsby-plugin-theme-ui',
+      options: {
+        themeModulePath: '@theme-ui/preset-funk'
```
jxnblk, line 37: This would change according to whatever the options above are, right?

packages/gatsby-plugin-theme-ui/gatsby-node.js
```diff
@@ -0,0 +1,58 @@
+exports.onPreInit = (__, options) => {
+  let { themePreset } = options
+
+  if (typeof themePreset === 'string') {
+    try {
+      options.themePreset = require(themePreset)
```
jxnblk, line 6: NB: Is this mutating the `options` object? Might be nice to define a new variable in this module's scope

packages/gatsby-plugin-theme-ui/gatsby-node.js
```diff
@@ -0,0 +1,58 @@
...
+
+function generateInstallInstructions() {
+  const { getConfigStore } = require(`gatsby-core-utils`)
+
+  const packageMangerConfigKey = `cli.packageManager`
+  const PACKAGE_MANGER = getConfigStore().get(packageMangerConfigKey) || `yarn`
+
+  const installKeyWord = PACKAGE_MANGER === `yarn` ? 'add' : 'install'
+
+  return `${PACKAGE_MANGER} ${installKeyWord}`
```
jxnblk, line 23: This smells like something that could be useful as a utility for other plugins

packages/gatsby-plugin-theme-ui/README.md
laurieontech: Bikeshedding is fair! I actually debated this quite a bit. My fear is twofold. One, if you pass it from a site using `blog-theme` for example you lose context. And two is that `prismPreset` is also a preset and therefore a bit confusing?

packages/gatsby-plugin-theme-ui/README.md
laurieontech: Ya, this needs to be updated with the refactor.

packages/gatsby-plugin-theme-ui/gatsby-node.js
laurieontech: If this doesn't mutate then we can't re-use themePreset to work for both JSON and a string because it won't resolve to JSON at this stage, making it incompatible with the schema customization.

packages/gatsby-plugin-theme-ui/gatsby-node.js
laurieontech: Agreed. Though we might be able to remove this now because we've updated the gatsby core webpack error. It's not quite as specific as this instruction set, but it does recommend the user install the dependency that's missing. Do we think that's sufficient? Or do we want both?

jxnblk: Ah yeah... I think relying on Gatsby's error messaging here rather than a custom one here might make more sense TBH

packages/gatsby-plugin-theme-ui/README.md
jxnblk: I think for Theme UI, the `theme` object is the main one and Prism is a secondary/optional library -- for the blog theme, maybe the theme's options are nested as `options.themeUIOptions`, which seems like a convention used some of the time, *or* we could map something like `options.themeUIPreset` to this `options.preset` in the blog theme

packages/gatsby-plugin-theme-ui/gatsby-node.js
laurieontech: That's fair. Will update.

packages/gatsby-plugin-theme-ui/README.md
laurieontech: That's fair. Updated.

packages/gatsby-plugin-theme-ui/gatsby-node.js
laurieontech: My only question is if we want this to blow up the site? We could conceivably attempt the require and put a warning message instead?

packages/gatsby-plugin-theme-ui/README.md
```diff
@@ -16,6 +16,31 @@ module.exports = {
...
+
+### Using options
+
+```js
+// gatsby-config.js
+module.exports = {
+  plugins: [
+    { resolve: 'gatsby-plugin-theme-ui',
+      options: {
+        themePreset: '@theme-ui/preset-funk'
```
jxnblk, line 37: I think this should be updated??
```suggestion
        preset: '@theme-ui/preset-funk'
```

packages/gatsby-plugin-theme-ui/gatsby-node.js
```diff
@@ -0,0 +1,44 @@
+exports.onPreInit = (__, options) => {
+  let { preset } = options
+
+  if (typeof preset === 'string') {
+    try {
+      options.preset = require(preset)
+    } catch {
+      reporter.warn(
+        `It appears your theme dependency is not installed. Only local styles will appear.`
```
jxnblk, line 9: Nice!

PR #879, opened 2020-04-29
Add Prism preset to options

laurieontech:
This allows users to pass a prismPreset option into gatsby-plugin-theme-ui and use it as the base for their prism configuration. Any local shadowing supersedes this base.

At the moment, this PR adds the prism preset package as a dependency and allows you to pass just the string, e.g. `night-owl` to the plugin. Alternatively, we can require a user to pass `@theme-ui/prism/presets/night-owl.json` and have them include the dependency locally (whether in a theme or the parent site) to prevent us from adding this dependency to the plugin.

Note that this is branched off of the PR code for #878.


laurieontech:
Note that I don't love having the dictionary sit in gatsby-node.js but we're limited with import support in node unless we want to add esm, which feels unecessary.

laurieontech:
Looking into adding this error behavior directly to gatsby core.

packages/gatsby-plugin-theme-ui/utils/preset-dictionary.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
```
laurieontech, line 1: Note sure what is happening with the prettier hook to change the format of the keys.

PR #880, opened 2020-04-29
Typography options

laurieontech:
Add the ability to pass a typography js theme into the plugin options.

This uses the approach of requiring the user to include the dependency themselves but adds `@theme-ui/typography` to this package so it can leverage the `toTheme` function. We may want to discuss this further.

Additionally, the if statements for various options are getting a bit messy, so once we review the various PRs and decide on a consistent approach I can clean that up.

Builds off #878


laurieontech:
After talking to @jxnblk we're going to shelve this option temporarily and consider it down the road.

PR #882, opened 2020-04-30
Export presets consistently by name and default

laurieontech:
We're inconsistent with how we define/export theme presets. While some still export subsets of settings, they should all export a full theme using the name of the preset as well as a default export.


jxnblk (approved):
:shipit:

johno (approved):
✨

PR #886, opened 2020-05-02
Fix link to @theme-ui/presets from gatsby-plugin-theme-ui

josephrace:
Fixing link to @theme-ui/presets from the gatsby-plugin-theme-ui README.md file


jxnblk:
Thanks!

PR #891, opened 2020-05-05
add Label to checkbox docs

### LINKED ISSUE #884: Can't use Checkbox without Label
**Describe the bug**
The `Checkbox` component doesn't work unless it's wrapped in a `label`.

**To Reproduce**
Render a `<Checkbox />` element, and try to click it. Nothing happens. Wrap it in a `label` tag; `<label><Checkbox /></label>` and you can check/uncheck it.

**Expected behavior**

The `Checkbox` component works without a `label` tag, or this behaviour is documented.

**Screenshots**

**Additional context**

[Codesandbox demo](https://codesandbox.io/s/magical-cherry-y2pjl?file=/src/App.js)

robphoenix:
closes #884

### DIFF
--- a/packages/docs/src/pages/components/checkbox.mdx
+++ b/packages/docs/src/pages/components/checkbox.mdx
@@ -11,9 +11,11 @@ import { Checkbox } from 'theme-ui'
 ```
 
 ```js live=true
-<Checkbox
-  defaultChecked={true}
-/>
+<Label>
+  <Checkbox
+   defaultChecked={true}
+  />
+</Label>
 ```
 
 ## Variants


lachlanjc:
Awesome. Maybe add some text so it’s clear you always should? “Checkbox”?

robphoenix:
@lachlanjc yeah, that would be good. Would be worth adding it to the `Radio` component as well, as I believe the same goes.

Do you know why these need to be wrapped in a `label` tag? Would be good to explain why, rather than just saying *You must do this*. 😄

jxnblk:
Awesome thanks! All form elements need `<label>` elements, otherwise they don't work. Rather than build a leaky abstraction here, the components are generally 1:1 with HTML elements

gormonn:
> Awesome thanks! All form elements need `<label>` elements, otherwise they don't work. Rather than build a leaky abstraction here, the components are generally 1:1 with HTML elements

It doesn't explain anything.
Who need more details [read this](https://dequeuniversity.com/rules/axe/4.0/label)

Issue #825, opened 2020-04-06
(@theme-ui/color) exporting getColor


hasparus:
**Is your feature request related to a problem? Please describe.**

I'd like to use _polished_ or other JS color utils to handle complex color transformations.

_example:_

```ts
flow(
  (t: Theme) => t.colors?.primary || "#eeffff",
  setSaturation(0.75),
  transparentize(0.92)
)
```

ATM the code above doesn't work, because `t.colors.primary` is `'var(--theme-ui-colors-primary, #e85554)'`, what's really useful for inspection and tweaks in browser dev tools, but it needs parsing to retrieve the hexadecimal number I'd like to operate on.

This parsing logic is currently defined in _@theme-ui/color_ in the function called `g`.

```ts
const g = (t: Theme, c: string) =>
  get(t, `colors.${c}`, c)
    .replace(/^var\(--(\w+)(.*?), /, "")
    .replace(/\)/, "");
```

_working version of the from the first snippet looks like this_

```ts
const parseColor = (cssVarColor: string) =>
  cssVarColor.replace(/^var\(--(\w+)(.*?), /, "").replace(/\)/, "");

flow(
  (t: Theme) => t.colors?.primary || "#eeffff",
  parseColor,
  setSaturation(0.75),
  transparentize(0.92)
);
```

**Describe the solution you'd like**

If `g` was exported as `getColor` I could change my code to

```ts
flow(
  (t: Theme) => getColor(t, "primary"),
  setSaturation(0.75),
  transparentize(0.92)
);
```

and avoid including the knowledge about that CSS variable is there in application code.

**Describe alternatives you've considered**

1. I can keep the `parseColor` function.

2. Alternatively, I can rewrite the code from my example to use functions exported from _@theme-ui/color_.

```ts
(t: Theme) => transparentize(saturation('primary', 0.75)(t), 0.92)(t)
```

I think exporting `g` as  `getColor` would make using other color libraries alongside _@theme-ui/color_ simpler.

jxnblk:
Thanks! I think exporting this would be fine, but also curious if there's a higher level API that might make sense as an addition to the `@theme-ui/color` package -- could you explain a little more about what you're doing in your first example?

jxnblk:
Thanks! It looked like something like composition was going on in your example. It'd be great if the Theme UI wrapper functions also did this and maybe provided a `compose` or similar function -- haven't looked at the code yet, but just thinking out loud...

danvim:
[...]

PR #894, opened 2020-05-07
Export getColor from theme-ui/colors

hasparus:
Closes https://github.com/system-ui/theme-ui/issues/825


jxnblk:
Awesome! Looks great, thanks!

PR #705, opened 2020-02-23
WIP: refactor(gatsby-plugin-theme-ui): add typescript

hasparus:
Part of #668.

Needs types for `theme-ui`.

---
Results of `npm pack`:
[gatsby-plugin-theme-ui-v0.3.0.zip](https://github.com/system-ui/theme-ui/files/4240444/gatsby-plugin-theme-ui-v0.3.0.zip)

PR #900, opened 2020-05-08
Convert editor to TypeScript

dcastil:
Partly addresses #668.

This is my first contribution to this project, so I added some TODOs and FIXMEs for things I couldn't help myself for now.


packages/editor/src/Theme/FontWeights.tsx
```diff
@@ -1,21 +1,22 @@
...
+import { jsx, useThemeUI, Theme } from 'theme-ui'
+import * as CSS from 'csstype'
 import Combobox from '../Combobox'
+import { EditorContext } from '../types'
 
 const weights = [100, 200, 300, 400, 500, 600, 700, 800, 900]
 
-export default props => {
-  const context = useThemeUI()
+const FontWeights = () => {
```
dcastil, line 9: Note: I'm changing all the `default export () => {}` to `const Foo = () => {}; default export Foo` due to https://github.com/gatsbyjs/gatsby/issues/23575.

packages/editor/src/Combobox.tsx
```diff
@@ -27,7 +27,7 @@ const keys = {
   13: 'return',
 }
 
-export default ({
+const Combobox = ({
```
dcastil, line 30: Note: I'm changing all the `default export () => {}` to `const Foo = () => {}; default export Foo` due to gatsbyjs/gatsby#23575.

dcastil:
@mxstbr Could you give me some initial feedback on the PR?

I converted all the source code to TypeScript so far (excluding tests), but there are still unresolved TypeScript errors. I marked all of them with a TODO or FIXME and they mostly deal with stuff which needs code changes outside of the package. Maybe you have more context on those.

packages/editor/src/ColorPicker.tsx
```diff
@@ -36,12 +37,26 @@ const Handle = props => (
   />
 )
 
+type InputProps = {
+  name?: string
```
dcastil, line 41: I should use the EditableInput prop types directly instead.

packages/editor/src/Sx/Space.tsx
```diff
@@ -67,64 +74,63 @@ export const Space = ({
...
       sx={{
         display: 'grid',
-        gridGap: 2,
+        gap: 2,
         gridTemplateColumns: 'repeat(3, 1fr)',
         alignItems: 'center',
       }}>
       <Field
+        // FIXME: Field type is comming from external package @types/theme-ui__components, not sure why type prop does not exist in type definition.
         type="number"
```
dcastil, line 91: I get following TypeScript error

```
Type '{ type: string; name: string; label: string; value: any; onChange: (e: any) => void; }' is not assignable to type 'IntrinsicAttributes & FieldOwnProps & Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement> | ... 19 more ... | PropsWithChildren<...>, "ref" | ... 256 more ... | "sx"> & { ...; }'.
  Property 'type' does not exist on type 'IntrinsicAttributes & FieldOwnProps & Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement> | ... 19 more ... | PropsWithChildren<...>, "ref" | ... 256 more ... | "sx"> & { ...; }'.ts(2322)
```

despite the definition in definitelyTyped being

```ts
export interface InputProps extends Assign<React.ComponentProps<'input'>, BoxOwnProps> {}

[…]

export type FieldProps<T extends React.ElementType> = FieldOwnProps &
    Omit<React.ComponentProps<T>, 'as' | keyof FieldOwnProps> & {
        /**
         * form control to render, default Input
         */
        as?: T;
    };

// `T` is far from unnecessary. We derive component props from it.
// tslint:disable-next-line no-unnecessary-generics
export function Field<T extends React.ElementType = React.ElementType<InputProps>>(props: FieldProps<T>): JSX.Element;
```

I don't understand why this is happening. When I add the prop `as={Input}`, the error goes away. We could use that as a workaround, but that's not particularly nice.

packages/editor/src/Sx/ThemeColorPicker.tsx
```diff
@@ -1,16 +1,28 @@
...
+  theme,
+  ...props
+}: ThemeColorPickerProps) => {
   const popover = usePopoverState()
   const context = useThemeUI()
   // todo: look into supporting v0.2 functionality
   // const { colors } = theme || context.theme || {}
   const _theme = theme || context.theme || {}
+  // FIXME: Where to add rawColors type?
   const colors = _theme.rawColors || _theme.colors || {}
```
dcastil, line 26: `Property 'rawColors' does not exist on type 'Theme'.ts(2339)`

I'm not sure where `rawColors` should get defined, a search through the codebase doesn't seem to bring any results. Probably on the `Theme` type directly, but not sure.

packages/editor/src/Theme/Colors.tsx
```diff
@@ -0,0 +1,66 @@
...
+  swatch: React.ReactNode
+  key: string
+  name: string
+}
+
+const Colors = (props: ColorsProps) => {
+  // TODO: I need to cast the context value because TypeScript assumes the default theme-ui ContextValue. Is there a better way to make TypeScript know of the EditorContext than casting the type?
+  const context = useThemeUI() as EditorContext
+  // FIXME: Where to add colorMode type?
+  const mode = context.colorMode
```
dcastil, line 21: I guess the type for `colorMode` will be added in scope of converting the `color-mode` package to TypeScript. Should I wait for it to get merged or try to work around it?

Same applies to `initialColorMode` type.

hasparus: That's correct.

Generally, if this is not a part of public API, there's nothing wrong with `as EditorContext` type assertion.
 
I'm not sure if `setTheme` is public. It seems it's not mentioned in the docs?
https://theme-ui.com/use-theme-ui

If we wanted to tell TypeScript that when we import `@theme-ui/editor` ContextValue is extended we could use module augmentation and declaration merging.

```ts
// in @theme-ui/editor
declare module 'theme-ui' {
  export interface ContextValue {
    setTheme: React.Dispatch<Theme>
  }
}
```

similarly to how we add `sx` prop to React components for use with the jsx pragma.

I am really not sure if `setTheme` is public though @jxnblk.

packages/editor/src/Sx/ThemeColorPicker.tsx
hasparus: @jxnblk What's `rawColors`?

packages/editor/src/Sx/Typography.tsx
```diff
@@ -36,12 +48,13 @@ export const SxTypography = ({
...
         sx={{
           display: 'grid',
-          gridGap: 2,
+          gap: 2,
           gridTemplateColumns: 'repeat(3, 1fr)',
         }}>
         <Field
           name={prefixName('fontSize')}
           label="Font Size"
+          // FIXME: Field type is comming from external package @types/theme-ui__components, not sure why value prop does not exist in type definition.
```
hasparus, line 57: It seems the `Field` doesn't default to `InputProps` properly.

packages/editor/src/ColorPicker.tsx
```diff
@@ -0,0 +1,216 @@
...
+          <Label>
+            Saturation
+            <Input
+              {...props}
+              value={round(props.hsl.s * 100)}
+              name="saturation"
+              label="s"
+              onChange={({ s }) => {
+                // FIXME: props.onChange() incorrectly expects param to be ColorState instead of { h, s, l }.
+                props.onChange({ ...props.hsl, s: s / 100 })
```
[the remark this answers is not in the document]
hasparus: > This component will be passed hex, rgb and hsl values as props for the current color. It is also provided an onChange prop that should be called to propagate a new color. Pass it a hex string, or an rgb or hsl object.

I think the docs actually describe what's on your snippet.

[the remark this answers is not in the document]
hasparus: https://github.com/casesandberg/react-color/blob/45baac6aa2d5d4f857b6ad578abca17008e3e1bd/src/helpers/color.js#L25

And the source code (as of commit you've sent) matches the type.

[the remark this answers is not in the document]
dcastil: Yes, the code is correct, but the type definition is incorrect. Per JavaScript code `data` would have following Type:

```ts
type Data = HSLColor | RGBColor | { hex: string }
```

But in DefinitelyTyped it is defined as

```ts
type Data = {
    hex: string;
    hsl: HSLColor;
    hsv: HSVColor;
    rgb: RGBColor;
    oldHue: number;
    source: string;
}
```

The code runs fine (it worked before the TypeScript conversion), but TypeScript won't let me compile and raises this error:

```
Argument of type '{ s: number; a?: number; h: number; l: number; source?: string; }' is not assignable to parameter of type 'ColorState'.
  Object literal may only specify known properties, and 's' does not exist in type 'ColorState'.ts(2345)
```

packages/editor/src/Theme/Colors.tsx
dcastil: Just curious: Does such an augmentation in any TypeScript file add the `setTheme` property to all `ContextValue`s in the codebase or is it scoped somehow?

packages/editor/src/Sx/Typography.tsx
dcastil: I was able to reproduce the problem with my own type (basically without `BoxOwnProps ` and `FieldOwnProps `), but couldn't figure out how to fix the issue. Is there a way to debug types easily?

I left a comment with some more info here: https://github.com/system-ui/theme-ui/pull/900#discussion_r422516796

packages/editor/src/ColorPicker.tsx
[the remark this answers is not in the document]
hasparus: Oh, I get it. It actually requires ColorState, when it builds it from `data`.

Well, don't worry about it.
I can help you out if you have some questions about contributing to DT,
but for the purpose of this PR let's just use assertion (`as any as ColorState`). 

I'm pretty confused by react-color's docs 🤷‍♂️
https://casesandberg.github.io/react-color/#api-onChange

packages/editor/src/Theme/Colors.tsx
hasparus: > Just curious: Does such an augmentation in any TypeScript file add the setTheme property to all ContextValues in the codebase or is it scoped somehow?

It's global. i.e. it adds `setTheme` as long as you have `@theme-ui/editor` installed.

It makes sense for `color-modes`, because there's usually ColorModeProvider at the root, but I'm not sure about the editor.

hasparus:
You can change `currentTarget` back to `target` if you replace `React.FormEvent` with `React.ChangeEvent`.

packages/editor/src/Sx/Typography.tsx
hasparus: Okay, I just changed Field typings locally to
```tsx
export function Field<T extends React.ElementType = React.ComponentType<InputProps>>(props: FieldProps<T>): JSX.Element;
```
and to
```tsx
export function Field<T extends React.ElementType = "input">(props: FieldProps<T>): JSX.Element;
```

both work.

You can also check it when you by the generic parameter to `Field`

```tsx
      <Field<"input">
        type="number"
      />
```

```tsx
      <Field<typeof Input>
        type="number"
      />
```

hasparus: The only thing it did is skipping the iteration over `keyof JSX.IntrinsicElements` in `ElementType`.

```tsx
    type ElementType<P = any> =
        {
            [K in keyof JSX.IntrinsicElements]: P extends JSX.IntrinsicElements[K] ? K : never
        }[keyof JSX.IntrinsicElements] |
        ComponentType<P>;
```

hasparus: Okay, so there are two things to do.
Firstly, we shouldn't use `@types/theme-ui__components`.
That's our bad twin from alternative dimension and he's out to get us.
Mea culpa :D

Things we have to do.
- [ ] Move the types from DT to this repo.
       Strong deja vu -- I think actually did it once. Is it in one of the open PRs?
- [ ] Do the change I mentioned in previous comment.
- [ ] Fix my sins and PR the change I mentioned in previous comment in DT for other people
  - [ ] Reflect that in .d.ts "test" in DT.

> Is there a way to debug types easily?

Well, they usually aren't that hard to debug.

In our case -- Unit tests.
- https://github.com/cartant/ts-snippet
- https://github.com/SamVerschueren/tsd
- https://github.com/ikatyang/dts-jest

dcastil: Cool, thanks for your help! 🙌 I think I even tried `<Field<'input'>>` but didn't wait long enough for TypeScript to recheck. 😅

> Strong deja vu -- I think actually did it once. Is it in one of the open PRs?

Yes, I found it in https://github.com/system-ui/theme-ui/pull/705/files#diff-c9edcada2823b2f3423667cdb4f5bf44R209. Would you like to change it in #705 or should I import the types into this one?

packages/editor/src/Sx/Space.tsx
dcastil: Fixed in 196bc704.

dcastil:
> You can change `currentTarget` back to `target` if you replace `React.FormEvent` with `React.ChangeEvent`.

Fixed in ecfe6bd8.

packages/editor/src/Sx/Typography.tsx
dcastil: >  Do the change I mentioned in previous comment.

I fixed the Field type issues in 196bc704.

packages/editor/src/ColorPicker.tsx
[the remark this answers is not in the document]
dcastil: Yeah, there are multiple components using onChange props and not all of them expect the same input. I fixed the issue in 9ba7a526.

packages/editor/src/Sx/Typography.tsx
hasparus: I suppose #705 won't be merged quickly tbh.
Dunno what approach will we take with gatsby plugins.

packages/editor/src/Sx/ThemeColorPicker.tsx
jxnblk: Can't remember if this was removed, but I think this *might* be a remnant from the previous version -- doing a code search looks like it might be some dead code that was missed...

dcastil: I’ll remove it in this PR then. 👍

packages/editor/test/Sx.tsx
```diff
@@ -44,243 +49,203 @@ describe('Sx.Typography', () => {
...
 describe('Sx.Margin', () => {
   test('edits sx.ml', async () => {
     const onChange = jest.fn()
-    const tree = render(
-      <Sx.Margin
-        theme={theme}
-        value={style}
-        onChange={onChange}
-      />
+    const tree = render(<Sx.Margin value={style} onChange={onChange} />)
```
dcastil, line 131: I had to remove the `theme` prop because the `Sx.Margin` component doesn't accept that in its prop type.

packages/editor/test/Sx.tsx
```diff
@@ -9,17 +9,20 @@ import {
...
-if (global.document) {
-  document.createRange = () => ({
-    setStart: () => {},
-    setEnd: () => {},
-    commonAncestorContainer: {
-      nodeName: 'BODY',
-      ownerDocument: document,
-    },
-  })
+if ((global as any).document) {
```
dcastil, line 16: I used `any` within tests as they are not within the production code. Hope that's okay.

packages/editor/src/Theme/Colors.tsx
dcastil: I added `any`s to color-mode-related things and a TODO to remove it when color-modes are converted to TypeScript. Let me know if I should do it differently.

packages/editor/src/Sx/Typography.tsx
dcastil: I didn't add the component types into my branch because I didn't need to change any of the types to get it to work. Let me know if I should still add them within this PR.

dcastil:
Btw. my editor (VS Code) is not showing the errors which are displayed here: https://github.com/system-ui/theme-ui/pull/900/checks?check_run_id=671810516

@hasparus Did you run into this issue before? Seems like VS Code is picking up types differently than Jest.

packages/editor/test/Sx.tsx
hasparus: yeah, don't worry about it

hasparus:
@dcastil, Jest is configured globally (it's like a separate project) for whole monorepo.
You can find its tsconfig in package.json `jest.globals["ts-jest"].tsConfig`.

This might be a bit confusing, I agree.
Since different packages in Theme UI have different levels of TypeScript adoption, local TS configs can differ. Jest doesn't know about them though, so all tests have the same TS options.

dcastil:
@hasparus Is anything missing here before it can get merged?

packages/editor/src/ColorPicker.tsx
```diff
@@ -0,0 +1,214 @@
...
+        },
+      },
+    }}
+  />
+)
+
+type CustomPickerForwardedProps = {
+  size?: number
+}
+type PickerProps = CustomPickerForwardedProps & CustomPickerProps
```
hasparus, line 91: Could we export `PickerProps` and turn it into an interface to make declaration merging possible in user's repository?

```suggestion
export interface PickerProps extends CustomPickerForwardedProps, CustomPickerProps {}
```

I agree that it looks worse, but it detects conflicts instead of resolving to `never` and it's possible to declaration merge it. https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces

packages/editor/src/ColorPicker.tsx
```diff
@@ -0,0 +1,214 @@
...
+              }}
+            />
+          </Label>
+        </div>
+      </div>
+    )
+  }
+)
+
+type ColorPickerProps = React.PropsWithoutRef<PickerProps> & {
```
hasparus, line 184: Similarly as above (https://github.com/system-ui/theme-ui/pull/900/files#r426841405).

It's still visible to the user, and one can do `React.ComponentProps<typeof ColorPicker>`, but importing `ColorPickerProps` is just a bit nicer.

packages/editor/src/Combobox.tsx
```diff
@@ -20,14 +20,27 @@ const Chevron = () => (
...
-export default ({
+type ComboboxOwnProps<T> = {
+  type?: string
+  name: string
+  label?: React.ReactNode
+  value: T
+  onChange: (val: string | T) => void
+  options?: T[]
+}
+type ComboboxProps<T> = ComboboxOwnProps<T> &
```
hasparus, line 38: Combobox is exported. Let's export the props too.

packages/editor/src/EditorProvider.ts
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { jsx, Context, useThemeUI, merge, Theme } from 'theme-ui'
+import { ThemeContext as Emotion } from '@emotion/core'
+import { EditorContextValue } from './types'
+
+const reducer = (state: Theme, next: Partial<Theme>) =>
+  merge<Theme>(state, next)
+
+type EditorProviderProps = {
```
hasparus, line 9:
```suggestion
export type EditorProviderProps = {
```

packages/editor/src/Sx/Colors.tsx
```diff
@@ -1,20 +1,30 @@
 /** @jsx jsx */
-import { jsx } from 'theme-ui'
+import { ColorState } from 'react-color'
+import { jsx, Theme } from 'theme-ui'
 import ThemeColorPicker from './ThemeColorPicker'
 
+type Color = string | ColorState
+type OnChangeArg = { color: Color } | { bg: Color }
+
+type ColorsProps = {
```
hasparus, line 9:
```suggestion
export type ColorsProps = {
```

packages/editor/src/Sx/Margin.tsx
```diff
@@ -0,0 +1,11 @@
+/** @jsx jsx */
+import { jsx } from 'theme-ui'
+import Space, { SpaceProps } from './Space'
+
+type MarginProps = Omit<SpaceProps, 'property'>
```
hasparus, line 5:
```suggestion
export type MarginProps = Omit<SpaceProps, 'property'>
```

packages/editor/src/Sx/Padding.tsx
```diff
@@ -0,0 +1,11 @@
+/** @jsx jsx */
+import { jsx } from 'theme-ui'
+import Space, { SpaceProps } from './Space'
+
+type PaddingProps = Omit<SpaceProps, 'property'>
```
hasparus, line 5:
```suggestion
export type PaddingProps = Omit<SpaceProps, 'property'>
```

packages/editor/src/Sx/Space.tsx
```diff
@@ -67,64 +74,59 @@ export const Space = ({
...
     <div
       sx={{
         display: 'grid',
-        gridGap: 2,
+        gap: 2,
         gridTemplateColumns: 'repeat(3, 1fr)',
         alignItems: 'center',
       }}>
-      <Field
+      <Field<'input'>
```
hasparus, line 89: Oh, is it still needed after the update to @theme-ui/components?

packages/editor/src/Sx/ThemeColorPicker.tsx
```diff
@@ -1,26 +1,37 @@
...
-import { jsx, useThemeUI } from 'theme-ui'
+import { jsx, useThemeUI, Theme } from 'theme-ui'
 import React from 'react'
 import tinycolor from 'tinycolor2'
-import { GithubPicker } from 'react-color'
+import { GithubPicker, ColorState } from 'react-color'
 import { usePopoverState, Popover, PopoverDisclosure } from 'reakit/Popover'
 
-export const ThemeColorPicker = ({ children, theme, ...props }) => {
+type ThemeColorPickerProps = {
```
hasparus, line 8:
```suggestion
export type ThemeColorPickerProps = {
```

packages/editor/src/Sx/Typography.tsx
```diff
@@ -1,26 +1,38 @@
...
 import { Fragment } from 'react'
 import { Field } from '@theme-ui/components'
 import Combobox from '../Combobox'
 
+type OnChangeArg =
+  | { fontFamily: string }
+  | { fontSize: number }
+  | { fontWeight: string }
+  | { lineHeight: string }
+type TypographyProps = {
```
hasparus, line 12:
```suggestion
export type TypographyProps = {
```

packages/editor/src/Theme/Colors.tsx
```diff
@@ -0,0 +1,66 @@
+/** @jsx jsx */
+import { jsx, useThemeUI } from 'theme-ui'
+// @ts-ignore: Remove @ts-ignore after @theme-ui/style-guide was transformed to TypeScript
+import { ColorPalette } from '@theme-ui/style-guide'
+import ColorPicker from '../ColorPicker'
+import { EditorContextValue } from '../types'
+
+type ColorsProps = {
```
hasparus, line 8:
```suggestion
export type ColorsProps = {
```

packages/editor/src/Theme/Colors.tsx
```diff
@@ -0,0 +1,66 @@
+/** @jsx jsx */
+import { jsx, useThemeUI } from 'theme-ui'
+// @ts-ignore: Remove @ts-ignore after @theme-ui/style-guide was transformed to TypeScript
```
hasparus, line 3: Is this @ts-ignore still needed?

packages/editor/src/Theme/LineHeights.tsx
```diff
@@ -0,0 +1,40 @@
+/** @jsx jsx */
+import { jsx, useThemeUI } from 'theme-ui'
+import { Fragment } from 'react'
+import { Field } from '@theme-ui/components'
+import { EditorContextValue } from '../types'
+
+export default () => {
```
hasparus, line 7:
```suggestion
export default function LineHeights() {
```

packages/editor/test/__snapshots__/ColorPicker.tsx.snap
```diff
@@ -4,7 +4,7 @@ exports[`renders with styles 1`] = `
 .emotion-9 {
   display: grid;
   padding: 8px;
-  grid-gap: 8px;
+  gap: 8px;
```
hasparus, line 7: Oh, this looks like unintended change. We'd like to preserve `gridGap` I think. Did you have any problem with the types with `gridGap`?

hasparus:
Hey @dcastil. I left a few comments, mostly nitpicks or Sand/Dust in Netlify's [nomenclature](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/).

I'm a bit worried about this `gridGap -> gap` though.

packages/editor/src/Theme/Colors.tsx
dcastil: Fixed in 3e034291.

packages/editor/src/Theme/LineHeights.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/Theme/Colors.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/Sx/Typography.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/Sx/ThemeColorPicker.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/Sx/Padding.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/Sx/Margin.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/Sx/Colors.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/EditorProvider.ts
dcastil: Fixed in 14451fab.

packages/editor/src/Combobox.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/ColorPicker.tsx
dcastil: Fixed in 14451fab.

packages/editor/src/ColorPicker.tsx
dcastil: Fixed in 14451fab.

dcastil: I just converted all `*Props` types to interfaces to make it possible to merge declarations. Maybe I misunderstood something and you just meant this one? 🤔 Let me know if I should change the others back.

packages/editor/src/Sx/Space.tsx
dcastil: Fixed in 698838e0.

packages/editor/test/__snapshots__/ColorPicker.tsx.snap
dcastil: Yeah I ran into a little rabbit hole with this one. 😄 It's intended, I had to change the [`gridGap`](https://github.com/system-ui/theme-ui/pull/900/files#diff-3a1a855ba5cdfdcc021fe760440538c7L86) property to [`gap`](https://github.com/system-ui/theme-ui/pull/900/files#diff-f952eb4f04acd9fa07eb7138709d1c4dR102) because TypeScript was complaining about the `gridGap` property not existing on the `SystemStyleObject` type.

That's because we define our CSS properties like this

```ts
export interface CSSProperties extends CSS.StandardProperties<number | string>, CSS.SvgProperties<number | string> {}
```

but only `gap` is part of `CSS.StandardProperties`. `gridGap` is part of `CSS.ObsoleteProperties` which are non-standard CSS props.

In this case `gridGap` is a deprecated prop and was changed to `gap` (so it can also be used for non-grid purposes). Technically `gap` might not work on some older versions of browsers which implemented the first version `gridGap`. But I'd think it would be more important to go forward with the standard.

What do you think? Should I rather include `CSS.ObsoleteProperties ` into the `CSSProperties` definition?

dcastil:
Hey @hasparus, thanks for the review! I fixed everything but the [`gap` issue](https://github.com/system-ui/theme-ui/pull/900#discussion_r427548249). Could you check again?

> mostly nitpicks or Sand/Dust in Netlify's [nomenclature](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/)

Ah cool, I know Netlify's Feedback Ladder!

packages/editor/src/ColorPicker.tsx
hasparus: I didn't want to ask for it except here for this intersection, but it's good. Thanks!

packages/editor/test/__snapshots__/ColorPicker.tsx.snap
hasparus: Goddamn, TypeScript knows CSS better than I do...

That's a good question.

Should we support obsolete CSS properties on type level? I suppose that's an "responsible adult" thing to do, because more properties will possible become obsolete in the future and we'd rather avoid breaking people's builds because of this. We could implement a warning at runtime if we'd like to enforce the standard, but type error may be a bit too harsh, and less user friendly.

cc @jxnblk

---

BTW One can still use `sx={{ color: 'primary', ...{ gridGap: 2 } }}` as a workaround for the obsolete property.

dcastil: > I suppose that's an "responsible adult" thing to do, because more properties will possible become obsolete in the future and we'd rather avoid breaking people's builds because of this.

I agree! Should I open a PR to add `CSS.ObsoleteProperties`?

PR #906, opened 2020-05-10
add to install theme-ui

ansonlowzf:
To be obvious to new developer, theme-ui module should be installed as well.

I directly land of this page with google search. 
I'm using CodeSandbox to test theme-ui, 
CodeSandbox not throw error. it's just not working

I'm confuse and thought gatsby-plugin-theme-ui include the theme-ui module 

Then I change to use VS Code
It throw an error said theme-ui module is not installed


hasparus:
Yeah, theme-ui is a peerDependency of the plugin.
It's properly documented in plugin's [README](https://github.com/system-ui/theme-ui/tree/master/packages/gatsby-plugin-theme-ui#gatsby-plugin-theme-ui), but not in Getting Started. Good PR.

PR #909, opened 2020-05-10
feat(chrome): update chrome to typescript

ivoreis:
Convert `chrome` package to typescript

I've added `@ts-ignore` to `@theme-ui/editor` since https://github.com/system-ui/theme-ui/pull/900 isn't merged yet. 

Relates to: https://github.com/system-ui/theme-ui/issues/668

### DIFF
--- a/packages/chrome/src/index.tsx
+++ b/packages/chrome/src/index.tsx
@@ -25,11 +25,20 @@ import {
 } from '@theme-ui/editor'
 import { Theme } from '@theme-ui/css'
 
-const runScript = (script: any) =>
+interface DevToolsExceptionInfo {
+  isError: boolean
+  code: string
+  description: string
+  details: any[]
+  isException: boolean
+  value: string
+}
+
+const runScript = (script: string) =>
   new Promise((resolve, reject) => {
     debounce(window.chrome.devtools.inspectedWindow.eval, 100)(
       script,
-      (result: any, err: Error) => {
+      (result: object, err: DevToolsExceptionInfo) => {
         if (err) {
           console.error(err)
           reject(err)
@@ -39,7 +48,9 @@ const runScript = (script: any) =>
     )
   })
 
-const mergeState = (state: any, next: any) => merge({}, state, next)
+type StateWithColorMode = { colorMode?: string; theme?: Theme }
+const mergeState = (state: StateWithColorMode, next: StateWithColorMode) =>
+  merge({}, state, next)
 
 const CopyTheme = ({ theme }: { theme: Theme }) => {
   const [copied, setCopied] = useState(false)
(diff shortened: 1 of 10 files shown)


ivoreis:
@hasparus / @jxnblk I'd appreciate some help about how to test this 😅

I've noticed that there's a top level script `dev:chrome` but that one doesn't work on master.

packages/chrome/src/index.tsx
```diff
@@ -51,7 +59,7 @@ const CopyTheme = ({ theme }) => {
 
 const Spacer = () => <div sx={{ my: 2 }} />
 
-const App = () => {
+const App: FunctionComponent = () => {
```
hasparus, line 62: `FunctionComponent` here is a bit redundant.

1. It states that this component takes children. (Try destructuring in props.)
2. We don't use `defaultProps` nor `propTypes` here.

packages/chrome/src/index.tsx
```diff
@@ -96,7 +104,7 @@ const App = () => {
     setColorMode,
   }
 
-  if (!context.theme) return false
+  if (!context.theme) return null
```
hasparus, line 107: Okay, I think I know where this FunctionComponent came from :D Probably no harm to leave it.

packages/chrome/src/index.tsx
```diff
@@ -15,13 +21,15 @@ import {
   LineHeights,
   FontSizes,
   Space,
+  // @ts-ignore
 } from '@theme-ui/editor'
+import { Theme } from '@theme-ui/css'
 
-const runScript = script =>
+const runScript = (script: any) =>
```
hasparus, line 28:
```suggestion
const runScript = (script: string) =>
```

https://developer.chrome.com/extensions/devtools_inspectedWindow#method-eval

packages/chrome/src/index.tsx
```diff
@@ -31,17 +39,17 @@ const runScript = script =>
     )
   })
 
-const mergeState = (state, next) => merge({}, state, next)
+const mergeState = (state: any, next: any) => merge({}, state, next)
```
hasparus, line 42: Is `state` and `next` `{ colorMode: string, theme: Theme }`? Just guessing tbh.

packages/chrome/src/index.tsx
ivoreis: Yeah that's fair enough but FunctionComponent is the default's React type for function based components, children (and other props) are optional too.

packages/chrome/src/index.tsx
ivoreis: I wasn't sure so that was why I left it open. I'll take a look to understand which type this is

packages/chrome/src/index.tsx
```diff
@@ -96,7 +114,7 @@ const App = () => {
     setColorMode,
   }
 
-  if (!context.theme) return false
+  if (!context.theme) return null
```
ivoreis, line 117: Replaced line [156](https://github.com/system-ui/theme-ui/pull/909/files#diff-715666d45aadf8a39f99bb87b894facdR156) with `context.theme` since we it's the same value as `state.theme` and we have a guarantee that `theme` is defined and/or not null.

ivoreis:
@hasparus anything pending on this PR or you reckon is good to be merged?

hasparus:
Hi man, the TypeScript side looks good. 
Thanks for the PR, it's really good.

I thought @types/lodash.merge and @types/lodash.debounce are missing, but this is a private package, so the ones from the monorepo are fine.

Looking at @theme-ui/editor and this import

```
import {
  Editor,
  Row,
  ColorPalette,
  ColorMode,
  Fonts,
  FontWeights,
  LineHeights,
  FontSizes,
  Space,
  // @ts-ignore
} from '@theme-ui/editor'
```

I kinda think the package is outdated.
Also the prepare script was renamed to `__prepare` in 0.3.0 alpha.
I bundled it and loaded it in my browser, but even after few changes in devtools.js I can't get it to work.

I suppose we can merge it and describe its current state in README.

@jxnblk Am I correct?

jxnblk:
> I kinda think the package is outdated.

Yeah, we never got this completely fleshed out, so it's probably missing a few pieces here and there. I don't want to quash any work here, but it might be worth discussing the merits of moving some of the remaining packages over to TS vs deprecating them. For this particular package, [Blocks UI](https://blocks-ui.com/) is shaping up to take it's place eventually IMO

ivoreis:
Cheers for the update @hasparus / @jxnblk. Anything else we wanna do with this PR or is it good to go?

hasparus:
It's good to go. We'll merge it soon.

Issue #907, opened 2020-05-10
Use const inside sx


arnaudjnn:
## What I have

```
<ul sx={{
   maxHeight: '280px',
   overflowY: 'auto',
   '&::-webkit-scrollbar, -ms-overflow-style': {
       display: 'none'
    }
}}>
   ...
</ul>
```
## What I want

```
const Scroll = {
   maxHeight: '280px',
   overflowY: 'auto',
   '&::-webkit-scrollbar, -ms-overflow-style': {
       display: 'none'
    }
}

<ul sx={Scroll}>...</ul>
```

## Error
```
177:25 Type '{ overflowY: string; maxHeight: string; '&::-webkit-scrollbar, -ms-overflow-style': { display: string; }; }' is not assignable to type 'SystemCssProperties | CSSPseudoSelectorProps | CSSSelectorObject | VariantProperty | UseThemeFunction | null | undefined'.
  Type '{ overflowY: string; maxHeight: string; '&::-webkit-scrollbar, -ms-overflow-style': { display: string; }; }' is not assignable to type 'CSSSelectorObject'.
    Property 'overflowY' is incompatible with index signature.
      Type 'string' is not assignable to type 'SystemStyleObject'.
    175 |                   sx={StyledMultiDataList}
    176 |                   render={({ data, handleChange }: RendreProps) => (
  > 177 |                     <ul sx={Scroll}>
        |                         ^
    178 |                       {data.map(category => {
    179 |                         if(category.childrens.length > 0) {
    180 |                           return (

**Comment by hasparus** [comment]
Okay, so this is TypeScript file, right?

The error message is not really cool, we should try to poke around it to improve it,
but what it means is `overflowY` is not a `string`.

**Comment by hasparus** [comment]
You can encounter similar problem while using `xstate`.
https://github.com/davidkpiano/xstate/issues/310

**Comment by hasparus** [comment]
This might be a good thing to add to the docs.

cc @deadcoder0904 @jxnblk

Relevant: https://xstate.js.org/docs/guides/typescript.html

---

```
Type 'string' is not assignable to type 'SystemStyleObject'
```

This appears because we try to match to recursive object last.
I'll try reordering some stuff in our types to improve the error message.

Let's keep this issue open for this and a possible docs improvement.

Issue #834, opened 2020-04-08
Usage of Theme UI with TypeScript


deadcoder0904:
TS noob here. I have installed `@types/theme-ui` as well as have the following in `theme.ts`:

```js
export const theme = {
  fonts: {
    body:
      'system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", sans-serif',
    heading: '"Avenir Next", sans-serif',
    monospace: 'Menlo, monospace',
  },
  colors: {
    text: '#000',
    background: 'white',
    primary: '#bf1',
  },
  modes: {
    dark: {
      text: '#fff',
      background: '#000',
      primary: '#0cf',
    },
  },
}
```

How do I convert it to TS so I can get autocomplete?

I tried the following:

```js
import {Theme} from 'theme-ui'

export const theme: Theme = {
  fonts: {
    body:
      'system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", sans-serif',
    heading: '"Avenir Next", sans-serif',
    monospace: 'Menlo, monospace',
  },
  colors: {
    text: '#000',
    background: 'white',
    primary: '#bf1',
  },
  modes: {
    dark: {
      text: '#fff',
      background: '#000',
      primary: '#0cf',
    },
  },
}
```

But I get red squiggly lines on `modes` with the error

> Object literal may only specify known properties, and ‘modes’ does not exist in type ‘Theme’

I tried finding what should be the type in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/theme-ui/index.d.ts but I can’t find it. Then I saw https://github.com/system-ui/theme-ui/issues/668 to wonder if it’s even complete. I just want the `theme` object to have correct keys :)

TheFirstMe:
@deadcoder0904 modes should be under colors? [Source](https://theme-ui.com/color-modes#defining-colors)

deadcoder0904:
Woah looks like it. I did that & it worked. Not sure where I found the above code but I vividly remember copying it from somewhere online 😂

Anything else that needs to be using TypeScript?

TheFirstMe:
I’m not sure I don’t do TypeScript.

erikdstock:
When you are starting a project from scratch, the more you can keep to using types the better - use `any` as an escape hatch when necessary, and learn how to contribute missing types (it is confusing for sure). For your theme interface what you have looks fine to me!

[...]

Issue #832, opened 2020-04-07
Theme UI v1 Roadmap


jxnblk:
This is an umbrella issue for features and additions to bring Theme UI to a stable v1. This list may not be exhaustive and is subject to change. More details, context, and additional issues will be added where needed.

- [ ] TypeScript conversion #668
- [ ] TypeScript: Strict-mode theme interface/schema: a fully fleshed out theme schema to improve interoperability.
- [ ] TypeScript: Type check property values against theme scales
- [ ] Variant composition, e.g. `variant: ['large', 'primary']` (or plural `variants`). This would potentially be a breaking change, meaning that the `variant` property itself would no longer support responsive arrays, but the values within the variant could. #586
- [ ] Babel plugin for JSX pragma #479
- [ ] Upgrade to Emotion 11 https://github.com/emotion-js/emotion/issues/1606
- [ ] Use Theme UI context instead of Emotion context
- [ ] Remove "default" theme – remove default `fontSizes` and `space` values
- [ ] `gatsby-plugin-theme-ui`: support configuration via `src/theme.{js,json,ts}`
- [ ] `gatsby-plugin-theme-ui`: support for presets via `gatsby-config.js`
- [ ] Scope all [configuration options](https://theme-ui.com/theming#configuration-flags) to `theme.config`
- [ ] Improve `prefers-color-scheme` media query support – allow color mode to be determined by media query without storing in `localStorage`. See #624, #787
- [ ] Support for CSS Logical Properties #794
- [ ] **Idea:** Simpler API for using presets
- [ ] **Idea:** Provide some solution for self-referencing scales in other scale definitions
- [ ] **Idea:** Plain object syntax responsive values using a `media` key. e.g. `media: { small: { fontSize: 3 } }`
- [ ] **Idea:** Nested property shorthands `font: { size: 3, weight: 'bold' }` [example branch](https://github.com/system-ui/theme-ui/compare/css-shorthands?expand=1)
- [ ] **Idea:** Universal `variant` prop, see #800
- [ ] **Idea:** Merge variants before parsing responsive arrays. This should allow responsive values defined in e.g. `theme.styles` to be overridden with non-responsive values with the `sx` prop. Related #720, #43

gausie:
@hasparus out of interest, why did you unpin this issue?

hasparus:
> @hasparus out of interest, why did you unpin this issue?

By accident. Thank you. Let's keep this one pinned.

LekoArts:
[...]

PR #910, opened 2020-05-10
Improve type error messages and tooltips

hasparus:
Okay this might be a small thing, but I want to add some polish (lol) to the types.

Generally, the problem we have is that our types, mostly imported from DT, may be a bit confusing.
I'd like to make Theme UI types as welcoming as possible for v1 (#832)

**Related issues:**

- https://github.com/system-ui/theme-ui/issues/834#issuecomment-625618512
  - [ ] Add TypeScript section to the docs to explain most common tricky things people can see
- https://github.com/system-ui/theme-ui/issues/834#issuecomment-625703446
  - [ ] Move all tests to TypeScript. We can't be sure Theme UI works in TypeScript if we don't typecheck our tests.
  - [x] Get rid of generic parameter for `merge`. Use `Theme` as argument.
- https://github.com/system-ui/theme-ui/issues/907
  - [ ] Get inspired by Emotion's `Interpolation` and display errors on `SystemStyleObject` better.
- We don't support vendor properties. **Why?** They are defined in `csstype` and we actually have them inside `StandardCSSProperties`, but they don't come to `sx` prop. I can't see a reason we shouldn't support vendor properties on type level, while they work at runtime.
  - [x] Let's add `CSS.VendorProperties<number | string>` to `CSSProperties` type.
- [ ] Add `label?: string` to `ThemeUIStyleObject` (https://github.com/DefinitelyTyped/DefinitelyTyped/issues/45020#issuecomment-633300398)
- [x] Rename `SystemStyleObject` to `ThemeUIStyleObject`.
  - System is a loaded word in our industry, and the name is inherited from styled-system. It might be more user-friendly to tidy up few types and use a mix of [nice meaningful names](https://jxnblk.com/blog/design-graph/) and dumb straightforwad names.
  
---

Few notes to explain what I'm doing.

- Interfaces can be merged with interfaces of the same name.
  - https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces
- Type aliases expand in error messages and tooltips.
- Interfaces don't allow property name conflicts in extended types.
  Intersections (`&`) turn a conflicting prop into `never`. This isn't something we often want.
  [(TS Playground with example)](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDeUAeAuKByAhlqAvgNwBQokUAQgsmplgEb7EklIlRTnQzWJwAyKqU7dYlABo1eAbSyosAXSgB6FVAA8AWigA7CADcIAJxIFW7TmqgBjAPa6AZgBsAljeABCW3YC2YV2cILWBXX2gTYztTTlddYBNHHBseaghUBN0AEwBnWAAaKmRzcyA)


hasparus:
@jxnblk Which colors in `theme.colors` should be required? ATM it's `primary`, `background` and `text`.

packages/theme-ui/test/color-modes.tsx
```diff
@@ -373,14 +372,14 @@ test('warns when initialColorModeName matches a key in theme.colors.modes', () =
 })
 
 test('dot notation works with color modes', () => {
-  const Button = (props) => {
```
hasparus, line 376: damn, can't Prettier decide on one style? I'll revert this change.

hasparus:
This is starting to become pretty big, and while I didn't tackle all todos yet, it might be good to start reviewing before it becomes overwhelming.

jxnblk:
> Which colors in theme.colors should be required?

Generally speaking, none should be *required,* but the color mode functionality does depend on the `colors.text` and `colors.background` keys being defined.

For the interoperability part, I think we'll want to create a "blessed" Theme interface (along with some color palette interfaces) after the TS conversion is out as an optional piece of the puzzle, but not sure if that's helpful here

packages/core/src/types.ts
```diff
@@ -1,11 +1,11 @@
-import { SystemStyleObject } from '@theme-ui/css'
+import { ThemeUIStyleObject } from '@theme-ui/css'
```
jxnblk, line 1: Not to bikeshed, but this is a *much* better name :)

PR #917, opened 2020-05-11
feat(sidenav): update sidenav to typescript

ivoreis:
Convert to typescript:
- [x] Sidenav

Relates to: #668


ivoreis:
@hasparus  Mind having a look please?

packages/sidenav/src/index.tsx
```diff
@@ -0,0 +1,381 @@
...
+      }}
+    />
+  </React.Fragment>
+)
+
+export const Sidenav = forwardRef<
+  HTMLDivElement,
+  {
+    open?: boolean
+    components: MDXProviderComponents
```
hasparus, line 86: This was previously treated as `MDXProviderComponents | undefined`, right?
Could we default to empty object or even leave undefined if it's okay with MDXProvider types?

packages/css/src/types.ts
```diff
@@ -50,7 +50,7 @@ interface CSSOthersObjectForCSSObject {
 /**
  * Map all nested selectors
  */
-interface CSSSelectorObject {
+export interface CSSSelectorObject {
```
hasparus, line 53: Are you using it in this PR?
It's a part of StyleSystemObject, so it makes sense to export it.
I'm just asking if I missed something.

yarn.lock
```diff
@@ -4041,6 +4041,13 @@
   resolved "https://registry.yarnpkg.com/@types/json-schema/-/json-schema-7.0.3.tgz#bdfd69d61e464dcc81b25159c270d75a73c1a636"
   integrity sha512-Il2DtDVRGDcqjDtE+rF8iqg1CArehSK84HZJCT7AMITlyXRBpuPhqGLDQMowraqqu1coEaimg4ZOqggt6L6L+A==
 
+"@types/mdx-js__react@^1.5.1":
```
hasparus, line 4044: We should have this package in `dependencies` of `@theme-ui/sidenav`, right?

packages/sidenav/package.json
```diff
@@ -3,6 +3,8 @@
   "version": "0.4.0-alpha.1",
   "main": "dist/index.js",
   "module": "dist/index.esm.js",
+  "types": "dist/index.d.ts",
+  "sideEffects": false,
```
hasparus, line 7: Good change.

packages/sidenav/src/index.tsx
ivoreis: Yeah and that was probably an oversight. Making components optional / allowing undefined will cause a type-check to fail with MDXProvider that requires components to be defined.

Screenshot 2020-05-16 at 09 10 13

packages/css/src/types.ts
ivoreis: Good point and I was using `SystemStyleObject` at some point but then I refactored it and while doing that I've noticed that we weren't exporting all the possible types of:
https://github.com/system-ui/theme-ui/blob/076cd7b5cab2a8a58f96da9f148a33590588f169/packages/css/src/types.ts#L467-L472
And this would throw a typecheck error saying that some types couldn't be found (since they weren't exported).

Are you ok having this here or shall I raise a different PR with it @hasparus ?

package.json
```diff
@@ -27,6 +27,7 @@
     "@babel/runtime": "^7.7.7",
     "@testing-library/react": "^9.1.3",
     "@types/jest": "^25.1.2",
+    "@types/mdx-js__react": "^1.5.1",
```
ivoreis, line 30: @hasparus https://github.com/system-ui/theme-ui/pull/917/files#r425106843 

I've added it here since it's a devDep and we're in a monorepo environment

yarn.lock
ivoreis: @hasparus please see here https://github.com/system-ui/theme-ui/pull/917/files#r426131822

package.json
hasparus: We're building a package though.

Our emitted types contain types imported from _@types/mdx-js__react_. (`MDXProviderComponents`)

We either install it for the user, or we require installing it in docs and peerDeps.
If we just leave it here, sidenav props `components` will be either `any` or fail the build for the user depending on their `skipLibCheck` setting.

packages/sidenav/src/index.tsx
hasparus: We can default to `{}` to avoid a breaking change.

package.json
ivoreis: yeah you're right 👍 I'll update it

packages/sidenav/src/index.tsx
ivoreis: Updated

package.json
ivoreis: Updated

yarn.lock
ivoreis: DevDep moved into package

packages/sidenav/src/index.tsx
```diff
@@ -0,0 +1,381 @@
...
+      }}
+    />
+  </React.Fragment>
+)
+
+export const Sidenav = forwardRef<
+  HTMLDivElement,
+  {
+    open?: boolean
+    components?: MDXProviderComponents
```
ivoreis, line 86: Keep components optional

packages/sidenav/package.json
```diff
@@ -18,6 +20,9 @@
     "react": "^16.8.6",
     "theme-ui": "^0.4.0-alpha.1"
   },
+  "devDependencies": {
+    "@types/mdx-js__react": "^1.5.1"
```
hasparus, line 24: Okay, but it's still a dev dependency. It doesn't exist for the end user.

Take a look here https://stackoverflow.com/a/46011417/6003547.

packages/sidenav/test/index.tsx
```diff
@@ -1,64 +1,86 @@
...
-      </ul>
+      </Ul>
     </li>
   </ul>
 )
 
 test('Sidenav renders', () => {
-  const json = renderer.create(<Sidenav>{links}</Sidenav>).toJSON()
+  const json = renderer
+    .create(<Sidenav components={{}}>{links}</Sidenav>)
```
hasparus, line 71: Since `components` are optional again, we can get back to to the old test here.

packages/css/src/types.ts
hasparus: I'm working on this part in https://github.com/system-ui/theme-ui/pull/910, because our type errors on `sx` are wonky.
Quite possibly, these types will change to something closer to @emotion/core Interpolation types.
Just a disclaimer in case you import them in other projects.

packages/sidenav/package.json
ivoreis: Gotcha 😄 Updated that to be a dep 👍

packages/sidenav/test/index.tsx
ivoreis: Done

packages/sidenav/package.json
ivoreis: Done

hasparus:
Okay, I'm merging it right now.
I got rid of `as ReactNode` assertions. It seems they're not needed.

`as SxStyleProp` can also be removed, but it needs a sort of weird syntax.

```tsx
        <div
          {...props}
          ref={ref}
          sx={{
            position: ['fixed', 'sticky'],
            top: 0,
            left: 0,
            bottom: [0, 'auto'],
            zIndex: 1,
            minWidth: 0,
            width: 256,
            maxHeight: ['100vh', 'none'],
            overflowX: 'visible',
            overflowY: 'auto',
            transition: 'transform .2s ease-out',
            transform: [open ? 'translateX(0)' : 'translate(-100%)', 'none'],
            bg: ['background', 'transparent'],
            ...{ WebkitOverflowScrolling: 'touch' },
            // WebkitOverflowScrolling is not supported in our types
          }}
        />
```

I just wanted to mention the workaround for posterity. For this particular case, I think I have a fix.

PR #923, opened 2020-05-14
Fix broken link

brunolemos:


packages/theme-ui/README.md
```diff
@@ -77,7 +77,7 @@ export default props => (
 )
 ```
 
-The `theme` object follows the System UI [Theme Specification](/theme-spec),
+The `theme` object follows the System UI [Theme Specification](/packages/docs/src/pages/theme-spec.mdx),
```
lachlanjc, line 80: This breaks the link, I’m pretty sure—the page is live here: https://theme-ui.com/theme-spec/ but this link doesn’t work: https://theme-ui.com/packages/docs/src/pages/theme-spec.mdx

brunolemos: ah, I'm basing it on the github readme, not the website. so I guess it needs to be an absolute URL to work on both.

jxnblk: Links like this are likely a result of copying and pasting links from the docs site source, but for links like this in the READMEs, they should probably be absolute URLs

jxnblk:
Awesome, thanks!!

PR #937, opened 2020-05-18
merge master to enable-highlighting

hasparus:
Okay, the git diff here isn't really fun. The only real change here is in [prism/src/index](https://github.com/hasparus/theme-ui/blob/cda32af7f66df15d74f36998c69a0d941c97b2e4/packages/prism/src/index.tsx). Everything stayed in the same place.


laurieontech:
Ha! You weren't kidding. But looks good to me. What a pain that Token is exported.

hasparus:
> What a pain that Token is exported.

Yeah, super annoying. It's part of public the API anyway, it could just be exported.

I considered copying it, but I'm afraid they'll do a non-breaking change to it and we'll wonder why the two Tokens are different.

PR #938, opened 2020-05-18
Move @theme-ui/components types to this repo

hasparus:
Okay, this should fix our problems with `@types/theme-ui` on GitHub Actions.

### DIFF
--- a/packages/components/index.d.ts
+++ b/packages/components/index.d.ts
@@ -1,0 +1,366 @@
+import * as React from 'react'
+import { StyledComponent } from '@emotion/styled'
+import { InterpolationWithTheme } from '@emotion/core'
+import { SpaceProps, ColorProps, MarginProps } from 'styled-system'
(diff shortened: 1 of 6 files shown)


hasparus:
However, I have no idea how this happens on Netlify. The package is in dependencies, it has to be installed.

> error Error in "/opt/build/repo/packages/docs/gatsby-node.js": Cannot find module '@theme-ui/presets'

jxnblk:
This PR LGTM -- not sure what's happening with Netlify since it seems like it's failing with something different now:
> Error: /opt/build/repo/packages/style-guide/src/context.ts(1,28): semantic error TS2307 Cannot find module 'theme-ui'.

hasparus:
If what I did right now in https://github.com/system-ui/theme-ui/pull/938/commits/e76823f44d90e31858a8700772e64db50c6de633 is connected, we were just missing a devDependency, so the packages were installed and built in wrong order.

PR #941, opened 2020-05-19
[`@theme-ui/style-guide`] Add missing size for ColorRows that are scales

knowler:
Without this, a `size` prop applied to a `ColorPalette` is only picked up for top-level colors that are not scales (e.g. typically `primary`, `text`, `background`).


jxnblk:
Awesome, thanks!

PR #943, opened 2020-05-21
Fix Safari closing sidebar on accordion button click

dcastil:
Safari removes focus from the sidebar when a button is clicked, triggering

```jsx
<div onBlur={() => setMenuOpen(false)}>
  <Sidebar />
</div>
```

I've set

```jsx
<button onMouseDown={(e) => e.preventDefault()} />
```

to prevent blur when clicking an AccordionButton.

Before:

Kapture 2020-05-21 at 13 25 15

After:

Kapture 2020-05-21 at 13 21 41

### DIFF
--- a/packages/sidenav/src/index.tsx
+++ b/packages/sidenav/src/index.tsx
@@ -152,6 +152,7 @@ export const AccordionButton: FunctionComponent<{
       title="Expand Section"
       disabled={disabled}
       {...props}
+      onMouseDown={(e) => e.preventDefault()}
       sx={{
         appearance: 'none',
         display: 'flex',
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks! Clicking on the demo seems to work, but this looks a little heavy-handed at first glance -- do you think there might be a better way to handle the accordion cross-browser, generally speaking??

dcastil:
I searched for solutions to this, but `<button onMouseDown={(e) => e.preventDefault()} />` seems to be the general solution. The `blur` event happens on `mousedown`, so that's the only place where it can be handled effectively.

```jsx
<button onMouseDown={(e) => e.currentTarget.focus()} />
```

also doesn't seem to work unfortunately.

Exact focus behaviour:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Clicking_and_focus
* https://zellwk.com/blog/inconsistent-button-behavior/#results

Suggestions for workaround:
* https://stackoverflow.com/questions/42758815/safari-focus-event-doesnt-work-on-button-element
* https://forum.vuejs.org/t/blur-before-click-only-on-safari/21598

jxnblk:
Thanks for the additional context! Seems legit, after thinking about this more!

PR #944, opened 2020-05-21
Update forms.mdx. Use `id` instead of `name` for `htmlFor`

yuyokk:
Team,

I've updated this page to point `Label`.`htmlFor` to [element id](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for) instead of name.

Please let me know if it makes sense for you.
Thanks

### DIFF
--- a/packages/docs/src/pages/components/forms.mdx
+++ b/packages/docs/src/pages/components/forms.mdx
@@ -26,6 +26,7 @@ import {
   onSubmit={e => e.preventDefault()}>
   <Label htmlFor='username'>Username</Label>
   <Input
+    name='username'
     id='username'
     mb={3}
   />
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/components/forms.mdx
```diff
@@ -26,13 +26,13 @@ import {
   onSubmit={e => e.preventDefault()}>
   <Label htmlFor='username'>Username</Label>
   <Input
-    name='username'
+    id='username'
```
jxnblk, line 29: Thanks! These elements should probably have both `name` and `id` attributes, shouldn't they?

yuyokk: As far as I know `name` can be skipped for labels access. If you want, I can add it back :)

yuyokk: @jxnblk added back. Let me know if anything else you think we need to change.

jxnblk:
Thanks!

PR #945, opened 2020-05-21
Fix inconsistent tabbing in docs navbar

dcastil:
While working on #943 I noticed some inconsistencies in the docs navbar.

* Multiple outline styles for different buttons
* SkipLink doesn't follow color scheme which makes it very obtrusive on dark mode

I opened a PR right away because the changes could be made quickly. If those things are on purpose, feel free to close this PR. 😊

| Before | After |
| --- | --- |
| Chrome light before | Chrome light after |
| Safari light before | Safari light after |
| Chrome dark before | Chrome dark after |
| Safari dark before | Safari dark after |

### DIFF
--- a/packages/docs/src/components/button.js
+++ b/packages/docs/src/components/button.js
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { jsx } from 'theme-ui'
 
-export default props => (
+export default (props) => (
   <button
     {...props}
     sx={{
@@ -16,9 +16,6 @@ export default props => (
       bg: 'muted',
       border: 0,
       borderRadius: 2,
-      ':focus': {
-        outline: '2px solid',
-      },
     }}
   />
 )
--- a/packages/docs/src/components/menu-button.js
+++ b/packages/docs/src/components/menu-button.js
@@ -16,7 +16,7 @@ const Burger = ({ size = '1em' }) => (
   </svg>
 )
 
-export default props => (
+export default (props) => (
   <button
     title="Toggle Menu"
     {...props}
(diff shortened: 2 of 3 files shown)


jxnblk:
Thanks for this! These were intentional, but could you elaborate on what you're trying to improve upon here or what the issue was?

dcastil:
I tried to solve two issues I had with the tabbing experience:

### Outlines

The outline around a focused element looks different dependent on the focused element.The hamburger menu and the color mode switch have a thicker black outline with sharp edges while the rest of the elements has the browser-native blue-ish outline with rounded corners.

I find a consistent outline more aesthetically pleasing and possibly less confusing (that part is negligible in the light theme, but might become more important with other themes).

So I removed the custom outline from both hamburger menu and color mode switch.

### SkipLink

The SkipLink is black text on white background, regardless of color mode.

I usually use dark mode on websites when available and get used to the dark look very quickly. I find it very unpleasing when there is sudden swap in text-color and bg-color. It always takes a little while so my eyes get used to read text that way (the problem is much bigger when reading a light section in a dark-themed website than the other way around). A good negative example is the dark mode the [TypeScript website](https://www.staging-typescript.org/docs/handbook/typescript-in-5-minutes.html) (visit it with dark mode enabled in your OS) implemented.

Screenshot 2020-05-22 at 21 57 17

I fixed it by using the current `text` and `background` colors.

dcastil:
Another option would be to add a custom outline to all elements instead of using the default one. That would make the look more consistent across browsers and fix the low contrast in the Safari outline in dark mode.

But then it's also easy to forget to implement that on any new element which would make it less consistent.

jxnblk:
Thanks for the context! For the skip nav link, part of the rationale in keeping that black on white is to be more visible and appear as expected for users who might have colors inverted or custom styles -- i.e. I think it's good for that one to not match custom styles, since they might pose issues.

For the other focus states, I think making them more consistent makes sense, so maybe those all can rely on browser defaults. What do you think?

dcastil:
Good to know, I agree with that! I removed the changes from SkipLink.

PR #948, opened 2020-05-22
useBreakpointIndex: Do not depend on window during setState initialization

### LINKED ISSUE #924: useBreakpointIndex defaultIndex breaks hydration
**Describe the bug**

You should never ever initialize React state differently on server and client (ie. any use of `typeof window === "undefined"` or `process.browser` must be avoided). Not respecting this will cause the UI to be misrendered (not correctly updated after hydration). If you are lucky then React will print a warning (eg. Warning: Prop `className` did not match. Server: "css-14r1nkj" Client: "css-37oxbl"). If you are less lucky, you won't see any warning and wonder why the UI is wrong.

**To Reproduce**

https://codesandbox.io/s/kind-satoshi-65goi

The data-index attribute on the div is not updated, it remains 0 after client loads on a larger screen (but the inner text of the div is updated correctly).

**Relevant Code**

https://github.com/system-ui/theme-ui/blob/6663b23facd5e3cd797f9cede10c08599a73b83f/packages/match-media/src/index.ts#L15-L34

The `getIndex()` function is used to initialize the default value... but the value is not updated after client hydration (ie. inside useEffect).

wereHamster:
Fixes #924

### DIFF
--- a/packages/match-media/src/index.ts
+++ b/packages/match-media/src/index.ts
@@ -12,39 +12,34 @@ export const useBreakpointIndex = (options: defaultOptions = {}) => {
   const breakpoints =
     (context.theme && context.theme.breakpoints) || defaultBreakpoints
 
-  const getIndex = useCallback(() => {
-    if (typeof window === 'undefined') {
-      if (typeof defaultIndex === 'number') {
-        if (defaultIndex < 0 || defaultIndex > breakpoints.length - 1) {
-          throw new RangeError(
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! I think this looks good, but would there be any potential "breaking" changes that this might cause?

wereHamster:
A potential layout shift. And of course if someone is relying on the current behaviour where the page may be misrendered until the user first resizes the window then PR will be a breaking change to them.

jxnblk:
Thanks, makes sense! I think we could go ahead and get this in with the v0.4 work, but we'll need to make note of it in the CHANGELOG

PR #951, opened 2020-05-22
Support a "default" key for object in scales

### LINKED ISSUE #936: feature request: "default" key for objects in scales
We're overhauling our colors. Formerly we had things like `colors.green`, `colors.darkGreen`, `colors.lightGreen`. We now would like to use a format like `colors.green.dark` or `colors.green.500`.

The issue is maintaining a backwards-compatible map of former colors to current colors. As-is, we can't have

```js
// colors
green: {
  dark: ...,
  core: ...,
  light: ...,
}
```

because `colors.green` is an object in the new world, but before `colors.green` resolved to a color value. In order to ship this new theme, we would need to update every callsite that used a color-theme-aware `sx` rule to use `colors.green.core`.

I would like to set a "default" value for objects, such that `colors.green` resolves to `colors.green.default` if `colors.green` is an object value and there is no subsequent `.$key` accessor.

We could work around this by using `colors.greens` (plural), avoiding the namespace collision. I dislike this because we would still have to maintain a `colors.green` that matches `colors.greens.default` by hand in order to change existing usages of `colors.green`. I also dislike this on aesthetic grounds (I just would prefer not having plurals in the namespace)

**Additional context**
Open to discussion on implementation particulars, I just wanted to get your thoughts on whether this sort of backwards-compatibility support is something you think belongs in theme-ui or should be the responsibility of consuming libraries.

folz:
The implementation is straightforward, but I wanted to discuss something that I think is a pretty large change to the meaning of `get` and/or theme paths.

Before this PR, theme-ui does not enforce that `get` on a theme path will return a primitive value suitable for inclusion in CSS. It assumes either that the theme author will write the theme such that all paths lead to primitive values (that is, avoid using objects or arrays at the end of a path), or that the end programmer will ensure all paths will point to a primitive value (that is, `'red.500'`, never `'red'`).

This PR changes the above by moving some of the "ensure path points to a primitive" responsibility to theme-ui, by establishing the convention that `path => object ? object.default`. This is a special case, though. Why not `path => object ? object.500`, or `path => object ? object.core`?  Should we support this for object-like values such as Map? What about other ways to store values: `path => array ? array['default']`? `path => symbol ? symbol.description`?

theme-ui also allows for `sx` to contain functions: `sx={{color: (theme: Theme) => theme.colors.blue.500}}`

This is less ergonomic but more precise than the `object.default` convention. It's also more typesafe, thanks to the `Theme` type. What if we allowed the theme to also contain themeGet functions? `path => function ? function(theme)`. This would allow things like

```
colors: {
  default: themeGet('colors.500'),
  500: 'color500',
  ...
}
```

Is that a reasonable extension of the "ensure path points to a primitive" responsibility?

Anyway, we don't need to answer literally every question above. Maybe the single special case of `object => default` is fine here. But I still wanted to bring up the possible implications to have a discussion about.

Closes #936


jxnblk:
Thanks for tackling this! I think this is a great addition to the API as-is -- it also sets up a nice convention for handling default variants.

> What if we allowed the theme to also contain themeGet functions? 

I think this *is* something we'd like to solve for. In the v1 issue, I put this as *Idea: Provide some solution for self-referencing scales in other scale definitions.* I don't think we need to solve for that in this PR, but feel free to take a stab at proposing what that could look like, if you have time

As far as the *ensure path points to a primitive* point, that's something I hadn't really considered, but might be good to at least warn users if an object or invalid value is being passed on as CSS.

packages/css/test/index.ts
```diff
@@ -6,6 +6,17 @@ const theme: Theme = {
     secondary: 'cyan',
     background: 'white',
     text: 'black',
+    colorScale: {
+      default: 'color500',
+      100: 'color100',
```
jxnblk, line 11: Nit: I'd prefer to see valid CSS color values here -- I think it helps with reading the tests and making sure it's doing what you'd expect to happen

packages/css/test/index.ts
```diff
@@ -206,13 +217,44 @@ test('works with functional arguments', () => {
...
+    color: 'color100',
+  })
+})
+
+test('returns object when accessing object value with no default key', () => {
+  const result = css({
+    color: 'colorScaleNoDefault',
+  })(theme)
+  expect(result).toEqual({
+    color: {
```
jxnblk, line 250: Might be good to add a comment to state that this (although expected) won't result in valid CSS, in case someone stumbles upon this test

packages/css/test/index.ts
```diff
@@ -206,13 +217,44 @@ test('works with functional arguments', () => {
...
+
+test('returns nested key when accessing key from object value with default', () => {
+  const result = css({
+    color: 'colorScale.100',
+  })(theme)
+  expect(result).toEqual({
+    color: 'color100',
+  })
+})
+
```
jxnblk, line 244: It'd be nice to have an additional test showing this working with a variant and returning an object

jxnblk (approved):
Left a few nits/comments, but nothing blocking -- this looks great, thanks!

folz:
Thanks for the review! I made the changes you suggested. I'll open a new issue/PR about themeGet functions in themes.

folz:
Opened an issue to track `themeGet` reference support here: https://github.com/system-ui/theme-ui/issues/961

jxnblk:
Sweet! Thanks, I think this looks good to go, but want to let it sit for a bit so others have time to review or comment

hasparus:
This is a really good PR. Why didn't we merge it?
I'll handle the conflicts.

hasparus:
@folz, I'm gonna change this to `__default` to ensure people are using it knowingly. It turned out to be a bit surprising.

PR #964, opened 2020-05-29
Link CSS properties to MDN

shawnbot:
Links to MDN will help folks who aren't familiar with all of the properties figure out why they'd use these props.

### DIFF
--- a/packages/docs/src/components/theme-scales.js
+++ b/packages/docs/src/components/theme-scales.js
@@ -44,7 +44,7 @@ export default props => {
                 {table[key].map((property, index) => (
                   <Styled.inlineCode>
                     {!!index && ', '}
-                    {property}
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/theme-scales.js
```diff
@@ -44,7 +44,7 @@ export default props => {
                 {table[key].map((property, index) => (
                   <Styled.inlineCode>
                     {!!index && ', '}
-                    {property}
+                    <a href={`https://developer.mozilla.org/en-US/docs/Web/CSS/${property}`}>{property}</a>
```
jxnblk, line 47: Using the `Styled.a` component should pick up the theme's link files:
```suggestion
                    <Styled.a href={`https://developer.mozilla.org/en-US/docs/Web/CSS/${property}`}>{property}</Styled.a>
```

jxnblk:
This looks good to me. Anything else you were working on (noticed this is marked as a *draft*) or is this good to go?

PR #965, opened 2020-05-29
Add variants to Theme interface

wKovacs64:
Per request from @hasparus in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45042, opening this PR to add variants to the `Theme` interface that match the Theme UI documentation for each component as well as here:

- https://theme-ui.com/theme-spec#variants
- https://theme-ui.com/components/variants

Some things to note:

1. `buttons` and `text` were already present, but incorrect (AFAICT) - probably leftover from the initial `@types/styled-system__css` copypasta. We may need to do something different if I've misunderstood these properties, though.
1. This change required removing everything from the `ThemeSketchy` interface in `@theme-ui/preset-sketchy`, so I just changed preset to use `Theme` directly. If there are intentions to make it a different API than a base theme, we can put it back and just leave it as an empty interface or maybe a type alias until there are properties to add. I spot-checked a few other presets and this was the only one that was explicitly typed.
1. I might've gone a little overboard on the doc comments. 😶 Let me know if you have any feedback one way or the other and I can tweak them accordingly.

### DIFF
--- a/packages/css/src/types.ts
+++ b/packages/css/src/types.ts
@@ -586,11 +586,10 @@ export interface Theme {
   radii?: ObjectOrArray<CSS.BorderRadiusProperty<TLengthStyledSystem>>
   shadows?: ObjectOrArray<CSS.BoxShadowProperty>
   zIndices?: ObjectOrArray<CSS.ZIndexProperty>
-  buttons?: ObjectOrArray<ThemeUICSSProperties>
   colorStyles?: ObjectOrArray<ThemeUICSSProperties>
   textStyles?: ObjectOrArray<ThemeUICSSProperties>
-  text?: ObjectOrArray<ThemeUICSSProperties>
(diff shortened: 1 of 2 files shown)


hasparus:
> We may need to do something different if I've misunderstood these properties, though.

Thanks!
That was bad 😅 You didn't misunderstand anything.

> This change required removing everything from the ThemeSketchy interface in @theme-ui/preset-sketchy, so I just changed preset to use Theme directly.

That's great!
`preset-sketchy` was born in [a separate repository](https://github.com/beerose/theme-ui-sketchy), and it needed this workaround to make it work in TypeScript.

> I might've gone a little overboard on the doc comments. no_mouth Let me know if you have any feedback one way or the other and I can tweak them accordingly.

I, personally, love them. They make getting started with the library easier, and bring DX in IDEs closer to what "language support level".

image

PR #977, opened 2020-06-02
docs(how-it-works): change ThemeProvider import

hellobrian:
This changes the `import` statement for `ThemeProvider` so that it's from `theme-ui` instead of `emotion-theming`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks, but this is intended to show how the implementation works and isn't a mistake

jxnblk:
Yeah, no worries -- if you have any suggestions on how to make that doc clearer, feel free to open another PR!

PR #981, opened 2020-06-05
Use media query value when useColorSchemeMediaQuery is set to true

### LINKED ISSUE #787: Unexpected `prefers-color-scheme` behaviour
**Describe the bug**

When toggling between light/dark mode on a supoprted device (in this case iOS, or MacOS via Firefox), Theme-UI's color scheme does not update. Updating the app to reflect the user's color mode requires the localstorage to be deleted.

> but may I add, setting up dark/light mode in it's current state has been an absolute breeze - thank you!

**To Reproduce**
1. Follow steps outlined in https://theme-ui.com/color-modes/
2. Clear local storage and refresh
3. App should render with your system preferences
4. Try toggling between light/dark mode in your system preferences and watch the behaviour of the app.

**Expected behavior**

As I've enabled `useColorSchemeMediaQuery`, I would expect the color modes to behave as a media query (in other words, on user-agent change - much like a breakpoint).

**Screenshots**

*An example in Theme-UI: (using `useColorSchemeMediaQuery`)*

Demo site toggling between dark and light mode on an iPhone

*Examples of 'expected behaviour' in non-Theme-UI applications* 

Twitter toggling dark and light mode

DuckDuckGo

**Additional context**
Related issues: https://github.com/system-ui/theme-ui/issues/367

### LINKED ISSUE #1372: system color mode

1. I visit a website styled with Theme UI which has `useColorModeMediaQuery` set to true during the evening, and I see dark mode.
2. When I visit it again in the morning next day, there's still dark mode.
3. If I clear my local storage, the website will use light mode.
4. When I visit it later at night, it's still going to be in light mode. 
5. Pretty inconvenient, right?

1. As a user, I'd like Theme UI to respect my color scheme preference until I change the color mode.
2. As a user, I'd like to be able to go back to this state, effectively chosing "System" color mode.

**Additional context**

At this moment when no color mode value is set in local storage and `useColorSchemeMediaQuery` is set to true, `@theme-ui/color-modes` sets current preference to local storage.

This is a mutation on read, which makes the little functional programmer inside me cry.
I'd like to get rid of this, and instead if there is no color mode set, use system's color scheme.

```tsx
// initialize state
  React.useEffect(() => {
    const stored = theme.useLocalStorage !== false && storage.get()
    document.body.classList.remove('theme-ui-' + stored)
    if (!stored && theme.useColorSchemeMediaQuery) {
      const query = getMediaQuery()
      setMode(query)
      return
    }
    if (!stored || stored === mode) return
    setMode(stored)
  }, [])

  React.useEffect(() => {
    if (!mode || theme.useLocalStorage === false) return
    storage.set(mode)
  }, [mode])
```

julianbenegas:
Currently, even if `useColorSchemeMediaQuery: true`, the color mode stored in localStorage wins. Also, the color mode from the media query gets stored to localStorage.

Update (4814af8): I made the media query color mode get stored in localStorage again, to make the initial flash prevention script still work.

Fixes #787 

Absolutely **loving** theme-ui btw.

### DIFF
--- a/packages/color-modes/src/index.js
+++ b/packages/color-modes/src/index.js
@@ -65,21 +65,29 @@ const useColorModeState = (theme = {}) => {
     } else if (stored) setMode(stored)
   }, [theme])
 
-  const handleColorSchemeChange = React.useCallback(({ matches, media }) => {
-    const modeEvaluated = getEvaluatedMode(media)
-    if (matches) setMode(modeEvaluated)
+  const handleColorSchemeChange = React.useCallback((e) => {
+    const modeEvaluated = getEvaluatedMode(e.media)
+    if (e.matches) setMode(modeEvaluated)
   }, [])
 
   React.useEffect(() => {
     if (theme.useColorSchemeMediaQuery && window.matchMedia) {
-      window.matchMedia(darkQuery).addListener(handleColorSchemeChange)
-      window.matchMedia(lightQuery).addListener(handleColorSchemeChange)
+      window
+        .matchMedia(darkQuery)
+        .addEventListener('change', handleColorSchemeChange)
+      window
+        .matchMedia(lightQuery)
+        .addEventListener('change', handleColorSchemeChange)
     }
 
     return () => {
       if (theme.useColorSchemeMediaQuery && window.matchMedia) {
-        window.matchMedia(darkQuery).removeListener(handleColorSchemeChange)
-        window.matchMedia(lightQuery).removeListener(handleColorSchemeChange)
+        window
+          .matchMedia(darkQuery)
+          .removeEventListener('change', handleColorSchemeChange)
+        window
+          .matchMedia(lightQuery)
+          .removeEventListener('change', handleColorSchemeChange)
       }
     }
   }, [theme])
--- a/packages/color-modes/src/index.js
+++ b/packages/color-modes/src/index.js
@@ -44,60 +44,50 @@ const getMediaQuery = () => {
   return 'default'
 }
 
-const getEvaluatedMode = (mediaString) => {
-  if (mediaString === darkQuery) return 'dark'
-  if (mediaString === lightQuery) return 'light'
-  return 'default'
-}
-
 const useColorModeState = (theme = {}) => {
   const [mode, setMode] = React.useState(
     theme.initialColorModeName || 'default'
   )
 
+  const setPreferredColorScheme = React.useCallback(() => {
+    const query = getMediaQuery()
+    setMode(query)
+  }, [])
+
   // initialize state
   React.useEffect(() => {
     const stored = theme.useLocalStorage !== false && storage.get()
     document.body.classList.remove('theme-ui-' + stored)
-    if (theme.useColorSchemeMediaQuery) {
-      const query = getMediaQuery()
-      if (!stored || stored !== query) setMode(query)
-    } else if (stored) setMode(stored)
+    if (theme.useColorSchemeMediaQuery === true) setPreferredColorScheme()
+    else if (stored) setMode(stored)
   }, [theme])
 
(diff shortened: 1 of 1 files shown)


julianbenegas:
I also added media query listeners (a0fc915), which is a minor detail, but nice.

I used the docs to test it:

[Before](https://www.loom.com/share/f6c11f13117e4cfab7c9d10763a6ea03) — [After](https://www.loom.com/share/b3a830f840bd49108ed9fd08c7ff865e)

julianbenegas:
Thinking about an issue with this. The pr as is gives sense to the `useColorSchemeMediaQuery` key, but also makes manual toggles not work alongside it. Even if localStorage says one thing, the media query will always persist.

What if the user has `useColorSchemeMediaQuery: true`, but only wants it as a default, and also add a color mode toggle.

Example use case: The end user having three options: light, dark and auto. Very popular pattern.

- If there's a value in localStorage, **always** use that one.
- If there's no value in localStorage and `useColorSchemeMediaQuery: true`, use color scheme media query—without saving it to localStorage!
- The initializeColorMode script could be tweaked to also check for color scheme media queries so that flash is prevented either way.

To enable "auto", the programmer just has to clear localStorage.

Let me know what you guys think and I'll update.

lachlanjc:
Personally, I'd like to make this a separate option from the existing option. I think there are different usecases for these functionalities, & while personally I'd use it on most sites, I can definitely think of applications where you'd want other color modes but not use the media query live. I'm not quite sure how to name these differently:

- We could rename the existing option `useInitialColorSchemeMediaQuery`, with this staying as `useColorSchemeMediaQuery`, but that'd be a breaking change for existing users unless they change their theme config upon upgrade
- We could name this `useLiveColorSchemeMediaQuery`, keeping the existing name for the existing behavior but encouraging using `useInitialColorSchemeMediaQuery` in the future & deprecate it
- Definitely plenty of other names we could use instead, but I think one of these two structural paths make sense

hasparus:
> We could rename the existing option useInitialColorSchemeMediaQuery, with this staying as useColorSchemeMediaQuery, but that'd be a breaking change for existing users unless they change their theme config upon upgrade
We could name this useLiveColorSchemeMediaQuery, keeping the existing name for the existing behavior but encouraging using useInitialColorSchemeMediaQuery in the future & deprecate it
Definitely plenty of other names we could use instead, but I think one of these two structural paths make sense

@lachlanjc (insert state machines meme) What do you think about this?

```
useColorSchemeMediaQuery: 'live' | 'initial' | true (same as initial for backwards compatibility) | false
```

I think I'd prefer to have a special color mode `"system"` which would pick `"light"` and `"dark"` from `theme.colors`, but I'm not strongly opinionated here.

lachlanjc:
Brilliant. Yeah, let’s do the string solution, that’s way better than conflicting Booleans!

lachlanjc (changes requested):
This is awesome work. Before we merge, let's implement this as a separate option from the existing functionality, using @hasparus's suggestion for syntax:

```
useColorSchemeMediaQuery: 'live' | 'initial' | true (same as initial for backwards compatibility) | false
```

albertusdev:
hello there!
great and exciting work on this PR since I'm looking to implement a manual system/light/dark theme toggler in my Gatsby site.
Is there any news/update on this PR progress that you can share with us?
Thanks!

@hasparus @lachlanjc

lachlanjc:
@albertusdev We’d still love for this feature to be implemented, but as this PR is stale, if you’d like to update it with the API I mentioned, that would be much-appreciated!

hasparus:
:rocket: PR was released in [`v0.12.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.12.0) :rocket:

PR #982, opened 2020-06-05
custom-properties: extend ThemeProvider to insert CSS Custom Props

### LINKED ISSUE #979: Expose theme values as CSS Variables
Right now we can't use theme values when styling native HTML elements. We have to either

1. Wrap them in custom components `<H1>`, `<H2>`, etc 
2. Use the `sx` prop

I'd like to extend `ThemeProvider` with the ability to expose theme values as CSS Variables so that in a stylesheet we could do something like:

```
h1 {
  margin: var(--space-3);
  font-size: var(--font-sizes-3);
}

h2 {
  margin: var(--space-2);  
  font-size: var(--font-sizes-2);
}
```

I've already done this (just for the colors property as a proof of concept), here's the code: 

```react
import React, { useRef, useEffect } from 'react';
import { ThemeProvider, ThemeProviderProps, Theme, useThemeUI } from 'theme-ui';

export const ThemeAdapter = (props: ThemeProviderProps<Theme>) => {
  const ref = useRef(null);
  const { theme: outerTheme } = useThemeUI();

  useEffect(() => {
    const theme =
      typeof props.theme === 'function' ? props.theme(outerTheme) : props.theme;

    Object.entries(theme?.colors || {}).forEach(([key, value]) => {
      ref.current.style.setProperty(`--colors-${key}`, value);
    });
  });

  return (
    <div ref={ref} className='theme-ui-adapter'>
      <ThemeProvider {...props} />
    </div>
  );
};
```
Thus, any elements inside the `<div className="theme-ui-adapter">` can make use of the CSS Variables. Also, because of the way scope works, we have theme/sub-theme functionality by default.

iampava:
* Changed from default export to named exports
* Added `withCustomProperties` to extend `ThemeProvidr` and insert the CSS Props in DOM. 
* Updated README

More info about the reasons here:  https://github.com/system-ui/theme-ui/issues/979.


iampava:
Tests are failing. Will update them.

iampava:
Tests are faling. Checked out on master and ran `yarn install && yarn test` and still failing there.

jxnblk:
Thanks! A few quick thoughts before I review this:

1. We shouldn't remove the default export, since that would mean a breaking change.
1. Why did you decide to use a higher order component? I would guess that we could handle this as a normal React component
1. Tests aren't failing on master, so I'd guess it might be related to the changes you made in the lock file, if that helps unblock you

iampava:
Hey, thanks for the input.

1. Reverted back to it. I was leaning towards a breaking change to keep the API consistent (all named exports) but this works as well. (PS: changed the version in `package.json`. Should I do this or not?)

2. Added another feature'ish - the ability to select the `class` on the wrapper. I went with a HOC so that, if you have multiple providers inside the same app, you don't have to duplicate the prefix and className every time.

```
<CustomPropsThemeProvider theme={ subTheme } prefix="myApp" className="myApp-theme-provider" />
``` 

What do you think? It's not a huge copy-paste so I'd be fine with removing the HOC

3. Yes, you're right. The tests weren't failing, the coverage was. Wrote some tests for the new function, but couldn't find a way to test that the CSS props have been declared on the element. 

Tried all sorts of stuff, but every time calls like `div.style.getPropertyValue('--color-primary')` were empty string. I'm not too good on testing so if you've ever stumbled upon this, maybe you could provide some hints/docs.

iampava:
Up!

iampava:
Hey @jxnblk, @hasparus. Can we merge this?

atanasster (approved):
Looks great, thanks for the contribution

itsmelion (approved):

PR #983, opened 2020-06-05
Correct Tony McDade name

DebsDee:
[Tony McDade](https://www.them.us/story/tony-mcdade-police-killing-tallahassee) was a Black trans man who has been deadnamed here. Please update this list to reflect his chosen name.


johno:
Thank you for the correction.

Issue #1000, opened 2020-06-13
ts-jest test fails


chiragprodsmiths:
TypeError: React__default.createContext is not a function

      1 | import * as React from 'react';
      2 | import { render, cleanup } from '@testing-library/react';
    > 3 | import { ThemeProvider } from 'theme-ui';
        | ^
      4 | import Login from '../index';
      5 | import { theme } from '../../../theme/index';
      6 | 

      at Object.<anonymous> (node_modules/@mdx-js/react/dist/cjs.js:117:33)
      at Object.<anonymous> (node_modules/@theme-ui/mdx/dist/index.js:8:15)
      at Object.<anonymous> (node_modules/theme-ui/dist/index.js:3:11)
      at Object.<anonymous> (src/components/login/tests/login.form.spec.tsx:3:1)

chiragprodsmiths:
@jxnblk @mxstbr 

Hey Guys can anyone help me out. this has stopped me write any tests. 

using ts-jest, jest, typescript.

hasparus:
Hey @chiragprodsmiths, could you share your code with me?

chiragprodsmiths:
> Hey @chiragprodsmiths, could you share your code with me? Do you have a reproduction repo, or happen to be working on open source? I'll try to help you out.
> 
> Which version of `theme-ui` are you using?

https://github.com/chiragprodsmiths/theme-ui-ts-jest

checkout readme for spec file that has issue
https://github.com/chiragprodsmiths/theme-ui-ts-jest/blob/master/README.md

the issue only happens when i bring in theme-ui.

hasparus:
Okay, I took a short look at your repository.

What's interesting, I got a different error in your repo. I'm puzzled.
`ThemeProvider` is supposedly undefined for me.

![image](https://user-images.githubusercontent.com/15332326/84644184-1d519200-aeff-11ea-8feb-cc7f4220da6e.png)


The problem appears on `theme-ui` version **0.3.1**, and upgrading to **0.4.0-alpha.3** doesn't fix it.

In you case, I'd consider changing `import React from 'react'` in _node_modules/@theme-ui/core_ to `import * as React from 'react'`.

chiragprodsmiths:
> Okay, I took a short look at your repository.
> 
> What's interesting, I got a different error in your repo. I'm puzzled.
> `ThemeProvider` is supposedly undefined for me.
> 
[...]

PR #1002, opened 2020-06-15
Friendlier TypeScript errors.

hasparus:
## Problem

---

Hey @wKovacs64 @sbardian, would you have some time to take a look at [packages/css/src/types.ts](https://github.com/system-ui/theme-ui/pull/1002/files#diff-8ece70673d60c30b919ddc7219038dcf)?

---

**Described changes and todos in comments 👇**


packages/css/src/types.ts
```diff
@@ -524,10 +536,10 @@ export interface ColorMode {
    */
   accent?: CSS.ColorProperty
 
-  [k: string]: CSS.ColorProperty | ObjectOrArray<CSS.ColorProperty>
+  [k: string]: CSS.ColorProperty | ObjectOrArray<CSS.ColorProperty> | undefined
 }
 
-interface ColorModesScale extends ColorMode {
+type ColorModesScale = ColorMode & {
```
hasparus, line 542: `modes` doesn't match ColorMode index signature (and it shouldn't), so we can't use an interface here.

```
Property 'modes' of type '{ [k: string]: ColorMode; } | undefined' is not assignable to string index type 'string | string[] | { [K: string]: string | string[] | ...; } | undefined'.
```

One can still augment `ColorMode`, so we don't lose anything here.

packages/css/test/errors.ts
```diff
@@ -0,0 +1,52 @@
+import { expecter } from 'ts-snippet'
```
hasparus, line 1: I'm feeling kinda bad installing a library with 16 stars, but no other solution I'm aware of (e.g. tsd, dtslint, dts-jest) integrates with existing Jest setup with no additional config like ts-snippet does.

packages/css/src/types.ts
```diff
@@ -465,24 +464,37 @@ export interface VariantProperty {
...
  * such that properties that are part of the `Theme` will be transformed to
  * their corresponding values. Other valid CSS properties are also allowed.
  */
-export type ThemeUIStyleObject =
-  | ThemeUICSSProperties
-  | CSSPseudoSelectorProps
-  | CSSSelectorObject
-  | VariantProperty
-  | UseThemeFunction
+export type ThemeUIStyleObject = ThemeUICSSObject | ThemeDerivedStyles
```
hasparus, line 497: You can see I removed `CSSSelectorObject` and merged all objects into `ThemeUICSSObject`.

The naming is not perfect here.

packages/css/src/types.ts
```diff
@@ -436,11 +431,15 @@ export interface ThemeUIExtendedCSSProperties
     AliasesCSSProperties,
     OverwriteCSSProperties {}
 
+export type StylePropertyValue<T> =
```
hasparus, line 434: Any property in a style object is either ResponsiveStyleValue<T> (which includes T), a function which returns such value or a nested style object.

packages/css/src/types.ts
```diff
@@ -14,7 +9,7 @@ type StandardCSSProperties = CSS.Properties<number | string>
  *
  * For more information see: https://styled-system.com/responsive-styles
  */
-export type ResponsiveStyleValue<T> = T | Array<T | null>
+export type ResponsiveStyleValue<T> = T | Array<T | null | undefined>
```
hasparus, line 12: I needed it to make _@theme-ui/css_ strict TS, but it makes sense in general. `array[index]` can always be undefined.

packages/css/src/types.ts
```diff
@@ -1,8 +1,3 @@
-/**
- * Copied and adapted from @types/styled-system__css
- * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/028c46f833ffbbb0328a28ae6177923998fcf0cc/types/styled-system__css/index.d.ts
```
hasparus, line 3: The changes here are pretty noticeable already, I'd say. We probably shouldn't link to "the source" anymore.

packages/css/src/index.ts
```diff
@@ -215,20 +223,23 @@ const transforms = [
...
     null,
-    ...breakpoints.map(n => `@media screen and (min-width: ${n})`),
+    ...breakpoints.map((n) => `@media screen and (min-width: ${n})`),
   ]
 
-  for (const key in styles) {
-    const value =
-      typeof styles[key] === 'function' ? styles[key](theme) : styles[key]
+  for (const k in styles) {
+    const key = k as keyof typeof styles
```
hasparus, line 238: A type of key in X isn't always assignable to keys of type of X, even if we forget about the prototype. This part of TypeScript is a bit inconvenient, but it does help with some bugs.

Explanation by Anders Hejlsberg:
https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208

packages/css/src/index.ts
```diff
@@ -238,12 +249,12 @@ const responsive = (styles: Exclude<ThemeUIStyleObject, UseThemeFunction>) => (
     for (let i = 0; i < value.slice(0, mediaQueries.length).length; i++) {
       const media = mediaQueries[i]
       if (!media) {
-        next[key] = value[i]
+        next[key] = value[i] || undefined
```
hasparus, line 252: Yeah, so `value[i]` can be `null` here, and we don't allow `null` values in `ThemeUICSSObject` :/
We could, I guess... since Emotion does.

packages/css/src/index.ts
```diff
@@ -284,8 +296,8 @@ export const css = (args: ThemeUIStyleObject = {}) => (
     const transform: any = get(transforms, prop, get)
     const value = transform(scale, val, val)
 
-    if (multiples[prop]) {
-      const dirs = multiples[prop]
+    if (prop in multiples) {
```
hasparus, line 299: This line is a nice example of how sometimes common JS doesn't translate to strict TypeScript.

I'd guess the reasoning is: Since we don't know if `prop` is a key of `multiples`, we shouldn't just try to retrieve it without checking. Getters can have side effects!

packages/core/src/index.ts
```diff
@@ -72,8 +72,14 @@ const arrayMerge = (destinationArray, sourceArray, options) => sourceArray
 export const merge = (a: Theme, b: Theme): Theme =>
   deepmerge(a, b, { isMergeableObject, arrayMerge })
 
-merge.all = <T = Theme>(...args: Partial<T>[]) =>
-  deepmerge.all<T>(args, { isMergeableObject, arrayMerge })
+function mergeAll<A, B>(a: A, B: B): A & B
+function mergeAll<A, B, C>(a: A, B: B, c: C): A & B & C
+function mergeAll<A, B, C, D>(a: A, B: B, c: C, d: D): A & B & C & D
+function mergeAll<T = Theme>(...args: Partial<T>[]) {
```
hasparus, line 78: I added a few overloads to infer the type without explicit annotation.

package.json
```diff
@@ -16,38 +16,41 @@
...
+    "ts-snippet": "^4.2.0",
+    "typescript": "^3.9.5"
   },
   "resolutions": {},
   "jest": {
     "preset": "ts-jest/presets/js-with-babel",
     "globals": {
       "ts-jest": {
         "tsConfig": {
+          "target": "es2016",
```
hasparus, line 53: I bumped Babel and Jest deps to use optional chaining in tests.

- [ ] The target here could be increased to `es2018` I guess.

packages/css/src/types.ts
```diff
@@ -465,24 +464,37 @@ export interface VariantProperty {
    *
    * @see https://styled-system.com/variants
    */
-  variant: string
+  variant?: string
 }
 
-export interface UseThemeFunction {
-  (theme: any): Exclude<ThemeUIStyleObject, UseThemeFunction>
+export interface ThemeDerivedStyles {
```
hasparus, line 470: "use" is already a loaded word in React world. Tell me what you think about the rename.
Another name I considered is `ComputedStyle`.

hasparus: Emotion calls it `FunctionInterpolation`, but since we don't do CSS syntax in tagged template literals here, I'm not sure it would makes sense.

packages/css/src/types.ts
```diff
@@ -465,24 +464,37 @@ export interface VariantProperty {
...
+  (theme: Theme): Exclude<ThemeUIStyleObject, ThemeDerivedStyles>
```
hasparus, line 471: I added `Theme` type to the argument here.

This is a breaking change for TypeScript ppl who added extra properties to their theme object and use them in functions.

- [ ] I'll draft a TypeScript section in the docs, and describe how to add additional properties to the Theme type.

packages/core/src/index.ts
```diff
@@ -3,9 +3,8 @@ import {
   ThemeContext as EmotionContext,
   InterpolationWithTheme,
 } from '@emotion/core'
-// @ts-ignore
 import { css, Theme } from '@theme-ui/css'
-import React from 'react'
+import * as React from 'react'
```
hasparus, line 7: See https://github.com/facebook/react/pull/18102

It _might_ help with https://github.com/system-ui/theme-ui/issues/1000. I didn't test it.

hasparus: We should probably consider using star import for React everywhere.

packages/css/test/errors.ts
wKovacs64: Eh, the author looks to be very active on GitHub - probably safe (and certainly convenient).

packages/css/src/types.ts
wKovacs64: I hate this (being too permissive) but you're absolutely right that it should aim to be least confusing in the event of an error. I *do* feel it should enforce correctness but unless we can have both... errors that make sense are probably more valuable than forbidding properties that will just be ignored anyway. 🤷‍♂️

packages/css/src/types.ts
wKovacs64: Agreed.

sbardian:
I didn't see where `CSSSelectorObject` [here](https://github.com/system-ui/theme-ui/blob/120192fa606b61048559e8317ee8ad3425344ddc/packages/css/src/types.ts#L49) was being used.  Same here though, you are also better at TS than me, tests pass and look good. :)

packages/css/src/types.ts
hasparus: I noticed later today, that this isn't too permissive.
There are variables in CSS 🤦
Arbitrary property names (we can't typecheck the leading dash) with values are 100% legit and useful CSS.

wKovacs64: Oh, good point. Then I guess we're set! 😄

packages/css/src/types.ts
hasparus: > I have no idea how common of a use case that is

It actually affects half of my projects using Theme UI. 😢

This is also breaking `(theme: MyExactTheme) => ...` usage, where components depend on exact shape of the theme. This makes sense a lot of sense for blogs, but it's a bug in apps where the theme can be customized by the user.

I have mixed feelings about this change.
Possibly, we could add a generic parameter and describe derived styles as `<T extends Theme = Theme>(theme: T) => ThemeUICSSObject` to allow this 🤔

```
sx={{
  bg: (t: MyTheme) => colorMode === 'dark'
    ? transparentize(t.colors.modes.dark.primary, 0.2)
    : darken(t.styles.form.backgroundColor, 0.6)
}}
```

hasparus:
> I didn't see where CSSSelectorObject here was being used.

Thanks! I forgot to remove it 🤦

>  you're way better at TS
>  you are also better at TS

You gentlemen, are highly misinformed. I have no idea what I'm doing.

packages/css/src/types.ts
wKovacs64: > Possibly, we could add a generic parameter and describe derived styles as `<T extends Theme = Theme>(theme: T) => ThemeUICSSObject` to allow this

That sounds pretty good...?

hasparus: > That sounds pretty good...?

Too good :/ It doesn't work.

wKovacs64: Of course. 🤦‍♂️

hasparus:
Note: Exact themes

[TS Playground](https://www.typescriptlang.org/v2/en/play?#code/PTAEBsEsCMCcENYE8BQKAmBTAxuRnQBbAe3QFdwCByAFwAtNDMBaMyK0AbxVFEwA8ADsVg1QkAHY1MsAGbxsBALJIAKgyYB5AG4zYkLF1ABfNLwHDRoGkkHK1GggF5QK9Y0w69BggOkT0AGcjeg8ALnEJWRlQVRNQAH5Y0AjOYwBuMz4hETFsYglAsSx9XXQAZRtKQIiAClk60KYI9yZQADJXBw8ASlAnAD5QMgkAawliAHcJPsHQKgmaZkhCQUomKUx0KhRTFBXLMU5Yx3jZWGJCeaaWNipMlBBhwJlAwQVMNHzCsUJ4UcwrWcoAAPHE-JgAsEgQNajQWrMhjQHt8itZTi4-gCgbVuLx8uARDUuDxeKBoApRgBzC4jdARKiTOiQaRUAA0pN40n48Pm0Dw2FG7M55PAZEwqRFvCgVLovKoAGIAIwADgA7PBoLJhWSyehEKMGQqAEwAZgArPAAJxqna60y8UzGHpoGx2UAAUX4ChoQP61lsmGIsnRHgeWFw+CIpAo1BurHYJPMOSskmkcg+XSBXlg+kMeLJNwiXp9QNJTrQB1yRhKkDKlSQ1TOFyutEcCfuaCeTMhoCQxDIsFDbUgwUWoHQSAk8EIkGwbNA4ymlHQVIImoHYhZwQEPtAgTo8Hdo-xBV0EkgkJoGBkda2DeqcP6SIAdASiS-+eKXzK5T0Hk8u6rJQXwFGiv40AAQmKmBKv6NxvsQhKwIEn4wT+kCysioE-KAa4SLe2BAi0GLDpgAFgIQFA0JAay+LmRIoKiYgQdB4oegxQ4uPhhFAohyGoV+mAYVhaBoEAA)

```ts
// library

declare module 'theme-ui' {
  export interface MyThemeOverride { }

  export type MyTheme = MyThemeOverride extends { theme: infer T } ? T : {};

  export const derivedStyles: (f: (theme: Theme & MyTheme) => unknown) => 'not-implemented'
}

import { Theme } from 'theme-ui';

// userspace

const makeTheme = <T extends Theme>(t: T) => t;

const theme = makeTheme({
  colors: {
    background: 'white',
    text: 'black',
    blue: {
      light: '#187abf',
      dark: '#235a97'
    }
  }
})

type ExactTheme = typeof theme;

declare module 'theme-ui' {
  export interface MyThemeOverride {
    theme: ExactTheme
  }
}

import { derivedStyles } from 'theme-ui';

// when your theme is not dynamic, knowledge about its exact shape is  convenient
derivedStyles(t => t.colors.blue.light);

// example

const lightBlue1 = theme.colors.blue.light;

const genericTheme: Theme = theme;

// multiple errors
const lightBlueError = genericTheme.colors.blue.light

```

hasparus:
Okay, this PR is probably finished if the page I added to the docs makes sense (@jxnblk)

The TS chapter for the docs is here: https://deploy-preview-1002--theme-ui.netlify.app/guides/typescript/

I'm not sure if it should be in guides or below Migrating.
If it stays in the guides, I'll add a link to it to the Guides page.

hasparus:
Thanks!

Based on a sample of size two of my own projects (serious statistics), I used exact theme type 3 times.
1. While passing colors to zdog's imperative API.
2. In `variant` prop of custom Link component.
3. In a Stack component, because I couldn't index `theme.space` with number (I changed `ObjectOrArray` in this PR, it should be okay right now.)

And I found `get` or optional chaining used few dozen times.

**What's important, I broke following scenario**

```tsx
css({ color: (theme: ExactTheme) => theme.colors.blue[300] })
```
use case in this PR. It worked previously, because `theme` here was typed as `any`. There is a small trade-off.


&nbsp;| `(theme: any) => somevalue`  | `(theme: Theme) => somevalue`
-|-|-
JS user | 😢 no code-completion | 😄 completion of scale names
TS user | 😢 forced explicit annotation <br/>😄 easy to annotate with exact theme | 😄 no need for annotation <br/>😄 interoperable by default <br/>😢 exact theme requires assertion (`as ExactTheme`)

There's a possible fix [few comments above](https://github.com/system-ui/theme-ui/pull/1002#issuecomment-648025013).

I'll move ExactTheme example to the top and add this. This should be pretty useful.

```
import { ContextValue } from 'theme-ui';

interface ThemeCtx extends Omit<ContextValue, "theme"> {
  theme: ExactTheme;
}

export const useTheme = (useThemeUI as unknown) as () => ThemeCtx;
```

packages/docs/src/pages/guides/typescript.mdx
```diff
@@ -0,0 +1,124 @@
+---
+title: TypeScript
+---
+
+# TypeScript
+
+Theme UI is written in [TypeScript](https://www.typescriptlang.org/) to provide
+better IDE support for both JS and TS users.
+
+Most of the public API should work seamlessly exactly as in JavaScript, but
```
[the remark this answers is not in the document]
jxnblk: I'm guessing that this line is meant to explain what is unique about using Theme UI with TS, right? Would something like the following make sense?

> When using Theme UI in a TypeScript project, most APIs are identical to their usage in JavaScript, but some advanced use cases require a different approach. This guide is intended to cover those advanced use cases.

Or, maybe:

> While most APIs in Theme UI should *just work* in TypeScript, there are a few advanced use cases which will differ slightly. This guide is intended to cover those use cases.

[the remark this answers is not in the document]
hasparus: Yes. "This guide is intended to cover those use cases." is what I wanted to say here. Thanks! Let me update it real quick.

packages/docs/src/pages/guides/typescript.mdx
```diff
@@ -0,0 +1,124 @@
...
+  },
+})
+
+export type ExactTheme = typeof theme
+
+// No error
+const lightBlue = theme.colors.blue.light
+```
+
+You can then reexport `useThemeUI` hook with narrowed type.
```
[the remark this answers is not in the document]
hasparus: Not exactly sure what if I understand you here, but _yes?_: Theme UI types don't know about user's theme exact shape. Everything will just behave like we're writing a bunch of reusable components for any theme.

This can be improved, I think. We could _"tell Theme UI about our theme"_ like in Overmind: https://overmindjs.org/core/typescript#1-declare-module. I only have a proof of concept of this code in one of the comments here.

packages/docs/src/pages/guides/typescript.mdx
```diff
@@ -0,0 +1,124 @@
+---
+title: TypeScript
+---
+
+# TypeScript
+
+Theme UI is written in [TypeScript](https://www.typescriptlang.org/) to provide
+better IDE support for both JS and TS users.
```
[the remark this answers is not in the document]
hasparus: I wanted to mention TypeScript, to convey that the typings are, and will be, always up to date.

Typechecking sx props is surely useful for refactoring and maintenance, though I think I'm focused on "creating a contraint-based styling DSL" i.e. text editor support throught TypeScript language server
 
How would you rephrase it? Or maybe should I just scrap this paragraph?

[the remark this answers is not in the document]
jxnblk: Not sure I have a super strong opinion on which way to go, but I'd say either:

1. `Theme UI is written in TypeScript.` (just state it as a matter of fact)
or 2. list out about three benefits together (can include IDE benefits)

packages/docs/src/pages/guides/typescript.mdx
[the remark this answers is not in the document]
jxnblk: Was mostly for my own clarification, and I made a typo, but it sounds like the `ExactTheme` type is *only* needed when you're using the custom `useThemeUI` hook, right? I.e. I wouldn't use that type anywhere else?

> This can be improved, I think. 

Yeah, totally -- I understand that is outside the scope of this PR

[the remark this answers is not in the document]
hasparus: > it sounds like the ExactTheme type is only needed when you're using the custom useThemeUI hook, right? I.e. I wouldn't use that type anywhere else?

ATM that's correct.
It's useful if you want to write something just for one specific theme.

I just made https://github.com/system-ui/theme-ui/pull/1090 so we can "turn on" using this type __everywhere else__.

Issue #1006, opened 2020-06-15
0.4.0-highlight.0 release messes with caret range version resolution


taion:
**Describe the bug**
By lexicographic sorting, `^` and `>=` ranges on the `0.4.0` alphas resolves to `0.4.0-highlight.0` rather than the more recent (and likely intended) `0.4.0-alpha.3`. This happens both for external things that depend on the alphas, and internally within the monorepo as well.

**To Reproduce**
Run `yarn add gatsby`.

**Expected behavior**
These ranges resolve to `0.4.0-alpha.3`, the actual latest alpha.

**Screenshots**
From my `yarn.lock`:

```
"theme-ui@>= 0.4.0-alpha.0", theme-ui@^0.4.0-alpha.3:
  version "0.4.0-highlight.0"
```

```
"@theme-ui/css@>= 0.4.0-alpha.0", "@theme-ui/css@^0.4.0-highlight.0":
  version "0.4.0-highlight.0"
```

**Additional context**
I noticed this because `0.4.0-highlight.0` exports some really odd types under `@theme-ui/css` that make building unrelated things fail due to `CSSSelectorObject` and `VariantProperty` being private, which has since been fixed in alpha.3.

atanasster:
Encountering the same issue

taion:
Can we do `0.4.1-alpha.0`? Or else bump the prerelease tag to `rc`.

jxnblk:
Packages should now be published as `0.4.0-rc.1` with the `@next` tag. Please feel free to reopen this issue if the problem persists or you run into any other issues

PR #1007, opened 2020-06-15
Change version to 0.4.0-rc.0

hasparus:
Hey @taion, @atanasster :wave:

This should solve #1006, am I right?


taion:
this should, yes, because `rc.0` is after `highlight.0` lexicographically, so `^0.4.0-rc.0` won't match `highlight.0`

atanasster:
@hasparus - fantastic, thanks so much.

Another option was to release an official version (not sure which is the next official after current 0.3.1) as we discussed in another thread.

jxnblk:
Thanks for this and sorry to anyone affected by this! I might have to do a little dance with Lerna to get this published correctly, but will post another comment in the linked issue once it's up

PR #1012, opened 2020-06-17
fix: add missing comma in color code example

guayom:
In the @theme-ui/color README, the advanced usage example is missing a comma so the css rendered is not valid


jxnblk:
Thanks!

PR #1035, opened 2020-07-02
(@theme-ui/components): Add switch component

jordanoverbye:
This PR adds a `Switch` component to the `@theme-ui/components` package

Preview: https://deploy-preview-1035--theme-ui.netlify.app/components/switch


packages/components/src/Switch.js
```diff
@@ -0,0 +1,74 @@
...
+            opacity: 0,
+            zIndex: -1,
+            width: 1,
+            height: 1,
+            overflow: 'hidden',
+          }}
+        />
+        <Box
+          css={{
+            padding: GUTTER,
```
jordanoverbye, line 28: Adding this to the `css` prop otherwise it will interpolated with a theme value

jxnblk:
Nice! I *definitely* want to get a quick a11y audit on this before merging but looks great!

laurieontech:
Thanks for the PR! This is accessible by keyboard which is great. However, there is no aria attribute for screenreaders to tell the different state toggles. We should add that before merging.

jordanoverbye:
Thanks for the feedback @laurieontech! I've just updated my PR. I'm a little unsure if what I did is correct or not, so please let me know if you need me to update anything else

laurieontech:
So the goal here is that there is a label for each state rather than the whole component. [React switch](https://github.com/markusenglund/react-switch#readme) has a good example of this.

cobraz:
The switch component would be the last missing piece for our kit (tabetalt/kit). Anything I can help with?

jordanoverbye:
Thanks for your message @laurieontech, but sorry I still don't know what you mean. I can't seem to find any examples on that website where the `label` itself is actually changing. Which example in the React Switch docs are you referring to? If its the first one, this can be achieved by something like this:


```jsx
function ExampleComponent() {
  const [checked, setChecked] = useState(false);

  const onChange = (e) => setChecked(e.target.checked);

  return (
    <div>
      <Label>
        <Switch onChange={onChange} checked={checked} />
        Enable email alerts
      </Label>
      <p>{checked ? 'The switch is on.' : 'The switch is off.'}</p>
    </div>
  )
}
```

As for the aria-labels, since we are spreading all of the props onto the input - the consumers of this component can add whatever aria attributes they need.

jordanoverbye:
@cobraz feel free to comment on this PR with any code suggestions

cobraz:
@jordanoverbye Thank you for this pull request. We're not in a hurry, just wanted to make that clear :) You are not holding us up at all! We are just super happy that this could be a part of the library, that way we don't have to implement it. My time is filled with a few other projects right now, but I can look into it later!

cobraz:
@laurieontech what exactly is needed for this to be merged? I'm considering if we need to create our own Switch component for [tabetalt/kit](https://github.com/tabetalt/kit) or not. :)

brambow:
I think what @laurieontech is referring to is React Switch's default UI that has icons inside the switch that change with the state; However, she refers to multiple `<labels>` which is not actually how React Switch does it (they use two props: `checkedIcon` and `uncheckedIcon` which are rendered into divs in the Switch background, not `<label>` elements). See the examples here: https://react-switch.netlify.app/ which show the entire switch wrapped in just a single `<label>`, just like with this example.

I'd disagree that a Switch component should have icon indicators like React Switch does as a default. A toggle switch is just styling for a checkbox, which only has a binary state of checked or not. It does not have two open-ended states that need to be displaying in an indicator. The switch is either on or off.

Really looking forward to seeing this component added to ThemeUI. I'm already using the code for it in my [@cartolab/elements](https://github.com/cartolab-gis/elements) library, but would prefer to see a Switch in ThemeUI, just like it already exists in Rebass.

atanasster:
@brambow , I believe this might be `aria-pressed="true"/"false"` = meaning different labels for the toggle state?  

Your elements library looks great, hope I can have a project to use in soon :)
Any reason you are on theme-ui 0.3.1 ? Do you have any issues with the 0.4 betas, as I believe @hasparus is ready to publish a 0.4 final.

brambow:
@atanasster It might be, but I don't see that aria attribute used anywhere in React Switch's source code either. I mostly commented in hope that it brought attention back to this PR! I am working on some changes that will push Elements to v2.0 and plan on evaluating 0.4.0 of theme ui in a few weeks as part of that. Just haven't had time yet.

atanasster:
@brambow thanks for the feedback - indeed I don't see it on react-switch either. 
I think we can add the aria-pressed and unless @laurieontech has more feedback, this PR cold be merged after 0.4 is released, possibly for a "1.0 alpha".

@hasparus wdyt?

hasparus:
@atanasster IMO we can merge it after updating changelog, but we still can't do a full Lerna release.

atanasster:
@hasparus - where is the changelog?

PR #1041, opened 2020-07-05
(@theme-ui/components): Pass Checkbox children to wrapping container

hasparus:
Hello 👋

**Currently**, one can pass children to `Checkbox`, which subsequently are passed to `input` and trigger this lovely error.

> Error: input is a void element tag and must neither have `children` nor use `dangerouslySetInnerHTML`.

[It looks pretty funny if you do it in the docs.](https://user-images.githubusercontent.com/15332326/86528568-94989700-bea9-11ea-80de-7c81eda808d7.png)

**After this PR**, the children are given to the wrapping Box.
What's cool, on top of avoiding error, it now allows this:

repeatedly toggling checkbox

---

Here's the example source if you wanna try it in the deploy preview.

```tsx
<Label sx={{ '> div': { display: 'flex', alignItems: 'center' } }}>
  <Checkbox>
    <span
      sx={{
        ':before': {
          content: '"No"',
        },
        'input:checked ~ &:before': {
          content: '"Yes"',
        },
      }}
    />
  </Checkbox>
</Label>
```

### DIFF
--- a/packages/components/src/Checkbox.js
+++ b/packages/components/src/Checkbox.js
@@ -2,19 +2,19 @@ import React from 'react'
 import Box from './Box'
 import SVG from './SVG'
 
-const CheckboxChecked = props => (
+const CheckboxChecked = (props) => (
   <SVG {...props}>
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks, good catch! I think the code change makes sense, but I'm fairly certain the code example in the description here using pseudoelements probably isn't very accessible; either we need to double check that or avoid documenting that use-case

PR #1072, opened 2020-07-18
Add typecheck scripts

hasparus:
:wave:

I added a _typecheck_ step as mentioned by @blainekasten in https://github.com/system-ui/theme-ui/issues/668#issuecomment-644978640.

It feels a bit wasteful to run this on every change, and it could be probably made faster with some TS `composite` projects and GitHub actions wizardry, but I guess it's not bad for now.

The .d.ts for for _@theme-ui/color-modes_ was needed to type values reexported from _theme-ui_.
This should make Theme UI TypeScript feel less "early access" for users.

---

BTW I tried using `lerna exec` instead of `lerna run` to leave the package.json files in packages intact, but it resulted in incomprehensible bash one-liner 😅

### DIFF
--- a/packages/gatsby-theme-ui-layout/package.json
+++ b/packages/gatsby-theme-ui-layout/package.json
@@ -6,7 +6,8 @@
   "peerDependencies": {
     "gatsby": "^2.13.1",
     "react": "^16.8.0",
-    "react-dom": "^16.8.0"
+    "react-dom": "^16.8.0",
+    "typecheck": "tsc --noEmit"
   },
   "license": "MIT"
 }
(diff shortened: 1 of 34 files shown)


packages/gatsby-theme-ui-layout/package.json
```diff
@@ -6,7 +6,8 @@
   "peerDependencies": {
     "gatsby": "^2.13.1",
     "react": "^16.8.0",
-    "react-dom": "^16.8.0"
+    "react-dom": "^16.8.0",
+    "typecheck": "tsc --noEmit"
```
LekoArts, line 10: Slipped into the wrong line here :)

hasparus: oh wow :D what have I done...

PR #1073, opened 2020-07-19
Update forms.mdx

BhanuKorthiwada:
rows property can accept number only


jxnblk:
Thanks for the PR. Does this only affect TypeScript usage? If so, I think the type definitions should be updated to support the native HTML attribute

hasparus:
@jxnblk


ATM the types for Textarea are transparent. There is a props spread, so we just take native `React.ComponentProps<'textarea'>` props and add `BoxProps` to them.

lachlanjc:
@hasparus merge?

PR #1090, opened 2020-07-26
(css): Allow injecting UserTheme and typecheck colors against scale

### LINKED ISSUE #1228: Support custom theme types in TS bindings
When using the `sx` prop, the `useThemeUI` hook, or any other API that exposes the theme object, it would be really nice to have a way to set a custom theme type for that object on a project-wide basis. So when I write

```tsx
<div sx={{ backgroundImage: ({ colors: { red, blue } }) => `linear-gradient(to right, red, blue)` }} />
```

`tsc` shouldn't throw an error, which it currently does because [the default `Theme` type's properties are all optional](https://github.com/system-ui/theme-ui/blob/v0.4.0-rc.5/packages/css/src/types.ts#L587-L787) (so the compiler has no way of knowing whether `colors` exists on the theme object).

We should somehow be able to use generics to solve this problem:

1. Leave out the ambient type that extends JSX elements with the `sx` prop. Explain how to extend the JSX interface with your own theme type in the docs. It should look something like this (whereas `SxProps` [isn't currently generic](https://github.com/system-ui/theme-ui/blob/v0.4.0-rc.5/packages/core/src/types.ts#L10-L23)):

```ts
import { SxProps } from '@theme-ui/core';

declare module 'react' {
  interface DOMAttributes<T> extends SxProps<MyTheme> {}
}
declare global {
  namespace JSX {
    interface IntrinsicAttributes extends SxProps<MyTheme> {}
  }
}
```

2. Rewrite the `useThemeUI` hook in a way that takes advantage of a generic type and explain how to use this feature in the docs. Something like:

```ts
import { useThemeUI } from 'theme-ui';

export const useTheme = useThemeUI<MyTheme>;
```

3. Repeat step 2 for all the other functions/API's in the Theme UI ecosystem that are currently hardwired to use the non-generic `Theme` type. For example, [`@theme-ui/color`](https://github.com/system-ui/theme-ui/blob/v0.4.0-rc.5/packages/color/src/index.ts#L7-L10).

I developed a brute-force workaround by disabling Theme UI's types in `tsconfig.json` and [manually redefining the ones that define the theme object and related API's](https://github.com/system-ui/theme-ui/issues/1188#issuecomment-719016511).

**Additional context**
Original issue described here: https://github.com/system-ui/theme-ui/issues/1188#issuecomment-719016511

### LINKED ISSUE #1297: [typescript] Sx properties should accept booleans
It's very important that the sx properties are typed so that you can write `sx={{ color: isActive && 'blue' }}`. It does behaves correctly when a boolean is passed, the property is simply ignored, it's only a typescript since it expects only strings.

### LINKED ISSUE #2006: Make theme values passed to css and sx properties typesafe
I'd like to find a way to make the value I pass to variant (and other theme aware properties) type safe and have it provide autocompleted values from my theme. Currently I can pass any string to variant and it will fail silently since variant accepts any string.

``<SomeComponent sx={{ variant: 'accepts any string and produces no error' }} />``

I'd like to able to pass a string that corresponds to what exists in my theme file and have errors produced if I pass a string that doesn't.

**Additional Context**
[Would this solve my issue?](https://github.com/system-ui/theme-ui/pull/1090)

hasparus:
This is IMHO the coolest PR I've sent to this repo. An actual feature!

A move towards *"TypeScript: Type check property values against theme scales"* **v1** goal. https://github.com/system-ui/theme-ui/issues/832

I belive we can treat Theme UI as an embedded domain-specific language for branded styling.

## Reasoning

Imagine you're working with _strict designers_. They specify what's _on-brand_ and what's not, the list of colors, spacings etc.

We keep these design tokens in our theme scales, but we can still use arbitrary values, merge them to master, deploy and get a dissapointed slack message from a designer. What if we could constain ourselves to predefined values on type level to make sure we don't do this?

Now we can.

I kinda like the idea of the runtime working as it is now and supporting arbitrary values all the time.
They are useful for prototyping, even if we don't want to merge them to master.

If we decide to not build the app on type errors, we effectively [make illegal states unrepresentable](https://fsharpforfunandprofit.com/posts/designing-with-types-making-illegal-states-unrepresentable/). 

## WIP

Any help is welcome! 🙌

- [ ] Benchmark TypeScript compilation times in projects using this branch in comparison with current develop.
- [x] I'd like to refactor _css/src/types_ a bit because it turned 800 lines of circular references. Maybe we can colocate types with the runtime code?
  - [ ] We need to import more types from Emotion, get rid of the ones that were copied from DefinitelyTyped.
  - Note: `Theme` depends on variants and scales. Variants depend on ThemeUIStyleObject. ThemeUIStyleObject depends on `Theme` and "platform" CSS types.
- [ ] Describe it in the docs.
- [x] Tests!

---

@nandorojo, will this work with [dripsy](https://github.com/nandorojo/dripsy)? I'd guess so, but I only skimmed through your codebase.

---

image

---

The solution is similar to

DefaultTheme injection in `styled-components`:
https://styled-components.com/docs/api#create-a-declarations-file
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/index.d.ts#L412

And declaring config in Overmind:
https://overmindjs.org/core/typescript#1-declare-module.

---

Closes https://github.com/system-ui/theme-ui/issues/1297.


nandorojo:
Ohh love this. If I understand correctly, I think it would close [this dripsy issue](https://github.com/nandorojo/dripsy/issues/6)? I've been thinking about the right way to do typescript intellisense a lot. This PR seems really elegant, I would definitely implement it.

hasparus:
Oh! I was looking for that gif!



You previously had it in the readme, am I right?

I think merging this (okay, finishing and merging, there's still some work ahead) closes the issue in Dripsy and the only thing left to do there would be an update of Theme UI version.

https://github.com/nandorojo/dripsy/blob/master/src/css/types.ts#L1-6

We don't need any codegen, the user will just opt-in to _strict mode_ (_forced design tokens mode_?) with declaration merging.

nandorojo:
Yup I put it in the readme when I was first dreaming up dripsy and didn't want to forget to work on. I can't wait to add it back on as a feature, not an idea. Super excited.

hasparus:
Okay, I've noticed a limitation.

Without codegen we can't concatenate strings on type level. What does it mean?

We can't statically know that
```ts
{
  colors: {
    gray: ["#eee", "#bbb", "#333", "#111"]
  }
}
```

results in
```
gray.0
gray.1
gray.2
gray.3
```

so nested scales would need codegen.

We could probably do a hack like 

```ts
type Color = keyof UserTheme['color'] | (string & {})
```

to allow arbitrary values.
Tbh I'd personally be fine with flat scales and writing `gray-1`.

nandorojo:
Does the new TypeScript 4.x tuple approach solve that?

https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/

(I don't know the answer.)

I typically use flat scales anyway, so I'd also be fine with that. I would prefer type consistency than nesting values.

nandorojo:
@hasparus Hey, just checking to see if you're still planning on working on this. Really excited for it.

Relevant from Material UI: https://material-ui.com/components/about-the-lab/#typescript

hasparus:
I'm wondering if we should migrate to `csstype` v3.
https://github.com/frenic/csstype/blob/master/index.d.ts#L18131

It uses `... | (string | {})` in every property that has `... | string` in v2, so we could just use the types from there and add the keys from exact theme to them.

> @hasparus Hey, just checking to see if you're still planning on working on this. Really excited for it.

@nandorojo 

Well, I am, but **slowly** :( Would you like to take over?

nandorojo:
@cmaycumber any shot you've done declaration merging before? I'm going to work off this branch to try to solve https://github.com/nandorojo/dripsy/issues/6, but I haven't done it before.

@hasparus would this PR get merged if we made progress on it?

cmaycumber:
@nandorojo No, I have not. But this feature is super exciting and I'd love to see it in dripsy and theme-ui, I think it would be a great addition and would be definitely relevant in my apps/websites. 

On that note, I'd definitely help as much as possible. I'll try to create a fork of this branch and work with it a bit to see if I can help.

nandorojo:
Awesome, that's great to hear. I'm going to look into @hasparus work on this so far and try to do the same!

nandorojo:
@hasparus is there a repository you use somewhere that has the JSDoc comments for the CSS properties? For example, where did you get this from:

```ts
export interface ColorScaleCSSProperties {
  /**
   * The **`color`** CSS property sets the foreground color value of an element's text and text decorations, and sets the `currentcolor` value. `currentcolor` may be used as an indirect value on _other_ properties and is the default for other color properties, such as `border-color`.
   *
   * **Syntax**: `<color>`
   *
   * **Initial value**: Varies from one browser to another
   *
   * | Chrome | Firefox | Safari |  Edge  |  IE   |
   * | :----: | :-----: | :----: | :----: | :---: |
   * | **1**  |  **1**  | **1**  | **12** | **3** |
   *
   * @see https://developer.mozilla.org/docs/Web/CSS/color
   */
  color?: Color
...
}
```

I don't see it in this format on mozilla's site. Thanks!

hasparus:
OMG you guys are picking this up! I'm so grateful. Sorry for slow responses :c

_BTW good news_: [TypeScript 4.1](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types) makes strictly typing nested scales possible.

> @hasparus is there a repository you use somewhere that has the JSDoc comments for the CSS properties? For example, where did you get this from...

All of these should already in Theme UI codebase. I think it's originally from CSSType, but there's no way to "inherit" these comments, so we need to copy them if we want them to appear in user tooltips.

nandorojo:
@hasparus no worries, thanks for the response. It might make sense for one of us to move this into our own PR if you don’t plan on working on it for now.

Would you mind telling me if this is going in the right direction? https://github.com/hasparus/theme-ui/pull/1

Thanks so much for this! Really excited for this feature.

hasparus:
@nandorojo the Stitches PR I mentioned in Twitter DMs.

https://github.com/modulz/stitches/pull/206

Our code and tests will end up looking pretty similar I suppose.

nandorojo:
Ah, the GIF in that PR is so satisfying. Thanks for passing this along.

packages/css/src/scales.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  borderTopColor: 'colors',
+  borderBottomColor: 'colors',
+  borderLeftColor: 'colors',
+  borderRightColor: 'colors',
+
+  outlineColor: 'colors',
+} as const
+
+export type Color = Exclude<ScaleProperty<FinalTheme['colors']>, 'modes'>
```
nandorojo: What if we still included keys from the different color modes?

We could do something like [this](https://www.typescriptlang.org/play?ts=4.1.0-dev.20201005#code/MYewdgzgLgBKA2IBOEYF4YG8BQM8wAckBLAWwEMkBPALhgHIBzJAUxbHoBpd9SQATFhDo58YmP0oBrET3HiiZSrQZUW8RAHcuc+fgD0+mJoAW5WMQBmMcjD6CYZ1LYTIYUMxYgdY5AEYgAG4sAPy6eu4sAB5QdPQExGBS9OH4AL66GWk2qKCQUNjYUFQELDAAwiCISOjuJSwg1q4ohcWlMABC5BAsldWoGACiUcDwAK6CADxSLFSNFVXInAz2QvQAfK31MACyAr2LKLUzc9Z9yADa9KsQ9AC6Fyfz50hXN-d3W+0AYonk8C8Bp1ugd+jAAD67faAwp5aAwSx-eB0X5gf6Ai53WpXRQUahcBhQaJQe5AA).

Screenshot: https://i.imgur.com/L8BU0M5.png

hasparus: Isn't this a problem, though? If we use `"secondary"` in default color mode, we'll have `color: undefined` somewhere in our styles.

I don't have a strong opinion here. I always declare my colors like this anyway TBH

```tsx
const light = { ... };
type ColorMode = typeof light;
const dark: ColorMode = { ... };
const other: ColorMode = { ... };

const initialColorMode = "light";
const modes = { light, dark, other };

const colors = {
  ...modes[initialColorMode],
  modes: omit(modes, initialColorMode)
}
```

nandorojo: I also don’t have a strong opinion on it. That’s probably the right way to do it, so we can stick without it.

samburgers:
good stuff everyone. so excited for this!

nandorojo:
I should be able to finish the scales this weekend.

nandorojo:
@hasparus The scales are all complete. Mind looking over them? 

TSDoc comments aren't added yet.

hasparus:
> @hasparus The scales are all complete. Mind looking over them?

Getting right to it today / tomorrow morning.

hasparus:
I'm annotating the scales with `Record<keyof ScaleProperties, 'scaleName'>` to ensure we don't miss anything. 

```tsx
const sizes: Record<keyof SizesCSSProperties, 'sizes'> = {}
```

nandorojo:
Are we good to check those off in your original post?

hasparus:
yup, I think so.

I’m pretty excited about how it's starting to look.

I'd also like to get deep paths in scales to make the "strict mode" compatible with all themes.

I'm a bit worried that we need more tests and better infra (like preview releases on CI) to start adding features like this smoothly.

nandorojo:
> I'd also like to get deep paths in scales to make the "strict mode" compatible with all themes.

Agreed. I think something like this would be good:

Screen Shot 2020-10-16 at 4 47 42 PM

Here's a [video](https://www.loom.com/share/6b60d24725b44d01b3feb6298a616537?from_recorder=1) of it.

Code sample [here](https://www.typescriptlang.org/play?ts=4.1.0-dev.20200921#code/C4TwDgpgBACghsAFgSQLZgDYB4AqAaKAaSggA9gIA7AEwGcoBrCEAewDMocA+KAXgCgoREuSp0otYACcAlpQDmgqAH5OAbUIBdERRr0AShADGLKdSyTZCgnEoguSoapwbtZXeICCUqXBAAZGSYsW3tHIRVhAB8oAAMAEgBvQgBfADok+CQ0TFxXAgBRUiMMAFdqCCwmVg4XLQJq9ihQtU0uLhTY8KEALmi4pNSMxKyUdGw6zQbmJsmOroioPsIlPsoIADcIKQBufn5QSFgERFweXmPs8dxpms4eGMbavYPwaFGANTgyyvxYHTE9FGZ3OShgAL0A0ScjY2yI6SSMLhhkknSUqmI7kBjBmtXCqhRwAh4mBc26kU+31Kv3yUEJDkWvSg6y2UnCa0221W-yxkKenHx6hgmnZzM5u32+wqJTgUmgbFKlCMwBkLEoUHkEGANx5okhpPaAAoWAAjABWfT+YBOfRgAEpbScvj8dTAuC9+CZKJIoEgIKhoBdEkoTBhTLQ+sHFmBZKhZSA+gByIwgWyJpQpPBKWgyABeEAjUDUAAYCAAmAgAFgIAA4pko2GrgABlPMFyPhRPFxN9ACMFc7vZ7UF71c7iFKmuHAE4AGwZ-gpZr0L2SF6romh0x8DVaw1+gMEZMsMNSWhpMCJu2SjcSNs7zXAfeIf0QI85-PnstXyWetU+xtKBbe8LkfZ9XyPQDgM-NIJynO0gA).

hasparus:
> If so, is there an easy way to extend the sx prop?

I think we can use declaration merging for that. I'll try that and send you a snippet.

nandorojo:
Awesome, thanks.

nandorojo:
@hasparus is there anything we're missing here to move towards a merge?

nandorojo:
@millsp has been incredibly helpful [here](https://github.com/millsp/ts-toolbelt/issues/154) with giving insight on generating template literal strings from an object type. It seems like we could use his solution for adding strict nested fields with TS 4.1.

hasparus:
@nandorojo @lachlanjc would you have a moment to give this PR another look?

There should be no breaking changes at runtime, and I think there's also no breaking changes on type level.

nandorojo:
Sure thing. CC'ing @cmaycumber too since he mentioned this would be useful in his app.

nandorojo:
@hasparus is there a way that I can install this into my app to stress test?

packages/css/src/types.ts
```diff
@@ -452,10 +452,24 @@ export type ThemeUIStyleObject = ThemeUICSSObject | ThemeDerivedStyles
...
 
 export type TLengthStyledSystem = string | 0 | number
 
+
+export declare namespace ColorMode {
+  // We won't autocomplete "mediumseagreen" inside of `sx` prop, because
+  // csstype's Property.Color is a huge union of literals you didn't include
+  // in your theme. That would be a tiny sabotage, right?
+  // Nevertheless, it's convenient to have them inside of the colors scale.
+  export type ColorValue = CSS.Property.Color;
```
nandorojo: @hasparus interesting. Does this mean it doesn't autocomplete but also doesn't break? I might be missing what this does if it can't autocomplete.

nandorojo:
> The only thing I'm left wondering is how to extend the `sx` prop. While `dripsy` follows the theme-ui spec, there are a few styles that CSS doesn't support that React Native (or React Native for Web) does.

Wanted to check in about [that comment](https://github.com/system-ui/theme-ui/pull/1090#issuecomment-710638853). 

@hasparus mentioned we could probably extend the `sx` prop with declaration merging. Not sure if that's beyond the scope of this PR or not. Luckily, the types in the code base are getting a lot cleaner, so this seems like it won't be too tough to implement.

Until the `sx` prop has declaration merging, I'm thinking I can just extend the `SxProp` (I forget what its replacement is called) and replace a few fields to fit the React Native styles. I'd have to copy some TS types over from `theme-ui`, though, to support scales and such.

lachlanjc:
> @hasparus is there a way that I can install this into my app to stress test?

@nandorojo I recommend using [`yarn link`](https://classic.yarnpkg.com/en/docs/cli/link) to try it out!

packages/css/src/types.ts
hasparus: It autocompletes builtins only inside `theme.colors`, because maybe you want `primary: hotpink` or `primary: ButtonFace`. 

The types for values assignable to `sx.color` and `sx.backgroundColor` is `Scales.Color` (the naming isn't perfect here mby ThemeUIProperty.Color? I'm open for suggestions.)

I aliased CSS.Property.Color to ColorValue because I was initially adding null here, but I left it to have a place for a comment.

hasparus:
I had to update to next version of TypeScript to work around infinite recursion on template literal types, and now `microbundle` won't build our code :D I think we have to pause this PR for a couple of days, because we're ahead of build tooling.

```
lerna ERR! yarn run prepare stderr:
(rpt2 plugin) Error: /home/runner/work/theme-ui/theme-ui/packages/css/src/scales/borders.ts(3,15): semantic error TS2305: Module '"./scales-utility-types"' has no exported member 'ScaleProperty'.
Error: /home/runner/work/theme-ui/theme-ui/packages/css/src/scales/borders.ts(3,15): semantic error TS2305: Module '"./scales-utility-types"' has no exported member 'ScaleProperty'.
```

pkieltyka:
btw, as an alternative to microbundle for a monorepo setup, I highly recommend checking out https://github.com/preconstruct/preconstruct -- which is a system created by the guys behind emotion, and I've recently adopted it and fell in love.

lachlanjc:
@pkieltyka Thanks! That’s happening in #1204 :)

nandorojo:
@hasparus which version of TS fixed the stack overflow bug for you?

hasparus:
@nandorojo `typescript@4.2.0-dev.20201211` includes PR https://github.com/microsoft/TypeScript/pull/41693

I'm having a different compilation problem on this one, though. I'm hoping I can solve it today after work.

nandorojo:
@hasparus Hey, just wanted to check on the status of this PR, I'd love to get it moving if there's anything I can do to get it in Dripsy. Thanks again!

hasparus:
@nandorojo

The released canary version is `0.6.0-canary.c5a716a.0` — I haven't tested it yet, though.

hasparus:
Problem I've noticed while working with the canary release: It's easy to introduce a circular reference if you build your exact theme from objects typed with stuff imported from Theme UI. This could be alleviated if we made the logic "purer" and passed the Theme as a generic parameter.

nandorojo:
I think that makes sense.

btw, ts-toolbelt released a stable dot path notation getter for TS string templates that has higher performance. There’s more discussion towards the end of this issue: https://github.com/millsp/ts-toolbelt/issues/154

 I believe it’s `F.AutoPath<T, S>`

hasparus:
Okay, I've tested that canary release. It defiintely lacks polish, and I had to work around my circular references.

I had to cast my `styles` to a simplified type, because the circular reference is inherent to it.

```ts
export const makeStyles = (s: ThemeStyles) =>
  s as Record<
    string,
    Record<string, number | false | string | undefined | null> | undefined
  >;
```

We could split our theme into "specification" and "implementation" parts — scales create define the theme, variants and styles use scales.

So, `styles` and variants can't be a part of user's exact theme, because this is possible to write. 

```ts
theme: {
  styles: {
    root: {
      color: theme => theme.styles.root.color
    }
  }
}
```

---

**Additional notes**

- [ ] If we go further this way, functions from `@theme-ui/color` should not depend on _Theme_ but on _BaseTheme_, maybe even on _Pick<BaseTheme, 'colors'>_. — I'm using `alpha` to define my colors scale, and since `alpha` depends on final theme it can't be used to define the final theme.

nandorojo:
When you say circular references, you're referring specifically to the function getters within the theme, right? (I don't actually use those myself, mostly just in component props, but never in the actual theme.)

hasparus:
> When you say circular references, you're referring specifically to the function getters within the theme, right?

Yes. Functions in the theme get theme as the argument. (I'm wondering if TS would allow a circular reference if we made theme "methods" instead `(this: Theme) = StylePropertyValue`). Variants, and styles are the _consuming_ part of the theme, scales are the _definition_ part of the theme.

But functional styles aren't the only problem. Take a look at this:

```ts
const root: ThemeUICSSObject = {
  backgroundColor: "background" 
}

const theme = makeTheme({
  styles: { root }
})

type ExactTheme = typeof theme;

declare module "@theme-ui/css" {
  interface UserTheme extends ExactTheme {}
}
```

`ThemeUICSSObject` depends on the final `Theme` type, final `Theme` type is `Assign<BaseTheme, UserTheme>`, and `UserTheme` depends on `theme.styles.root`, so we have a circular.

**CodeSandbox with the example: https://codesandbox.io/s/circular-explosions-4re43?file=/src/App.tsx**

We could change the API to work around this
```ts
function makeScales<TExactScales extends Scales>(scales: TExactScales): TExactScales {
  return scales
}

const scales = makeScales({ /* ... */ })

type MyScales = typeof scales

declare module "theme-ui" {
  export interface UserScales extends MyScales {}
}
```

And stop caring about keeping styles and variants in Theme type. 
We could say "Using functional styles to access `theme.styles` and variants is discouraged, use `sx.variant` instead".

nandorojo:
@hasparus I think this makes sense. The declaration merging is only going to be once you create the app. It’s totally fine to ask people to have a `MyScales`, `MyColors`, etc types. They can then feed those into their theme. I don’t think we need declaration merging on the entire theme.

Plus, if we have a `makeScales` function and such, we can add some validation `makeScales<T extends ...>`.

nandorojo:
Basically, I’m thinking the final theme type shouldn’t rely on the user theme. The theme has too many things that the user shouldn’t need to edit and will create circular references. They should just add their own types for colors, scales, etc, and the final theme object can pull all of those in.

nandorojo:
Also, since the “theme” itself doesn’t need intellisense, and only components do, maybe the user types should just get mapped onto the `sx` prop, but get ignored when writing variants, etc on your actual theme if needed. It seems like a big step up to have strict types on the sx prop, and at this point I’d love if there’s a way to get this PR to land :)

nandorojo:
In case it's of interest, I implemented this from scratch for Dripsy v3: https://github.com/nandorojo/dripsy/pull/124

yerffejytnac:
Any updates on this? Haven’t seen any commits in a while and was curious if it’s still a feature being actively developed.

🙏

hasparus:
@yerffejytnac I'm gonna salvage some parts of this PR, but after testing a prerelease we discovered this is not exactly the public API we'd like to have, as it can be a bit of a footgun.

I'm recently using Theme UI a lot more than I used it in the past years, so while I won't promise you an expected release date, it's quite likely I'll need this or an alternative feature will get implemented soon :)

MuhammadJamaluddin:
Any updates on this?

PR #561, opened 2020-01-10
Keep raw color values in theme-ui context

jxnblk:
#535 introduced a change in how raw color values are passed through Theme UI context. This PR reverts that change to be more similar to the current behavior, not requiring a separate `theme.rawColors` object

PR #1095, opened 2020-07-28
(@theme-ui/color-modes): migrate to typescript

beerose:
Part of https://github.com/system-ui/theme-ui/issues/668.


packages/css/src/types.ts
```diff
@@ -641,6 +641,7 @@ export interface Theme {
    * Define the colors that are available through this theme
    */
   colors?: ColorModesScale
+  rawColors?: ColorModesScale
```
hasparus, line 644: It's only mentioned once in the code, and it's missing from the Theme Spec in the docs.
I'd guess we could keep this as undocumented feature or maybe even drop it? @jxnblk 

It seems to me that `rawColors` may have been left here for compatibility after https://github.com/system-ui/theme-ui/pull/561

---

https://github.com/system-ui/theme-ui/blob/a96e723bad9c0d8b03c858698045acf01587cf9d/packages/color-modes/src/custom-properties.js#L76

https://theme-ui.com/theme-spec

beerose: I'll remove it from `Theme` type and work around here. There you can drop it whenever you want.

jxnblk: Yup, I think that's only there for backwards compatibility, and I suspect it's been long enough that we could drop that in v0.4

hasparus:
```suggestion
```

packages/color-modes/src/custom-properties.ts
```diff
@@ -74,7 +79,7 @@ export const createColorStyles = (theme = {}) => {
     })(theme)
   }
   const colors = theme.rawColors || theme.colors
```
hasparus, line 81:
```suggestion
  const { colors } = theme
```

PR #1107, opened 2020-08-07
chore(core): improve types

yi-tuan:
Signed-off-by: yi-tuan <1260320217@qq.com>

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -22,32 +22,12 @@ export interface SxProps {
   sx?: SxStyleProp
 }
 
-export interface IntrinsicSxElements {
-  p: JSX.IntrinsicElements['p'] & SxProps
-  b: JSX.IntrinsicElements['b'] & SxProps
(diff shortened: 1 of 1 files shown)


lachlanjc:
@hasparus Can you take a look at this?

hasparus:
I prefer the original code for two reasons

1. From user's standpoit: It uses an interface, so there's a shorter tooltip, and shorter eventual error message.
image
image

2. From contributor's standpoint: It's dumber. A big inlined type is easier to wrap your head around than an intersection of mapped type with smaller inlined type.

What's your opinion, @lachlanjc?

lachlanjc:
I don't have enough TypeScript experience to weigh in on this—I understand the basics, but personally don't use TS on any projects, so I'm not sure what should be preferred here. Sorry :)

lachlanjc:
If you don't think this is a big win, I'm totally cool with closing since it's been several months here

PR #1115, opened 2020-08-10
fix(docs): Fix preset page demo text color

### LINKED ISSUE #808: (docs) text in /demo (lorem.mdx) doesn't take color from the theme
**Describe the bug**
The text of lorem.mdx on `/demo` page doesn't take color from the theme.

**To Reproduce**
Go to https://theme-ui.com/demo/

**Expected behavior**
I suppose it should respect the color from theme for readability.

**Screenshots**
image
image

lachlanjc:
Closes #808

New behavior:

| Base | Others |
| --- | --- |
| 2226C9CB-23AD-4EA2-B1EB-05921385FF00 | 6639BFE0-F47C-4A35-8C34-EACDBF75FC26 |
| 9E1CB6D6-6398-43C6-9776-BD4C859422C1 |  6060383C-8E84-48F9-BA13-93DD727972B1 |

### DIFF
--- a/packages/docs/src/components/presets-demo.js
+++ b/packages/docs/src/components/presets-demo.js
@@ -50,7 +50,7 @@ export default () => {
           </Select>
         </label>
         <ThemeContext.Provider value={preset}>
-          <Styled.root>
+          <Styled.root sx={{ bg: 'background', color: 'text' }}>
             <Styled.h2>Colors</Styled.h2>
             <ColorPalette omit={['modes', 'header']} />
             <Styled.h2>Typography</Styled.h2>
--- a/packages/docs/src/components/presets-demo.js
+++ b/packages/docs/src/components/presets-demo.js
@@ -50,7 +50,7 @@ export default () => {
           </Select>
         </label>
         <ThemeContext.Provider value={preset}>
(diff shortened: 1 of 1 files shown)


hasparus:
Nice one! This fixes #808! 

I just noticed that my issue description was written in super confusing way BTW 😅

lachlanjc:
You’re good, I’d seen the bug too! This isn’t the prettiest solution visually but I think it’s the only answer that always works & makes sense—otherwise we’re stripping parts of the theme you’re previewing out to integrate with the site’s theme.

hasparus:
Yeah, maybe some padding would help to make it more visually pleasing... Okay, I really doubt we can do correct and visually pleasing both in this case.

I wondered if the select shouldn't change the entire theme of the docs, but the themes in showcase may actually be not compatible with each other and the docs, so the solution you have right now is better.

lachlanjc:
Went back & forth but wanted to keep the changes minimal. Implemented:

61ED297D-FEAC-4185-8E2E-90E82A490009
7EE87450-810A-4F20-89FF-E9766646D4C9

jxnblk:
Awesome, thanks!

PR #1116, opened 2020-08-10
(components): Add flexShrink 0 to Checkbox + Radio

### LINKED ISSUE #1091: Give Checkbox and Radio default flexShrink: 0

Since `Label` already has `display: flex`, I'm finding this to be a really useful form pattern:

```jsx
<Label>
  <Checkbox name="storeChannels[]" value="Online" />
  <Text>Does this store accept online orders?</Text>
</Label>
```

Currently when `Text` flows to multiple lines on narrower screens, `Checkbox` shrinks horizontally along with `Text` due to default `flex-shrink: 1`. For a bunch of stacked checkboxes, this causes staggered text alignment:

```text
[ ]   Single line label
[ ]  Label that extends to
     two lines
[ ] Label that extends and
    extends and extends to
    three lines
```

To resolve I'm wrapping `Checkbox` (and `Radio`) with `<Box sx={{ flexShrink: 0 }}>`, but I think this would be a useful built-in property.

lachlanjc:
Closes #1091

### DIFF
--- a/packages/components/src/Checkbox.js
+++ b/packages/components/src/Checkbox.js
@@ -65,6 +65,7 @@ export const Checkbox = React.forwardRef(
           mr: 2,
           borderRadius: 4,
           color: 'gray',
+          flexShrink: 0,
           'input:checked ~ &': {
             color: 'primary',
           },
--- a/packages/components/src/Radio.js
+++ b/packages/components/src/Radio.js
@@ -2,19 +2,19 @@ import React from 'react'
 import Box from './Box'
 import SVG from './SVG'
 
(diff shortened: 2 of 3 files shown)


hasparus:
Snapshots have failed because of added

```
    +   -webkit-flex-shrink: 0;
    +   -ms-flex-negative: 0;
    +   flex-shrink: 0;
```

so that's okay.

Jest's output is pretty messy -- it seems that line endings didn't serialize properly?

lachlanjc:
Just updated the snapshots, thanks for the reminder!

PR #1120, opened 2020-08-10
feat(docs): Add filenames to code blocks, including MDX Layout page

lachlanjc:
Closes #749. Using an MDX [code block meta string](https://mdxjs.com/guides/live-code#code-block-meta-string ), allows displaying file names on code blocks in the docs. This is implemented on the MDX Layout Component guide page, to clarify they’re the same file iterated on.

| Deep | Swiss |
|:--|:--|
| IMG_2164 | IMG_2165 |

### DIFF
--- a/packages/docs/src/components/code.js
+++ b/packages/docs/src/components/code.js
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import { jsx, Styled, Alert } from 'theme-ui'
+import { jsx, Styled, Text } from 'theme-ui'
 import Prism from '@theme-ui/prism'
 import { LiveProvider, LiveEditor, LivePreview, LiveError } from 'react-live'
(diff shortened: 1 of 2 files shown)


hasparus:
Lovely!

I'm wondering if `Alert` is the right component, I always thought about it as the component for _notices_ or _toasts_ but we don't use `role="alert"` there, so it should be okay I guess.

lachlanjc:
@hasparus I've switched over to `Text`, thanks!

Should I make a PR adding `role="alert"` to that component, just so it's clearer/more semantic?

hasparus:
> Should I make a PR adding role="alert" to that component, just so it's clearer/more semantic?

@lachlanjc I don't know tbh. We'll need to take a look how other component libraries to this.

PR #1121, opened 2020-08-11
feat(@theme-ui/components): Add repeat prop to Grid component

### LINKED ISSUE #1036: Add Grid support for auto-fill instead of auto-fit
I'd like to reduce the amount of markup necessary to get `auto-fill` grid behaviour instead of the current hard-coded `auto-fit` in the `<Grid>` component.

Currently a responsive grid requires only this much markup:

```js
<Grid width={[96, 128, 160]} />
```

Unfortunately if the grid has too few children to fill a row, [`auto-fit` collapses empty columns](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#Syntax). This causes my remaining grid items to expand to fill the available space, which in my case is not what I want (I'm using `<AspectRatio>` and this results in some VERY large grid members).

Currently I need this much markup to compensate:

```js
<Grid
  columns={[
    `repeat(auto-fill,minmax(96px,1fr))`,
    `repeat(auto-fill,minmax(128px,1fr))`,
    `repeat(auto-fill,minmax(160px,1fr))`,
  ]}
/>
```

I'd suggest adding support for a shorter syntax like this:

```js
<Grid width={[96, 128, 160]} repeat="fill" />;
```

There might be better syntax options too.

lachlanjc:
Closes #1036! Includes tests, snapshots, & docs.

### DIFF
--- a/packages/components/src/Grid.js
+++ b/packages/components/src/Grid.js
@@ -1,22 +1,22 @@
 import React from 'react'
 import Box from './Box'
 
-const px = n => (typeof n === 'number' ? n + 'px' : n)
+const px = (n) => (typeof n === 'number' ? n + 'px' : n)
 
-const widthToColumns = width =>
+const widthToColumns = (width, repeat) =>
   Array.isArray(width)
-    ? width.map(widthToColumns)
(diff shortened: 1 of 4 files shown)


jxnblk:
Awesome work, thanks!

PR #1122, opened 2020-08-11
(components): Render extra Embed props on iframe tag

### LINKED ISSUE #966: Embed additional props for accessibility and others
When using the <Embed> component, it would be nice to be able to pass down props onto the iframe. I couldn't find anything in the docs for it, but perhaps I've missed something?

Currently the <Embed> component wraps an iframe in the div and props like aria-label or title are attached to the parent div instead of the iframe. Perhaps having something like ```iframeProps: {}``` is too verbose and not in keeping with ThemeUI but I can't think of anything that would work better.

An alternative could be for me to create a custom component with the nessecary props and I'm happy to do it if that's the recommended guidelines.

**Additional context**
It's mostly for some accessibility reasons but I'm sure there are other cases where it would be helpful.

lachlanjc:
Closes #966, this PR makes extra props on the `Embed` component, especially/such as `title` or `aria-label`, be rendered on the `iframe` tag instead of the container div, to allow better accessibility when using that component.

This is a very slightly potentially breaking change for developers who are using container style props (`mb={3}`) directly on `Embed`, since those could now behave differently inside another div, but I don’t think that’s a common usage. Added a quick note about accessibility to the `Embed` docs as well.

Includes tests, snapshots, & docs.


hasparus (approved):
Love it.

hasparus:
Okay, one thing that's left before releasing this is updating the changelog. This could be breaking, right?

hasparus:
The types for Embed also change, since it no longer supports box props (space, color, variant).

https://github.com/system-ui/theme-ui/pull/1120

I think we can get rid of space and color props, but pass variant to the root child.

hasparus:
Hmm... maybe we could still spread on Box, but pick a greater set of iframe props to pass on to the child iframe?

lachlanjc:
Yep, I realize I've forgotten to update the changelog with several recent PRs (like print color mode support). I will add that. Sounds good with `variant` too!

hasparus (approved):

lachlanjc:
This actually wasn't quite ready to merge after our discussion @hasparus—we can use prop filtering to make using space props keep working here without a breaking change, but that can be a separate PR before we release stable

PR #1126, opened 2020-08-12
fix(docs): Fix 9 jsx-a11y console warnings

lachlanjc:
E58F3CAE-6B41-4A82-9094-ABF0CD72B12C

Working on the docs, 11 console warnings always print out, so this PR fixes 9 of them!

- One remaining one is the events on noninteractive elements; I spent awhile researching & I can’t find a solution that doesn’t worsen accessibility to make the warning go away
- The other remaining one seems to be [a bug in the library](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/477), reporting that the `<option>` tags inside a `<select>` aren’t labeled (but they are)

### DIFF
--- a/packages/docs/src/components/graph.js
+++ b/packages/docs/src/components/graph.js
@@ -1,6 +1,6 @@
 /** @jsx jsx */
 import { jsx, get, useThemeUI } from 'theme-ui'
-import { useState, useEffect, Fragment } from 'react'
+import { useState, useEffect } from 'react'
 import merge from 'deepmerge'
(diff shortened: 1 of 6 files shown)


hasparus:
Nice one! I'll take the liberty of merging it.

lachlanjc:
Fab :) any idea what's happening with the maintenance/publishing of this project @hasparus? I've been using it every day for a year, on nearly all my projects, switched my whole org to it, & seeing it going stale is concerning.

PR #1127, opened 2020-08-13
Fix small typo in 'sketchy' line

diemkay:
And also bring 'sketchy' theme in line with other preset examples (theme name takes you to its docs).


hasparus:
Thanks for the PR, @diemkay! Good catch!

I just noticed we have another problem with this preset's docs. https://github.com/system-ui/theme-ui/issues/1128

diemkay:
My pleasure, and thanks for being so prompt! I'm just learning about all this and it's very exciting.

PR #1141, opened 2020-08-19
fix: Use default export from package instead of named export version

### LINKED ISSUE #1097: Webpack 5: Should not import the named export 'version'
**Describe the bug**

https://webpack.js.org/migrate/5/#cleanup-the-code

> Using named exports from JSON modules: this is not supported by the new specification and you will get a warning. Instead of import { version } from './package.json' use import package from './package.json'; const { version } = package;

**To Reproduce**

Try to use `webpack@^5.0.0-beta.22` and `theme-ui@0.3.1` 

getting 

```
WARNING in ./node_modules/@theme-ui/core/dist/index.esm.js 37:23-30
Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)
 @ ./node_modules/theme-ui/dist/index.esm.js 1:0-37 2:0-65 2:0-65 2:0-65 2:0-65 2:0-65 59:9-12

```

in nextjs (with webpack5 enabled) it won't even compile...

windkomo:
Fix #1097

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -6,12 +6,14 @@ import {
 import { css, Theme } from '@theme-ui/css'
 import * as React from 'react'
 import deepmerge from 'deepmerge'
-import { version as __EMOTION_VERSION__ } from '@emotion/core/package.json'
+import packageInfo from '@emotion/core/package.json'
 
 import './react-jsx'
 
 export * from '@theme-ui/css/dist/types'
 export * from './types'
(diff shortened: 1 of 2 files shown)


hasparus:
Thanks for the PR, @windkomo!

hasparus (approved):

J-Mwaura:
I changed file from extension .json to .ts and then used export default.

PR #1170, opened 2020-09-03
fix(components): add missing dependency to @types/styled-system

hasparus:
When we release this, people should be able to use Theme UI 0.4 in TS project with no `skipLibCheck`.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -16,7 +16,8 @@
     "@styled-system/color": "^5.1.2",
     "@styled-system/should-forward-prop": "^5.1.2",
     "@styled-system/space": "^5.1.2",
-    "@theme-ui/css": "^0.4.0-rc.5"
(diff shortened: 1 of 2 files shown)


hasparus:
The problem happened, because we it as transitive dependency.

hasparus:
@atanasster, you're using `@theme-ui/components`, right? Have you encountered this problem?

atanasster:
@hasparus - i am using theme-ui/components but importing the whole theme-ui package and never had this issue. Using it with gatsby as well and never had the issue. Here are my configs: 

https://github.com/ccontrols/component-controls/tree/master/ui/components

atanasster:
@lachlanjc - do you know what happened with this one? Should I re-open and review or we decided to close it?

lachlanjc:
@atanasster We decided to close because this dependency is already there on master iirc

atanasster:
ah, thanks, removing from my tasks :)

PR #1182, opened 2020-09-08
Fix theme editor output in docs

mattfwood:
Currently if you go to the [customize / editor page in the docs](https://theme-ui.com/customize/) and edit the theme, the content rendered in the output doesn't update.

I managed to make it work by putting the output editor inside the `<EditorProvider />` and then reading the theme value from context.


hasparus:
Hi @mattfwood 👋

Great PR! Thanks!

mattfwood:
Just want to check in on this @hasparus @jxnblk , we could use this fix for allowing our users to create custom themes

hasparus:
@jxnblk I can cherry-pick this to [`previous`](https://github.com/system-ui/theme-ui/tree/previous) and then we can release 0.3.2 stable, and updated docs.

PR #1215, opened 2020-10-16
fix: space value type

theowenyoung:
unified space type


hasparus:
Thanks for the PR, @theowenyoung :)

Issue #1037, opened 2020-07-04
Add search to the docs (Algolia's DocSearch, maybe?)


julianbenegas:
**Is your feature request related to a problem? Please describe.**
Having a search thing in the docs would be awesome. Although by now I know where everything is, I remember when getting started with theme-ui, I had a hard time finding the things I was looking for.

**Describe the solution you'd like**
Algolia has a [specific solution for documentations](https://docsearch.algolia.com/), used by [React](https://reactjs.org/) and many other doc sites.

**Additional context**
I see that it's not so complicated to integrate, it's free, and would be an amazing improvement to the docs. I'd be happy to help with the integration!

FlashBlaze:
Are there any updates on this?

hasparus:
I just applied to DocsSearch, waiting for Algolia response email.

lachlanjc:
@hasparus Any update on DocsSearch?

hasparus:
> @hasparus Any update on DocsSearch?

We're in. Slight caveat, AFAIK it indexes only the main domain, but it should be enough.

lachlanjc:
This is live as of last winter!

PR #1223, opened 2020-10-23
(docs) alphabetizes components list in sidebar

simpixelated:
I'm assuming a CHANGELOG update is not required, but happy to do so.


lachlanjc:
Hmm. I’m cool either way but I actually love the order they’re in now—they’re grouped by category/how often you need them, so I can always find what I’m looking for there super fast instead of trying to think about what letters they use

hasparus:
I'd probably stay with the current order. Variants and Box on the top just seem "logical", though I understand that this might be subjective.

simpixelated:
I had though about leaving Variants at the top, since its not technically a component. Box would naturally remain near the top. I guess if the current order is working for people, it's not a problem. But as someone relatively new to the project, I find it difficult to see a logical pattern and end up scrolling up and down trying to find what I want.

hasparus:
Hmm... maybe then

```
- Variants
- Box
- Everything else alphabetically
```

@lachlanjc what do think?

I feel that a search bar would be the solution to this problem.

Maybe we could group these components by "topic"? Layout, Typography, Form Controls, uhh others at the bottom? I'm not sure where would I put Embed.

atanasster:
I agree by topic. Utilities at bottom with Embed + Portal, Keyboard, Responsive etc?

lachlanjc:
Closing in favor of search (#1037)

Issue #1177, opened 2020-09-06
Ability to customize open/close icon and its position in the Select component


boi87:
**Is your feature request related to a problem? Please describe.**
Using the Team UI Select, I would like to be able to customize the icon with the pointing down triangle and possibily also be able to decide whether to align it to the right or to the left of the default dropdown text. Is it something that is already achievable?
Also, I am not sure if there is a boolean `open` prop.

![image](https://user-images.githubusercontent.com/25410752/92329667-33d43900-f069-11ea-935d-dd0021148bc1.png)

**Describe the solution you'd like**
- A prop for the `icon`, which ideally accepts either a `src` or a some JSX as svg 
- Another prop to decide its position against the text would also be nice. I'm not sure if this is already possible somehow. If it is, any hint on how to do it would be very appreciated.
- A boolean prop which changes on toggle (e.g. `open`) would be a very nice touch too!

**Describe alternatives you've considered**
At the moment I'm sticking to the default design...

Thank you!!

atanasster:
@boi87 - this is not possible in the current code base and built-in tracking the keyboard and mouse events for open/close state can be more complicated than desired at this stage. 

What do you think of this simpler solution that can be turned into a PR: https://theme-ui-draft.netlify.app/components/select#custom-arrow-icon 

- it allows you to replace the arrow icon
- you can set the icon with all possible props - margins/size etc
- you can track the keyup/click/blur events on your end and pass in an open or close icon.

source: 
https://github.com/atanasster/theme-ui/blob/select-custom-arrow/packages/docs/src/pages/components/select.mdx

boi87:
@atanasster - Hi and thanks for taking time to assist with this feature request! 
Allowing to pass a prop
It definitely would be an improvement.

atanasster:
@boi87 , thanks very much, I will then submit it as a PR. And possibly for a future release we can also integrate builtin handling of the open/close state.

boi87:
@a

Issue #1151, opened 2020-08-24
Feature: customize Select arrow


flo-sch:
**Is your feature request related to a problem? Please describe.**

While working on a small design system together with a designer, I realized that there seems to be no way to (fully) customize the Select's `<DownArrow />` svg at the moment.

**Describe the solution you'd like**

Optimally, a way to fully customize the appearance from the theme.
But to my knowledge, there is no browser-compatible, CSS-only way to achieve that?

Another option is an opt-in prop to specify your own SVG, which would default to the current one?

**Describe alternatives you've considered**

Its style can be overriden using selectors such as shown in [this issue](https://github.com/system-ui/theme-ui/issues/842#issuecomment-613643538), but for instance the `path` it uses cannot (not in a native, browsers-compatible way?)

```js
// theme.js
forms: {
  select: {
    '& + svg': {
      width: '1em',
      height: '1em',
      ml: '-1.5em',
      fill: 'none',
      path: {
        // this does not seem to work in Firefox
        d: 'path("M5 8L8 11L11 8")',
        stroke: 'currentColor',
        strokeLinecap: 'round',
        strokeLinejoin: 'round'
      }
    }
  }
}
```

Right now I have built a custom Select with a `useLayoutEffect()` hook overriding those, but I find that a bit cumbersome, also I do not like the query selectors that are tied to the DOM of the current `Select` (which could change in the future):

```jsx
/** @jsx jsx */
import { Select, jsx } from 'theme-ui';
import { useRef, useLayoutEffect } from 'react';

const CustomSelect = (props) => {
  const select = useRef();

  useLayoutEffect(() => {
    if (select.current && select.current.parentNode) {
      const svg = select.current.parentNode.querySelector('svg');

      if (svg) {
        const path = svg.querySelector('path');

        if (path) {
          svg.setAttribute('viewBox', '0 0 16 16');
          path.setAttribute('d', 'M5 8L8 11L11 8');
        }
      }
    }
  });

  return (
    <Select
      ref={select}
      sx={{
        '& + svg': {
          width: '1em',
          height: '1em',
          ml: '-1.5em',
          fill: 'none',
          path: {
            stroke: 'currentColor',
            strokeLinecap: 'round',
            strokeLinejoin: 'round'
          }
        }
      }}
      {...props}
    />
  );
};
```

**Additional context**

https://theme-ui.com/components/select

[...]

PR #1232, opened 2020-10-31
feat: select custom arrow

atanasster:
issue: #1177 #1151

added arrow prop to `Select`

### DIFF
--- a/packages/components/src/Select.js
+++ b/packages/components/src/Select.js
@@ -34,9 +34,7 @@ export const Select = React.forwardRef(({ arrow, ...props }, ref) => (
         bg: 'transparent',
       }}
     />
-    {arrow ? (
-      arrow
-    ) : (
+    {arrow || (
       <DownArrow
         sx={{
           ml: -28,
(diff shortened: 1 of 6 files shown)


packages/components/src/Select.js
```diff
@@ -34,12 +34,16 @@ export const Select = React.forwardRef((props, ref) => (
...
-    <DownArrow
-      sx={{
-        ml: -28,
-        alignSelf: 'center',
-        pointerEvents: 'none',
-      }}
-    />
+    {arrow ? (
+      arrow
+    ) : (
```
lachlanjc, line 39:
```suggestion
    {arrow || (
```

lachlanjc (approved):

packages/components/src/Select.js
atanasster: @lachlanjc  thanks a lot for reviewing - I checked in changes with the simplified syntax

hasparus:
Sweet! Could you update the changelog? (I feel we need a PR template with a todo list)

hasparus:
What's the difference between this and #1151?

lachlanjc:
@hasparus #1151 is the issue, this is a PR, unless I’m missing something?

hasparus:
> @hasparus #1151 is the issue, this is a PR, unless I’m missing something?

🤦

I was so sure there's been a duplicate...

PR #1236, opened 2020-11-03
feat(docs): Add custom content for sketchy preset demo

### LINKED ISSUE #1128: Preset `sketchy` docs don't show its fonts and components.
**Describe the bug**

Preset `sketchy` docs don't show its fonts and components.

**To Reproduce**
1. Go to preset-sketchy demo website. https://themeui-sketchy.netlify.app/
2. Go to its page in Theme UI docs. https://theme-ui.com/presets/sketchy
3. Notice the discrepancy.

**Expected behavior**

- *(minimum)* Theme UI docs should dynamically fetch fonts for presets.
- *(would be nice)* Theme UI docs could link to external preset's own docs with a Message at the top of the page?
- *(would be awesome)* Presets could declare an MDX or JS file to be rendered in the docs?

**Screenshots**

demo website         |  Theme UI docs
:-------------------------:|:-------------------------:
  |

atanasster:
issue #1128

- fix: the context was not picked up, so even color palette was not correctly displayed
- fix: the font needed to be dynamically loaded in the header.
- feat: added components.mdx instead of lorem.mdx. The file https://github.com/beerose/theme-ui-sketchy/blob/master/src/components/components.mdx is copied from @beerose repository and we should give credits in the appropriate place.

sites:
before: https://theme-ui.com/presets/sketchy
now: https://theme-ui-draft.netlify.app/presets/sketchy


hasparus (changes requested):
Hey atanasster! Thanks for the PR!

image

I think borders and buttons don't really make sense for most of the presets, because they don't have variants.

I wouldn't say that beerose's components.mdx was meant to be rendered with different presets.
image

hasparus:
Instead of _one size fits all_ solutions like we have right now, I think we should allow presets to affect how their docs are displayed. We already have _one size fits all_ right now.

I had mentioned a few ideas in the issue you linked.

> **Expected behavior**
>
> - *(minimum)* Theme UI docs should dynamically fetch fonts for presets.
> - *(would be nice)* Theme UI docs could link to external preset's own docs with a Message at the top of the page?
> - *(would be awesome)* Presets could declare an MDX or JS file to be rendered in the docs?

atanasster:
@hasparus thanks for the feedback, a few changes: 

- the JSON text area is now outside of the theme provider.
- all the 'variants'  components are now dynamically displayed depending on the theme:
  1. borders
  2. buttons
  3. badges
  4. alerts

example with borders variants: https://theme-ui-draft.netlify.app/presets/sketchy
example no borders variants: https://theme-ui-draft.netlify.app/presets/swiss

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
-
+- docs `preset/presetname`fix fonts and preview components. Issue #1128
```
lachlanjc: Let's make this a little easier to understand

packages/docs/src/components/components.mdx
```diff
@@ -0,0 +1,217 @@
...
+export const Radii = () => {
+  const { theme } = useThemeUI();
+  if (typeof theme.radii === 'object') {
+    return (
+      <React.Fragment>
+        <Divider />
+        <h2>Borders</h2>
+        <Flex sx={{ flexWrap: "wrap" }}>
+          {Object.keys(theme.radii).map(key => (
+            <Box key={`border_${key}`} p={3} m={2} color="text" bg="muted" sx={{
```
lachlanjc: Make sure to run Prettier :)

packages/docs/src/components/components.mdx
```diff
@@ -0,0 +1,217 @@
...
+  onSubmit={e => e.preventDefault()} sx={{width: '100%', maxWidth: "500px"}}>
+  <Label htmlFor='username'>Username</Label>
+  <Input
+    name='username'
+    mb={3}
+  />
+  <Label htmlFor='password'>Password</Label>
+  <Input
+    type='password'
+    name='password'
```
lachlanjc: These fields need an `id` for the labels to work correctly, or put the input inside the label

packages/docs/src/components/components.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+## Card
+
+<Card
+  mb={5}
+  sx={{
+    maxWidth: 400,
+    padding: "30px"
+  }}>
+  <Text>
```
lachlanjc: Switch to `Paragraph` now?

lachlanjc (changes requested):
This is a great improvement! A few little things to go. Also, not sure if this is an issue with the preset or the docs, but two buttons aren't using the right font on iOS:

image

CHANGELOG.md
atanasster: updated into 2 separate items - the reported bug, and also adding the components

packages/docs/src/components/components.mdx
atanasster: I did now - it was originally taken from @beerose repo

packages/docs/src/components/components.mdx
atanasster: Thanks, fixed

packages/docs/src/components/components.mdx
atanasster: switched

lachlanjc:
This is looking great! Are these borders supposed to look different though?
image

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+- docs site - fix the fonts on `preset/presetname`. Issue #1128
+- docs site - add preview components on `preset/presetname`.
```
lachlanjc: Personally I feel like CHANGELOG should be just for the library, since it's something we hand write, & people can assume the docs are always getting better. I'm not sure if people are that curious about what specific upgrades we're making to docs if they're more casual users not looking at GitHub?

CHANGELOG.md
```diff
@@ -122,6 +125,8 @@
 
 ### `@theme-ui/core`
 
+> > > > > > > develop
```
hasparus: .

hasparus:
Yeah, I still don't think we should render the same example for all themes. It's hard to get right — borders seem to be broken for `sketchy`, Card makes little sense for `swiss`. I think most of the presets we have were meant just as color themes and mostly for text display. `sketchy` is the only one with themed components, and I think we should copy its docs into ours, because what happens when somebody else makes a theme with different docs?

I think we should just respect the font and link to preset's own docs with a big visible banner.

image
image

atanasster:
I think that all presets would benefit from testing how the main components render, not just sketchy - some of the presets do have issues with rendering some color combinations, or spacing. Also having each preset to supply its own mdx file seems a bit of an overkill for maintenance. 
Anyway, we can just close this PR for now.

CHANGELOG.md
atanasster: > Are these borders supposed to look different though?

Not sure - already this broke once as useThemeUI stopped working and I had to replace it with useTheme from emotion.
I can check whats wrong with borders in a couple of days when I get some time - but it seems @hasparus doesn't like the implementation, so maybe easiest is to close this PR and let someone else take a stab?

hasparus:
> Also having each preset to supply its own mdx file seems a bit of an overkill for maintenance.

True, but I thought we could just display Lorem.mdx _or_ a link to preset's own docs.

> I think that all presets would benefit from testing how the main components render, not just sketchy - some of the presets do have issues with rendering some color combinations, or spacing. 

We could have lorem.mdx and components.mdx in tabs I suppose?

CHANGELOG.md
hasparus: > Not sure - already this broke once as useThemeUI stopped working and I had to replace it with useTheme from emotion.

When did `useThemeUI` break?

hasparus: > it seems @hasparus doesn't like the implementation, so maybe easiest is to close this PR and let someone else take a stab?

It's not that I don't like it, I just think it's hard to get right. I think we could give it a stab starting from this point.

atanasster:
> We could have lorem.mdx and components.mdx in tabs I suppose?

I just replaced Lorem with Components. If a large body of text (ie Lorem) is desired, it can be added to Components. I really don't see a need for Lorem - but am not too strong on it and someone else can do a different PR.

CHANGELOG.md
atanasster: > When did `useThemeUI` break?

In this PR - when the buttons stopped displaying their theme borders. I dont really know what happened tbh, just fixed by going upstream to emotion

atanasster: > It's not that I don't like it, I just think it's hard to get right. 

I don't mind either way - imo the best solution was to re-use the Components for all themes (its a pretty good piece of code to showcase any theme), and basically replace the more minimalist Lorem. 

For linking to other sites - this can be done additionally in credits.links. 
For creating mdx files for each theme - I really dont like it.

lachlanjc:
@hasparus Any issue with merging this as-is? I think it’s providing a better experience for the sketchy docs than we currently are:

Screenshot


Before: https://theme-ui.com/demo  
After: https://theme-ui-ldlav21vx.vercel.app/demo

lachlanjc:
@hasparus ping on opinions here?

hasparus:
:rocket: PR was released in [`v0.14.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.14.0) :rocket:

PR #1237, opened 2020-11-05
Add support for automatic JSX runtime

### LINKED ISSUE #1160: Syntax error: pragma and pragmaFrag cannot be set when runtime is automatic.
I started having this error since updating Gatsby to the most recent version and updating React to v17.

`Syntax error: pragma and pragmaFrag cannot be set when runtime is automatic.`

edit: The same issue now exists with the latest version of React `react@16.14.0`, right now the only fix is to go back to `react@16.13.1 `and `react-dom@16.13.1`

dcastil:
Closes #1160

React introduced the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) and Emotion added support in https://github.com/emotion-js/emotion/pull/1970. So everything is in place for Theme UI to support the new JSX transform. 😊

To dos:

- [x] Resolve discussions I added below.
- [x] Test automatic JSX runtime manually
- [x] Write tests
- [x] Update documentation


packages/parse-props/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@theme-ui/parse-props",
```
dcastil, line 2: I created a new package for the `parseProps()` functionality from `@theme-ui/core` which Theme UI puts in front of Emotion's jsx function. The reason for that is that we need a separate entry point for the new jsx transform, it needs to be a file or directory called `jsx-runtime` and `jsx-dev-runtime`. This is the only way I found to share the logic between the new and old jsx transform without duplicating the logic in the resulting bundles.

I'm not very happy about this approach though because this package is for internal use only and now it's exposed to the public. Maybe there is another way to solve this? Or maybe it's okay to ship the code twice?

packages/core/jsx-runtime/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@theme-ui/core-jsx-runtime",
```
dcastil, line 2: The bundles for jsx-runtime and jsx-dev-runtime need to be accessible via the paths `foo/jsx-runtime` and `foo/jsx-dev-runtime` where `foo` can be any path but the rest is fixed.

I added them both to `@theme-ui/core` so one can use the comment

```js
/** @jsxImportSource @theme-ui/core */
const foo = <div sx={{ color: 'hotpink' }}>Hello <b>world</b></div>
```

which is then transformed to

```js
import { jsx as _jsx } from '@theme-ui/core/jsx-runtime'
import { jsxs as _jsxs } from '@theme-ui/core/jsx-runtime'

const foo = _jsxs('div', {
  sx: {
    color: 'hotpink'
  },
  children: [
    'Hello ',
    _jsx('b', {
      children: 'world',
    }),
  ],
})
```

without the need to install a new dependency.

But maybe it makes sense to expose the same entry points in the `theme-ui` package?

dcastil, line 2: I also don't know how npm determines which files end up in the published package. I think we need to specify the directories `jsx-runtime` and `jsx-dev-runtime` explicitly to be included.

packages/core/jsx-runtime/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "source": "../src/jsx-runtime.ts",
+  "main": "dist/index.js",
+  "module": "dist/index.esm.js",
+  "types": "dist/index.d.ts",
+  "dependencies": {
+    "@emotion/core": "^10.1.1",
+    "@theme-ui/parse-props": "^0.4.0-rc.5"
```
dcastil, line 9: The version here needs to keep in sync with the overall repo version. I can imagine that Lerna wouldn't update this one here because it isn't registered as workspace package. 🤔

packages/core/jsx-runtime/package.json
dcastil: I added the JSX runtime to the `theme-ui` package as well in 085a1479. Now both of these comments can be used:

```js
/** @jsxImportSource theme-ui */
/** @jsxImportSource @theme-ui/core */
```

similar to how it worked with the classic JSX runtime.

packages/parse-props/package.json
arshad: I wonder if we could use a *subpackage* approach similar to what emotion is doing. The only caveat might be `microbundle` which does not support multiple entry point.

dcastil: I think we could also just include the code in both bundles. `@theme-ui/parse-props` is roughly 300 Bytes now, shouldn't be too bad to have it in the bundle twice.

packages/core/jsx-runtime/package.json
hasparus: `files` and `npmignore`?

dcastil: [The npm docs about the `files` field](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files) say

> Omitting the field will make it default to ["*"], which means it will include all files.

We aren't using it in Theme UI so I think we don't need to change anything.

packages/core/package.json
```diff
@@ -18,15 +21,16 @@
...
   },
   "dependencies": {
-    "@emotion/core": "^10.0.0",
+    "@emotion/core": "^10.1.1",
     "@theme-ui/css": "^0.4.0-rc.5",
+    "@theme-ui/parse-props": "^0.4.0-rc.5",
     "deepmerge": "^4.2.2"
   },
   "peerDependencies": {
     "react": "^16.11.0"
```
dcastil, line 30: Maybe we should make `v16.14.0` the overall React peer dependency version range because technically

```js
import { jsx } from 'theme-ui/jsx-runtime'

jsx('div', {})
```

will break on React below `v16.14.0`.

But then we would force everyone to upgrade React even though not everyone wants to use the automatic JSX runtime.

packages/core/jsx-runtime/package.json
dcastil: Random thought: Maybe we can change this line to `"@theme-ui/parse-props": "*"`?

packages/core/src/index.ts
```diff
@@ -35,27 +32,6 @@ export * from './types'
...
-  for (let key in props) {
-    if (key === 'sx') continue
-    next[key] = props[key]
-  }
-  const css = getCSS(props)
-  if (css) next.css = css
-  return next
-}
-
 export const jsx: typeof React.createElement = (type, props, ...children) =>
```
dcastil, line 35: Random idea: If the automatic runtime is the new default, then we could move this `jsx` function to a separate package to decrease bundle size.

This would be a breaking change, so we could do it gradually by creating the package, importing the function here and display a little deprecation warning during development that this will be removed in the next release or `v1.0.0`.

Thoughts on this?

RobIsHere: I noticed recently how they do it in CRA. Maybe it helps. In create react app the new jsx runtime is auto-enabled or -disabled by this code at https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L74
```

const hasJsxRuntime = (() => {
  if (process.env.DISABLE_NEW_JSX_TRANSFORM === 'true') {
    return false;
  }

  try {
    require.resolve('react/jsx-runtime');
    return true;
  } catch (e) {
    return false;
  }
})();

```

dcastil: Thanks for the info @RobIsHere! I think we can't use this approach because then we would assume that everyone with React v16.14.0 or newer uses the new runtime, although not everyone uses it. CRA can enforce the automatic runtime whenever it's available because CRA itself controls the babel config of a project and can modify the config there respectively. But we can't do that in Theme UI, we need to support both ways for now.

packages/core/package.json
lachlanjc: It'd certainly be great not to get warnings about using React 17 if we do explicitly support it, so sounds good to me!

packages/core/jsx-runtime/package.json
dcastil: I changed it to `"@theme-ui/parse-props": "*"`, so we don't need to worry about that.

packages/core/package.json
dcastil: I added React v17 to the peerDependency range in e87faca and kept the base version of the dependency. 😊

dcastil: Bumped peer dep to `^16.14.0 || ^17.0.0` 39d1b6c

packages/core/src/index.ts
dcastil: `jsx` function is a one-liner, we'll just keep it, doesn't make sense to move anywhere.

arshad:
This commit thread is a roller coaster of _emotion_ (pun intended?)

> fml tpyescript is so hard
> why am I doing this to myself

Thanks for putting in the work for the rest of us @dcastil 👍

dcastil:
@arshad Haha yes, I was fighting TypeScript and as I fixed type bugs, new came up again and again. 😄 But now it works!

hasparus:
I'll handle the conflicts and changelog.

dbertella:
@dcastil is there any additional library to install in order to use `jsxImportSource` because at the moment I can't get it to work with the example in the readme. I opened an issue here if you need more context #1474 not sure if I'm just missing something very obvious here. Thanks!

hasparus:
@dbertella's comment is already answered in another issue, but I'll leave this here for posterity.

**To use Theme UI with automatic runtime you need Theme UI version greater than 0.5.0-alpha.**

However, classic runtime still works, so if you're already using it you don't need to migrate, it's possible to configure this in your Babel config or by setting `DISABLE_NEW_JSX_TRANSFORM=true` env var in CRA (see https://github.com/system-ui/theme-ui/issues/1474#issuecomment-772337040).

PR #1242, opened 2020-11-07
Pass size prop down to IconButton

marcwiest:
Close and IconButton both accept the size prop, but it's not pass down.


hasparus:
Thanks for the PR, @marcwiest! Nice catch!

PR #1243, opened 2020-11-07
ts: tailwind theme add types

### LINKED ISSUE #1139: Tailwind preset is not compatible with `Theme` TypeScript type
**Describe the bug**

The [Tailwind preset](https://theme-ui.com/presets/tailwind) is not compatible with `Theme` TypeScript type.

**To Reproduce**
1. Copy & paste the JSON theme definition and assign it to a variable with type `Theme` (via `import { Theme } from 'theme-ui'`)
2. Watch out for type errors (Note: There are more TypeScript errors once you've fixed/commented out the initially reported errors)

Here is a repo on CSB: https://codesandbox.io/s/repro-theme-ui-tailwind-typescript-olnsp

**Expected behavior**

The presets should be compatible with the `Theme` type.

**Screenshots**

image

atanasster:
issue #1139

@hasparus - I suspect most other themes will also need to be updated.


packages/preset-tailwind/src/index.ts
```diff
@@ -237,7 +238,7 @@ export const buttons = {
   },
 }
 
-export const colors = {
+export const colors: Theme['colors'] = {
```
hasparus: Do we have to annotate `colors` and `baseFonts`?
I'm thinking about exact theme types, and annotating it explicitly here would forget information about color names / font family names.

packages/preset-tailwind/src/index.ts
```diff
@@ -281,25 +282,25 @@ export const fontSizes = [
...
+  normal: 400,
+  medium: 500,
+  semibold: 600,
+  bold: 700,
+  extrabold: 800,
+  black: 900,
 }
 
-export const fontWeights = {
+export const fontWeights: Theme['fontWeights'] = {
```
hasparus: Similar comment as above. I'd drop fontWeights and baseFontWeights annotations.

atanasster: We dont have to annotate them - i did it thinking of it as a tutorial to build your own themes, it will catch typing issues earlier.
I can remove them - wdyt?

hasparus: We'll export `makeTheme` for TS users and preset maintainers. I thought about it for a longer time, and I think there`s little reason not to include it. Let's keep this open for a bit, I'll show you what I mean sometime on Discord, WDYT?

atanasster: sounds good

lachlanjc:
@hasparus How close is this to ready?

PR #1246, opened 2020-11-09
Fix support for rgb/hsl colors in @theme-ui/color

### LINKED ISSUE #1245: @theme-ui/color `getColor` removes closing parenthesis from valid color values
**Describe the bug**
`@theme-ui/color` functions removes closing parenthesis from valid color values.

**To Reproduce**
1. Use a theme color value in a format with parentheses, such as `hsl(a)` or `rgb(a)`
2. Pass that theme key to one of the color functions in `@theme-ui/color`, such as `alpha`
3. See error: `Error: Couldn't parse the color string. Please provide the color as a string in hex, rgb, rgba, hsl or hsla notation.`

The regex in [`getColor`](https://github.com/system-ui/theme-ui/blob/72bd950a345fdf58d0027f92ef2cf3149f970afb/packages/color/src/index.ts#L7) is stripping the `)` off, and passing something like `rgba(0,0,0,0.4`.

NickChristensen:
Fixes #1245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### `@theme-ui/color`
+
+- Fix support for rgb/hsl color values
+
 ## v0.4.0-rc.6 2020-11-08
 
 ### `@theme-ui/core`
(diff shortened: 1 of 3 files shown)


hasparus:
OMG. Great catch, @NickChristensen! Thanks!

PR #1252, opened 2020-11-10
fix: field component should use id if passed

deckchairlabs:
Sometimes a field name is not unique enough. This PR allows passing an id to the field and the related label `htmlFor`


hasparus:
thanks for the PR, @deckchairlabs! Totally agreed.

lachlanjc (approved):

Issue #1254, opened 2020-11-11
Hidden props for button doesn't hide button


cktang88:
**Describe the bug**
I tried applying
`<Button hidden={true}>hello</Button>`
but the button is still visible.

**To Reproduce**
See above.

**Expected behavior**
I'd expect the button to be hidden with that attribute.

**Screenshots**
N/A

**Additional context**
Using v0.3.1

There was a previous indication this was a bug here: https://github.com/system-ui/theme-ui/discussions/1253#discussioncomment-126903

lachlanjc:
Do we document somewhere that `hidden` is an available prop? I’ve never heard of it/seen it in the code.

atanasster:
@lachlanjc , `hidden` is an html attribute that is applicable to all tags, its not a theme-ui thing

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

lachlanjc:
We can close this @hasparus!

PR #1255, opened 2020-11-11
Button hidden prop

atanasster:
Issue: #1254

Fix Button not respecting hidden prop

`<Button hidden>Click me</Button>`


lachlanjc (approved):
Looks great! Maybe add a test too so this doesn't break in the future? https://github.com/system-ui/theme-ui/blob/master/packages/components/test/index.js I don't have strong feelings about it.

atanasster:
thanks @lachlanjc - great suggestion, I added a Button / hidden test

hasparus (approved):
Amazing!

deckchairlabs:
Should this check be moved to Box, since all/most components would be affected by this?

atanasster:
> Should this check be moved to Box, since all/most components would be affected by this?

Box and others actually work fine, since it doesnt set the display style. Afaics its only related to Button

lachlanjc:
We should probably do a pass-through of other components if this is something we want to consistently support. I'd expect Grid & Flex don't work since they're setting `display` as well

deckchairlabs:
From what I understand the `hidden` attribute is more semantic, for screen readers etc. So something all components should support in my opinion.

But that's probably best suited for a separate PR dealing with accessibility in general.

PR #1256, opened 2020-11-12
fix: min-content on checkbox root

deckchairlabs:
Adding `min-content` as `minWidth` to checkbox root element prevents clipping on narrow width parents.﻿

### DIFF
--- a/packages/components/src/Checkbox.js
+++ b/packages/components/src/Checkbox.js
@@ -39,7 +39,7 @@ const CheckboxIcon = (props) => (
 
 export const Checkbox = React.forwardRef(
   ({ className, sx, variant = 'checkbox', children, ...props }, ref) => (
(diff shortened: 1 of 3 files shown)


lachlanjc:
This totally makes sense. Are there any specific default cases you can show where this changes the result? & should we have this on `Radio` as well?

deckchairlabs:
I've put together a codesandbox showing the result, had to hack it since the root component of both Checkbox and Radio doesn't accept any props. I'll update the PR to include the fix for the Radio component as well.

https://codesandbox.io/s/pedantic-lamport-mr9c6?file=/src/App.tsx

lachlanjc:
Amazing, thanks for both of those!

lachlanjc (approved):

PR #1261, opened 2020-11-14
Upgrade to Emotion v11

### LINKED ISSUE #1258: Upgrade to emotion v11
Hi there, just FYI that emotion v11 has reached a prod release: https://github.com/emotion-js/emotion/releases/tag/emotion%4011.0.0

dcastil:
Closes #1258

To dos:

* [x] Test whether Theme UI docs are working with Emotion v11
* [x] Fix TypeScript issues in Jest tests


dcastil:
The last remaining TypeScript issue is that Emotion's `styled` components using the `withComponent` modifier [don't accept a `css` prop](https://github.com/system-ui/theme-ui/runs/1401033949#step:7:328).

The Box component in `packages/components/test/types.tsx:89` accepts it, but the SectionBox in `packages/components/test/types.tsx:84` doesn't.

@hasparus do you know how this could be resolved?

deckchairlabs:
@dcastil I suppose box would need to specify that it does take `css` prop? As the props only seem to be inferred when passing an actual component, and not a string tag?

https://github.com/emotion-js/emotion/blob/master/packages/styled/types/base.d.ts#L36

lachlanjc:
Thank you so much for your work here @dcastil!!

hasparus:
Hey @dcastil, is there a way I can DM you? Twitter, email or sth like that?

I'm trying to set up some means of communication between frequent contributors (AKA the place to yell "@hasparus, you promised me that code review a week ago"), and I'd like to send you an invite link :)

dcastil:
@hasparus That's cool! I DMed you on Twitter.

examples/dark-mode/src/layout.js
```diff
@@ -1,10 +1,10 @@
 /** @jsx jsx */
 import { jsx, useColorMode, Layout, Header, Main, Container } from 'theme-ui'
-import { Global } from '@emotion/core'
+import { Global } from '@emotion/react'
```
hasparus, line 3: I'm so happy they finally renamed it.

packages/core/src/index.ts
```diff
@@ -1,12 +1,13 @@
...
-  InterpolationWithTheme,
-} from '@emotion/core'
+  Interpolation,
+} from '@emotion/react'
 import { css, Theme } from '@theme-ui/css'
 import * as React from 'react'
 import deepmerge from 'deepmerge'
-import packageInfo from '@emotion/core/package.json'
+import packageInfo from '@emotion/react/package.json'
+import {} from '@emotion/react/types/css-prop'
```
hasparus, line 10: Let's leave this a side note here. I think we should also make `sx` prop types opt in, as we have users who may prefer to use `@theme-ui/components` instead of JSX pragma.

hasparus (approved):
LGTM.
This would be a small breaking change, as users will have to update their emotion deps, and fix a bunch of TS errors. Do you think we should bump to v0.5.0 while merging it?

I'm aware we didn't release 0.4.0, but the `rc` versions are "stable enough" and a bunch of projects are using them.

hasparus:
I've added a test, replaced `emotion/core` with `emotion/react` in the keyframes guide in the docs, and drafted a bit for the changelog on [test-components-ref-type branch](https://github.com/system-ui/theme-ui/tree/test-components-ref-type).

fcisio:
> Hey @dcastil, is there a way I can DM you? Twitter, email, or sth like that?
> 
> I'm trying to set up some means of communication between frequent contributors (AKA the place to yell "@hasparus, you promised me that code review a week ago"), and I'd like to send you an invite link :)

You should look into creating a [Discord Server](https://discord.com/)!
It's free and it's as complete as Slack. Others like Gatsby and Storybook are on there, but you can make it as small and contained or as big as you want.

Hope you find this useful 🎉

hasparus:
Nice idea, @fcisio!
Here's the invite https://discord.gg/2bJVCbuKwA

pkieltyka:
thanks all for putting this together, this will be a big improbement to theme-ui. Just my 2 cents -- I think either a release as 0.4.0-rcX or 0.5.0 is fine. 0.5.0 is probably safer, but I also think any user on a "rc" release understands it beta-ware and can also pin to a specific version for their needs. As well, the upgrade path from emotion 10 to emotion 11 for an app is quite simple -- to give an example my app with ~300 components upgraded without any changes going from emotion 10 to 11. The only changes I had to make was updating my import statements and the Theme type: https://emotion.sh/docs/emotion-11

hasparus:
Don't fix conflicts, I'll merge it in a bit :)

PR #1267, opened 2020-11-18
feat(color-modes): Add configuration option for printing color mode

### LINKED ISSUE #1144: Request for configuration option for color mode on printing
When printing out a webpage as a PDF or on paper, there’s no way to have Theme UI automatically apply a certain color mode.

Just like `initialColorModeName`, I’d like there to be a new configuration option `printColorModeName`, which applies that color mode inside `@media print`.

Haven’t considered alternatives, this seems like the easiest approach.

**Additional context**
It’s difficult right now to add great printing styles to Theme UI sites, especially when they’re in dark mode. You end up with needing global CSS with lots of `!important` to make everything the right colors.

lachlanjc:
Closes #1144, adding a configuration option `printColorModeName` for setting a color mode to be used in `@media (print)` (so users don’t print out in dark mode). By using the media query & custom properties, there’s no client-side JavaScript event listeners, so it should work even on a static-exported site.

PR includes updated TypeScript type, documentation, & two tests.

### DIFF
--- a/packages/color-modes/src/custom-properties.ts
+++ b/packages/color-modes/src/custom-properties.ts
@@ -13,6 +13,7 @@ const numberScales = {
 const reservedKeys = {
   useCustomProperties: true,
   initialColorModeName: true,
+  printColorModeName: true,
   initialColorMode: true,
   useLocalStorage: true,
 }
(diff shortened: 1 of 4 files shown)


lachlanjc:
Looks like the failing tests are unrelated to the changes & have been failing since at least #1249 :/

hasparus:
Brilliant

hasparus:
> Looks like the failing tests are unrelated to the changes & have been failing since at least #1249 :/

Yup, it seems we broke it while fixing the docs build, I'll handle it.

packages/color-modes/src/custom-properties.ts
```diff
@@ -78,15 +79,24 @@ export const createColorStyles = (theme: Theme = {}) => {
...
     styles[key] = objectToVars('colors', modes[mode])
   })
 
+  if (printColorModeName) {
+    const mode =
+      printColorModeName === 'initial' ||
+      printColorModeName === initialColorModeName
+        ? colors
+        : modes[printColorModeName]
+    styles['@media (print)'] = objectToVars('colors', mode)
```
hasparus, line 97: @lachlanjc shouldn't this be `@media print`? (without parens) I just noticed it doesn't work on Brave/Linux.

lachlanjc: Yep, you're totally right. I always forget about that since it's the only media query I use without parentheses

Issue #1234, opened 2020-11-02
babel-plugin-theme-ui


atanasster:
This is an umbrella issue to discuss specifications and work on theme-ui theme transformations with a babel-plugin (or webpack loader)

initial POC implementation:
https://github.com/atanasster/babel-plugin-theme-ui

atanasster:
@fcisio in #1205

So I could do something like this:

export const theme = {
  colors: {
    primary: '#FFFFFF',
    secondary: '#000000',
  },
  input: {
   '--bg': 'colors.primary',
   '--bg-hover': 'colors.secondary'
  },
};
As for my suggested approach, here's an example of the source and output, just in case it can be helpful.

export const theme = {
  colors: {
    primary: '#FFFFFF',
    secondary: '#000000',
  },
  input: {
   bg: 'primary',
   fontFamily: 'heading',
   '--bg-hover': 'secondary',
   '--color-hover': 'primary',
   '--bg-random': 'secondary',
   '--color': 'primary',
   '--fontFamily': 'heading',
   '--space-gap': [2, 3, 4],
  },
};

atanasster:
A few notes:

```
input: {
 '-- fontFamily': 'heading',
  '--bg-hover': 'secondary',
```
does not exactly work, since `heading` is a key in the `fonts` and `secondary` is a key in the `colors` section.

What currently will work is to specify the full key path:
```
input: {
 '-- fontFamily': 'fonts.heading',
  '--bg-hover': 'colors.secondary',
```

This could be made simpler but we will risk name collisions, for example both `fonts` and `colors` can have a key `body`

atanasster:
@fcisio example was very useful, I have it transformed into a test spec:

https://github.com/atanasster/babel-plugin-theme-ui/blob/master/tests/fixtures/custom-keys

```
export const theme = {
  space: [0, 4, 8, 16, 32, 64, 128, 256, 512],
  fonts: {
    heading: 'heveltica, arial, sans-serif',
  },
  colors: {
    primary: '#FFFFFF',
    secondary: '#000000',
  },
  input: {
   bg: 'colors.primary',
   fontFamily: 'fonts.heading',
    '--bg-hover': 'colors.secondary',
    '--color-hover': 'colors.primary',
    '--bg-random': 'colors.secondary',
    '--color': 'colors.primary',
    '--fontFamily': 'fonts.heading',
    '--space-gap': ["space.2", "14px", "space.3", "space.4"],
  },
};
```
[...]

PR #1271, opened 2020-11-18
fix: css custom vars recursive reference

atanasster:
issue: #1234

- remove default color value in css var invokation: 
component styles:  
`background-color: var(--theme-ui-colors-primary, #3333ee);` becomes now `background-color: var(--theme-ui-colors-primary);`

- css vars definitions now reference the actual color value instead of recursively reference themselves: 
body styles
`--theme-ui-colors-primary:  var(--theme-ui-colors-primary, #3333ee);` becomes now `--theme-ui-colors-primary: #3333ee;`

Run example for `<Button />` and observe color props in dev tools: https://theme-ui-draft.netlify.app/components/button


fcisio:
Hey @atanasster 
I forked this PR and tried some things!

I think we don't need to _"remove default color value in css var invocation"._
> If we remove the default value (fallback), we lose support for some browsers. And keeping it doesn't affect the end result. As long as the variable has a valid definition, that fallback will not be used.

The only necessary changes are those regarding the body styles _"css vars definitions now reference the actual color value instead of recursively reference themselves"_.

So we should only commit the changes on `packages/color-modes/src/index.tsx`

atanasster:
Thanks @fcisio, i will remove those. Initially i put them to avoid dom change exactly shen switching dark mode - and in theory without the fallback values it it should have less of a flash. But i see what you mean, even though i don’t understand why it happens :)

atanasster:
@fcisio - I reverted the css variables fallback values, but unfortunately still see the flicker:

https://theme-ui-draft.netlify.app/components/button/

Not sure what's causing it

atanasster:
Ok - fixed it. The issue with the flicker was actually not caused by the fallback values but using the color and background values as keys into the theme colors (which are later transformed to css vars but on the original theme). 

I now changed it so color and background are using css vars names. Here is the latest build: https://theme-ui-draft.netlify.app/components/button/

I also tried to go back to the initial solution without using fallbacks to avoid dom updates when switching themes/dark modes, just so we can get a feeling of any perceived performance improvement: https://theme-ui-css-vars.netlify.app/components/button/ 

Anyone up for some performance measurements?

hasparus:
nice work!

atanasster:
thanks @hasparus - whats your take on using the fallback values ? They should really not be needed - they are actually taken from the same color keys as where the variables are defined, so we should never have a discrepancy. 

However it could introduce some incompatibilities - maybe we remove the fallback values in a later update?

hasparus:
I think I lack context here. I don't know the reason for including theses fallback values.

lachlanjc:
Yeah, I've always wondered about this. Have they been there since the beginning or were they added later?

fcisio:
I thought that these fallbacks were used if Custom Properties were not supported, but they're actually only used if the variable is not found.

In our case, the variable will never "not be found". So it would be safe to just remove those for performance sake.

hasparus:
> In our case, the variable will never "not be found". 

uhh... what if you... copy your rendered HTML node to an iframe? :D Wouldn't this be the case?

Are these fallback a performance problem? @fcisio

fcisio:
> uhh... what if you... copy your rendered HTML node to an iframe? :D Wouldn't this be the case?

Not sure I understand but yeah, should still work.

In this example, the fallback would be used:

```css
.element {
  color: var(--color-red, red);
}

body {
  --color-blue: blue;
  --color-yellow: yellow;
  --color-green: green;
  /* --color-: red; */
}
```

Since the colors are generated and set by the theme, we shouldn't have to worry about missing variable declarations.

As for performances, I don't know just how much more performant it is. But essentially:
- [With fallback](https://theme-ui-draft.netlify.app/components/button/):
On color mode switch, all the elements with colors have to render new classNames

- [Without fallback](https://theme-ui-css-vars.netlify.app/components/button/): 
On color mode switch, elements with colors don't have to rerender since the value is the same.

You can compare the two and see what I mean.

hasparus:
> Since the colors are generated and set by the theme, we shouldn't have to worry about missing variable declarations.

So, with fallback I can safely move my rendered elements outside of the DOM to another document (in an iframe).

I was really fishing for any reason to keep the fallback, and this is the only one I found.
I don't think that usecase would be common. 

We could make this configurable for backwards compatibility.

atanasster:
@hasparus - I think you would still need to have theme ui provider at the root. 
Can you show me an example of the use case that you mention above (move just the rendered components to a separate iframe)?

As for backward compatibility, the fallback props were used just because there was a bug in the vars declarations themselves. 

In any case, the current implementation is with the fallback props, so we can merge it and keep the discussion open for removing the fallback options.

hasparus:
> @hasparus - I think you would still need to have theme ui provider at the root.

If I render something with fallback values, I could move it anywhere else, right?...

Okay, my counterexample is too weak. I'd have to copy classnames to the iframe (tags from head), so I might as well add a provider. I don't have any other idea for defense of the fallbacks. 

> In any case, the current implementation is with the fallback props, so we can merge it and keep the discussion open for removing the fallback options.

It's already merged :) We're exactly at the point in time you mentioned.

atanasster:
Ah very cool, I am a bit behind the last couple of days :)

Then another poll, or advisory board call? I would say remove the fallbacks while we are still in the 0.xxxx versions and see if anyone has use-cases to add back the fallbacks?

hasparus:
> I would say remove the fallbacks while we are still in the 0.xxxx versions and see if anyone has use-cases to add back the fallbacks?

Reasonable.

atanasster:
> Reasonable.

Great, I will push a new PR for it

hasparus:
Okay, I found the problem.
This PR breaks `@theme-ui/color`.

Issue #1079, opened 2020-07-22
transitions token support


pacxiu:
Hey,

do you have in plans to support `transitions` tokens such as other base one?

For now if I've some repeating transition eg. `0.3s ease-in-out` and want to use that outside of theme file in `sx` prop I've to add it as string rather than `transition: 'standard'` token one. Are there maybe some ways to overcome this that I might have missed?

lachlanjc:
We can close now that it’s merged @hasparus!

PR #1272, opened 2020-11-19
feat: add support for transitions token

atanasster:
issue: #1079

added support for transitions in theme sclaes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.4.0 UNRELEASED
 
+-- Add support for transitions token. issue: #1079 
 ## v0.4.0-rc.12 2020-11-18
 
 - Use correct version of @theme-ui/components in theme-ui package.
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -65,6 +65,7 @@ export const scales = {
   caretColor: 'colors',
   columnRuleColor: 'colors',
   opacity: 'opacities',
+  transition: 'transitions',
   margin: 'space',
   marginTop: 'space',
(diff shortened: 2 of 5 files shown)


lachlanjc (approved):

fcisio:
Hey @atanasster 
This enhancement made me realize that your `babel plugin` could solve future similar needs.

> 🤔  I wonder if that's a pattern that we want to allow or not. 

Tested it and it currently works! One could simply create a new scale, named anything they want, and then reference it in the variant.

```js
space: [0, 4, 8, 16, 32, 64, 128, 256, 512],
speed: ["0.35s", "0.5s", "0.75s"],
ease: {
  in: "ease-in",
  out: "ease-out",
  inOut: "ease-in-out",
},
colors: {
  text: "#000",
  background: "#fff",
},
```

This is how I personally manage transitions in the theme.

> By setting the values on CSS variables, you can better control the output. Plus this means that the transition property doesn't need to be in the variant. It could very well be in straight Emotion or in the `SX` prop.

```js
a: {
  "--speed": "speed.1",
  "--ease": "ease.out",

  color: "primary",
  transition: "color var(--speed, 0.35s) var(--ease, ease)"
},
```

atanasster:
@fcisio - oh thats very nice. Are you able to already use the plugin for your app, or just testing it out? I know the colors cs custom vars still need the fix, but the rest should be ok.

fcisio:
@atanasster I'm just testing out right now.
And yep! My plan is to use everything, then start using the colors once they're fixed!

hasparus (approved):
Atanas ♥

Issue #1027, opened 2020-06-25
Values of space keys when referenced in variants are not populated in CSS


mirabilio:
**Describe the bug**

Create an array assigned to a `space` key for responsive values (called "lg" in this issue). Use that key in a variant for a property. Theme-ui does not populate those values in the CSS as expected. Using `[...theme.space.lg]` does work, however. 

**To Reproduce**
1. Use this theme:
```
theme.space = [
  "0", 
  "0.25rem",
  "0.5rem", 
  "0.75rem",
  "1rem",
  // ...
];
theme.space.sm = [1, 2, 3, 4, 5]; // step 1
theme.space.md = [2, 4, 6, 8, 10]; // step 2
theme.space.lg = [3, 6, 9, 12, 15]; // step 3
theme.space.xlg = [4, 8, 12, 16, 19]; // step 4

theme.people = {
  mx: "lg", // margin-right and margin-left are not present in CSS. However, [...theme.space.lg], DOES work. 
}
```
2. Render a component a la:
```
const People = ({people}) => {
  // [...]
  return (
    <Grid sx={{ variant: "people" }}>
      {peopleList}
    </Grid>
  )
}
```
**Expected behavior**
`margin-right` and `margin-left` values should be populated in `People` component's CSS like so:
```
@media screen and (min-width: 1366px)
.css-6857ss-People {
    grid-auto-rows: min-content 1fr;
    grid-template-columns: 1fr 2fr;
    margin-left: 12rem;
    margin-right: 12rem;
    grid-gap: 1.25rem;
}
@media screen and (min-width: 1024px)
.css-6857ss-People {
    grid-auto-rows: min-content 1fr auto;
    grid-template-columns: 1fr;
    margin-left: 6rem;
    margin-right: 6rem;
    grid-gap: 1rem;
}
// [...]
```
The above CSS results when using `theme.people` definition:
```
theme.people = {
  mx: [...theme.space.lg]
}
```
**Actual behavior**
`margin-left` and `margin-right` are missing:
```
@media screen and (min-width: 1366px)
.css-149upde-People {
    grid-auto-rows: min-content 1fr;
    grid-template-columns: 1fr 2fr;
    grid-gap: 1.25rem;
}
@media screen and (min-width: 1024px)
.css-149upde-People {
    grid-auto-rows: min-content 1fr auto;
    grid-template-columns: 1fr;
    grid-gap: 1rem;
}
// [...]
```

hasparus:
[...]

Issue #720, opened 2020-02-24
Responsive values is variant styles override sx styles


rpavlovs:
Responsive values in variant styles override `sx` styles.

**To Reproduce**
```js
// theme.ts

variants: {
  buttons: {
    primary: {
      paddingX: ['24px', '32px'],
      paddingY: ['8px', '12px'],
	}
  }
}
```
```js
// component.styles.ts

// twitter button is a primary button

twitterButton: {
	variant: 'button.primary',
    px: '5px',
    py: '5px'
}
// final style: => {px: '24px', py: '8px'}
// expected: { px: '5px', py: '5px'}

twitterButton: {
	variant: 'button.primary',
    px: ['5px'],
    py: ['5px'],
}
// final style: => {px: '24px', py: '8px'}
// expected: { px: '5px', py: '5px'}

twitterButton: {
	variant: 'button.primary',
    px: ['5px', '5px', '5px'],
    py: ['5px', '5px', '5px'],
}
// final style: => {px: '5px', py: '5px'}
// as expected 
```

**Expected behavior**
`sx` prop styles should always override variant styles.

jxnblk:
Thanks! I'm not sure I understand the issue here. Would you mind creating a Code Sandbox to demonstrate what's happening?

lachlanjc:
Pretty sure I’m having the same issue: My `cards.primary` defines `p: [2, 3]`. To set padding to 0, I have to do `sx={{ p: [0, 0] }}`—just `p: 0` only applies on mobile.

jxnblk:
@lachlanjc if that's the same thing happening in the issue description, then yeah, that's a known issue without a clear solution, but I suspect the example above might be pointing to something else??

rpavlovs:
Yeah, it’s the same issue. Thanks for a better explanation @lachlanjc

kensongzhu:
Just start using theme-ui and hit the same issue 
I reproduced the codesandbox: https://codesandbox.io/s/trusting-matsumoto-ebnc5

It seems related to the [Box](https://github.com/system-ui/theme-ui/blob/fbdd88ea280c0d9880d39a114cb6d19f0d4da2ca/packages/components/src/Box.js#L12). The system props styles of `base`, `variant` and `sx`
are generated seprately and then merged. 

In other words, e.g. whatever media queries generated in `variant` is taken anyway and won't be **removed** with `sx`, right?  

If I twist the code a bit as following, does it make sense?
```js
const aliases = {
[...]

Issue #1030, opened 2020-06-26
Variant responsive values not working alongside other responsive values


julianbenegas:
**Describe the bug**
You have a component with the `variant` property inside `sx`. This `variant` holds a responsive value, say `fontSize: ["12px", "16px"]`. If you then put another responsive value to the `sx`, then the variant's responsive value won't work.

**To Reproduce**
1. Go to [this sandbox](https://codesandbox.io/s/nameless-sun-hspem-hspem?file=/src/App.js)
2. Resize the window
3. See that the text doesn't change size

**Expected behaviour**
The text should change it's fontSize

**Additional context**
It happens when using variant variant inside `sx`. If variant is used as a prop, it works as expected.

jxnblk:
Thanks! This looks like an odd bug... possibly related to #720 and #832

PR #1273, opened 2020-11-21
fix: sx variant responsive

atanasster:
issue: #1030

when a variant was used as an sx prop, its values were merged **after** applying responsive mods. This PR moves the variant props merging **before** applying responsive.

The original issue: https://codesandbox.io/s/nameless-sun-hspem-hspem?file=/src/App.js
After the fix (slightly modified): https://theme-ui-responsive-variants.netlify.app

I have seen some other similar issues related to responsive if you guys can check if they are being addressed with the PR @lachlanjc @hasparus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.5.0 UNRELEASED
 
+- Fix sx prop variant responsive. Issue: #1030
 ## v0.5.0-alpha.0 2020-11-20
 
 - BREAKING: Upgrade to Emotion 11, and `csstype` 3. PR #1261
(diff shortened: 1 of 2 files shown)


lachlanjc:
Amazing! Does this fix #720 as well?

atanasster:
@lachlanjc -- I am not sure. This fixes when the variant is part of the sx prop 

`<Box sx={{ variant:'xx', p:[1, 5] }} />` - this affected (fixed)

`<Box variant="xx" sx={{ p:[1, 5] }} />` - this NOT affected (istm that it was working fine before)

lachlanjc:
Got it. Do we close #1027 with this?

PR #1274, opened 2020-11-21
(docs): Add note about using Components’ as prop with TypeScript

### LINKED ISSUE #1124: `as` prop types on Box are not supported in TypeScript
**Describe the bug**
While using typescript, the svg types are not recognized.

**To Reproduce**
1. Import the Box component
2. Use the `as` prop with value "svg"
3. Try to use any `<svg>` specific prop

**Expected behavior**
The types must extend the native svg ones.

**Screenshots**
Screen Shot 2020-08-11 at 16 39 27

**Additional context**
[Sandbox Link](https://codesandbox.io/s/bug-box-svg-tsx-kpdo8)

lachlanjc:
Closes #1124. I don’t have much TypeScript experience, so I hope I got the wording right here :)

### DIFF
--- a/packages/docs/src/pages/components/index.mdx
+++ b/packages/docs/src/pages/components/index.mdx
@@ -7,6 +7,8 @@ import { Note } from '../..'
 # Components
 
 A number of built-in UI components are available for layouts, grids, buttons, form elements, and more.
+Components can be used as an alternative to the [JSX pragma](/jsx-pragma) for using Theme UI features,
+or in parallel to provide specific UI components.
 
 ```jsx live=true
 <Box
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/components/index.mdx
```diff
@@ -95,6 +97,14 @@ The `sx` prop on Theme UI components can be used *without* the `jsx` pragma comm
 
 </Note>
 
+<Note>
+
+If you’re using TypeScript, Theme UI merges prop types with the props of components passed in the `as` prop.
```
hasparus, line 102: > Theme UI merges prop types with the props of components passed in the `as` prop.

We do?

I remember I just surrendered with Box for `@types/theme-ui__components`, because supporting `as` prop on Box hurt the performance so much I couldn't pass DefinitelyTyped tests (and VSCode was starting to behave funny).

Box is a styled component, so it was already quite heavy on TypeScript, and adding `as` was the last straw that breaks the back.

We only support `as` in TypeScript on `Field`, where it's crucial.

image

hasparus: We could support `as` if we got rid of styled-system, and emotion/styled (so `as` instead of `withComponent`).

We could deprecate `withComponent` first, though.

lachlanjc: Yep you're totally right, I misread a sentence in https://github.com/system-ui/theme-ui/issues/1124#issuecomment-672719902

Will take a look at this later today!

hasparus: Thanks!

PR #1276, opened 2020-11-21
(docs): Clarify usage of child/pseudo-selectors in sx

### LINKED ISSUE #940: Add guide for using child selectors in sx prop
There are many features of Emotion that Theme UI inherits by default, but that might not be clear to users of this library on its own. Child selectors is one of those features. Although general usage should be discouraged, a guide to using child selectors when working with non-React code, external components that do not accept the `className` prop, or other edge cases where child selectors offer a work-around would be good to add to the docs site.

### LINKED ISSUE #1074: Specifying a CSS selector
Is there a way to specify a CSS selector (e.g. `:link` or `:visited`). If not, is this of your interest as a user?

Currently, I am using the following in my global css:
```css
:link {
  color: white;
}
```
I would like this to be part of the theme-ui config instead.

lachlanjc:
Clarifies how you use (including nested) child selectors with sx/Emotion, as well as psuedo-selectors. Also includes a note on using `BaseStyles` in this section since I think it’s relevant.

Closes #1074 
Closes #940

### DIFF
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -2,6 +2,7 @@
 title: 'The `sx` Prop'
 ---
 
+import { Note } from '..'
 
 # The `sx` Prop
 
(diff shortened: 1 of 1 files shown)


hasparus (approved):
LGTM

PR #1277, opened 2020-11-21
(docs): Clarify merging theme values from multiple files

### LINKED ISSUE #359: Guide request for breaking longer theme files into separate components
Hello!

Amazing to see the docs expand and all the content you have created for this since it launched!  Thank-you for the hard work on this - it makes using theme-ui so much better. As my theme-ui `index.js` file gets longer and more complicated with variants and styles I have tried a few times to split it into separate files using `deepmerge`. I keep hitting errors when I do this and just revert back to my big `index.js` file as this works fine for me with no problems.  My hope in breaking it into smaller pieces is to make it easier to shadow in gatsby themes and also just to make it simpler for editing/management going forward...ignore the fact that I haven't setup typography.js yet, I really should be using that instead of a bunch of `fontSizes`, `fontWeights`, etc.

My index.js file when broken down looks something like this:
```
import merge from "deepmerge"
import colors from "./colors"
import fonts from "./fonts"
import fontSizes from "./fontSizes"
import fontWeights from "./fontWeights"
import lineHeights from "./lineHeights"
import spaces from "./spaces"
import sizes from "./sizes"
import breakpoints from "./breakpoints"
import styles from "./styles"
import variants from "./variants"

export default merge(
  colors,
  fonts,
  fontSizes,
  fontWeights,
  lineHeights,
  spaces,
  sizes,
  breakpoints,
  styles,
  variants
)
```

This seems to work with deepmerge but then the following code, which was working fine before with one big `index.js` starts failing...using the `sx={{}}` object.

```
gridTemplateRows: [
          theme => theme.sizes.headerHeight,
          theme => theme.sizes.headerHeightTablet,
          theme => theme.sizes.headerHeightLaptop,
        ],
```

It no longer seems to be able to reader those values which is odd to me.  I am sure there is a reasonably easy way to do this and I imagine other gatsby theme others are thinking similar things as their variants usage starts to expand and grow leading to big `index.js` files.

### LINKED ISSUE #885: Multiple files for themes
This is more of a documentation suggestion (i've not tried to edit it directly as guidance may be needed)

Following this doc page I initially had trouble splitting my theme into multiple files
https://theme-ui.com/guides/merging-themes/

If I wanted a file for all my typographic stuff, one for all colors and modes, and perhaps one for all the styles I am faced with including nested objects, or even single whole objects. The guide has examples where the file itself imports a single simple object.

So if my colors file looked like this:
```javascript
export default {
  colors: {
    text: "#000",
    background: "#fff",
    primary: "#07c",
    modes: {
      dark: {
        text: "#fff",
        background: "#000",
        primary: "#0cf",
      },
    },
  },
}
```

to import it I would need to spread the imported object like this:
```js
import colors from "./colors"
export default {
  ...colors,
}
```

This might be something obvious to many people, but it wasn't to me at first. This method does suit the way that many people might organise their stylesheets, into areas of concern, rather than very simple small files.

If people think this is a useful edition into the docs I can fork and add something.

lachlanjc:
Closes #885 
Closes #359

### DIFF
--- a/packages/docs/src/pages/guides/merging-themes.mdx
+++ b/packages/docs/src/pages/guides/merging-themes.mdx
@@ -13,7 +13,6 @@ This guide shows a few common ways to merge themes together.
 To use a preset as the basis for a custom theme, it's recommended that you use a deep merge utility.
 The `theme-ui` package exports a preconfigured version of the `deepmerge` package that can be used for this.
 
-
 ```js
 // example theme based on preset-future
 import future from '@theme-ui/preset-future'
@@ -59,4 +58,23 @@ export default {
 }
 ```
 
+## Merging theme values from multiple files
 
+For the core scales (`colors`, `fontSizes`, `space`, etc), it’s typically simpler to understand a theme when values are defined in one file.
+However, especially for projects with many [styles](/styling-mdx) or [variants](/guides/variants), you may want to split values that end up
+as part of one object across multiple files. You can use the same kind of imports combined with the `merge` utility for this.
+
+```js
+// example theme/index.js
+import { merge } from 'theme-ui'
+import colors from './colors'
+import fonts from './fonts'
+import textStyles from './styles/text'
+import tableStyles from './styles/tables'
+
+export default {
+  colors,
+  fonts,
+  styles: merge(textStyles, tableStyles)
+}
+```
--- a/packages/docs/src/pages/guides/merging-themes.mdx
+++ b/packages/docs/src/pages/guides/merging-themes.mdx
@@ -71,10 +71,13 @@ import colors from './colors'
 import fonts from './fonts'
 import textStyles from './styles/text'
 import tableStyles from './styles/tables'
-
+import { buttons, forms, links } from './variants'
 export default {
   colors,
   fonts,
-  styles: merge(textStyles, tableStyles)
+  styles: merge(textStyles, tableStyles),
+  buttons,
+  forms,
+  links
 }
 ```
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/guides/merging-themes.mdx
```diff
@@ -59,4 +58,23 @@ export default {
...
+as part of one object across multiple files. You can use the same kind of imports combined with the `merge` utility for this.
 
+```js
+// example theme/index.js
+import { merge } from 'theme-ui'
+import colors from './colors'
+import fonts from './fonts'
+import textStyles from './styles/text'
+import tableStyles from './styles/tables'
+
```
hasparus, line 74: What do you think about adding some more here? I'd like to show that shorthand property names and object spreading _still work_ like with any other object.

```suggestion
import { buttons, forms, links } from './variants'
```

packages/docs/src/pages/guides/merging-themes.mdx
```diff
@@ -59,4 +58,23 @@ export default {
...
+export default {
+  colors,
+  fonts,
+  styles: merge(textStyles, tableStyles)
```
hasparus, line 78:
```suggestion
  styles: merge(textStyles, tableStyles),
  buttons,
  forms,
  links
```

hasparus (approved):

packages/docs/src/pages/guides/merging-themes.mdx
lachlanjc: Looks great, will implement!

PR #1279, opened 2020-11-22
(docs): Update sx page with full properties list

lachlanjc:
We certainly lose some readability with this change, making it harder to digest the list & find the most common ones since there are just so many properties. If anyone has suggestions for shortening this or an elegant way of truncating some (the block/inline/inset properties are not used by a lot of folks afaik), that’d be very welcome, but I think we should give a thorough list somewhere.


hasparus:
It could be folded with `<details />` or grouped by scale, if you're lookig for ideas.

hasparus (approved):

PR #1281, opened 2020-11-23
chore(deps-dev): Drop unused repng dependency on docs

lachlanjc:
`repng` is only used via `npx`, & has a ton of dependencies itself, so this will make initially installing deps on Theme UI a lot faster.

Closes #984


hasparus (approved):
Great idea.

PR #1292, opened 2020-11-23
chore: remove deprecated devtools from workspaces

hoobdeebla:
Moves the theme-ui devtools extension from `packages` to `deprecated`. Simplifies dependency management and reduces monorepo node_modules size.


lachlanjc (approved):
SGTM!

Issue #1250, opened 2020-11-10
'Text' component to use 'p' or span as default


isaac-martin:
Have just picked this library up on a project, and so far so good. Except I worry about the fact that the default HTML element that is used for the `<Text>` primitive is a `div` - would there be pushback for me to PR to swap this to a `p` or `span` and add an `as` prop to change the base renderer?

lachlanjc:
We have the `as` prop already! But yeah, a `<p>` tag by default would be awesome, I don’t know why we haven’t changed that already.

atanasster:
Changing it to a default `<p />`will be a breaking change/different margin defaults, right?

lachlanjc:
Yep, definitely a breaking change. Also if you're styling children via selectors etc. Would make more semantic HTML by default though.

atanasster:
I think it would be correct for Text to have been a `<span />` and a separate Paragraph component

lachlanjc:
Nonetheless, span is inline by default while p is block by default, so definitely still a breaking change

isaac-martin:
> Changing it to a default `<p />`will be a breaking change/different margin defaults, right?

We _may_ be able to default the margins / spacings to be the same as div - maybe making it a span so that it is inline by default.

atanasster:
I dont know if we should introduce a breaking change, but if we do, i vote for `<span />` :)

flo-sch:
Might be just me, but I almost use `<Text as="p" />` for actual HTML paragraphs, and very rarely need an inline element

Both the breaking-change and having another "Paragraph" component defaulting to a "p" sound good to me

atanasster:
@flo-sch - would you want to contribute a `<Paragraph />` component?

hasparus:
@flo-sch @atanasster do we need a Paragraph component? I mean, would it be different to `<Text as="p" />`?

atanasster:
@hasparus i think we need a Paragraph - it will have its own theme key.
Most component libraries i have used before theme-ui do have a Text, which is a span and a separate Paragraph component, so it will be an easier transition as well.

lachlanjc:
Interesting. Yeah, I very occasionally use Text inline but it's almost always a paragraph, & I sometimes forget to include the `as`. I'm in favor of making Text return a `p` by default, since that's what I've always expected anyway, but the theme keys are a totally valid reason for a separate component too.

atanasster:
[...]

PR #1298, opened 2020-11-25
feat: paragraph component

flo-sch:
Relates to #1250

Add a new `<Paragraph />` component, defaulting to HTML `<p>` element (getting the semantics).
Also add a new doc page for it.


flo-sch:
First try, let me know what you guys think? @atanasster @isaac-martin @hasparus 

Few notes:
- `<Text />` remains unchanged, since converting it to a span would be a BC-breaking change, potentially for a separate PR, potentially later (not sure which release this would be included in)?
- No different themeKey **yet**, since `<Heading />` does not have one neither, I am wondering about whether this would be confusing?

packages/docs/src/pages/components/paragraph.mdx
```diff
@@ -0,0 +1,52 @@
+---
+title: Paragraph
+---
+
+# Paragraph
+
+Primitive typographic component for text blocks, defaults to `<p>`, and without margin.
```
hasparus, line 7: Nitpick. There's something not right with this sentence.

hasparus (approved):
LGTM, I'll wait for second opinion for a bit, but it's okay to merge I think

atanasster:
Looks great - just a few small items
- I would suggest to have a separate theme key (variant) `paragraph` for `<Paragraph />` inside the text space
- I have seen Paragraph in some libs to have a default max-width, or default margin-bottom values (tailwind-css, grommet). We can keep for now away from such defaults, but having a separate theme variant key would allow to separate styling from the `<Text />`

hasparus:
> I have seen Paragraph in some libs to have a default max-width, or default margin-bottom values (tailwind-css, grommet). We can keep for now away from such defaults, but having a separate theme variant key would allow to separate styling from the <Text />

Yes, please. Default max-width. We can set something simple there like `760px` to avoid resize on font load, but let's set it to set up a precedent. If somebody want full page wide paragaphs, they can suffer.

We use the last value from `sizes` scale, multiply it by two, whatever, but I'm strongly in favor of setting that max-width.

flo-sch:
Alright, the reason for using the same themeKey is the consistency with Heading: why would heading variants be based on theme.text when paragraphs are separated?

But if you think it is better, I will add it

atanasster:
@flo-sch - sorry, I meant the variant (default vs paragraph) - and I called it incorrectly `theme key`, but not the _themeKey prop. Exactly like Heading.

flo-sch:
Ah alright, yes no problem with a default `paragraph` variant, that makes sense :)

atanasster:
Tailwindcss also resets margins to 0 for paragraph (i think it comes from normalizecss)

flo-sch:
Box already does that, that's why I did not, but perhaps it is good not to rely on inherited here?

atanasster:
I think users are complaining Box does it, since its supposed to be non-opinionated.

Its up to you though, i am just drafting suggestions  as they come to mind. Your PR looks fantastic in any way.

packages/docs/src/pages/components/paragraph.mdx
flo-sch: Something like my frenglish, perhaps? 😁

flo-sch:
I have added those defaults as part of `sx` props to make sure anyone can override them (if they want to), also with a hardcoded MQ (since I thought it is better not to rely on breakpoints...? Not too sure myself about that anyway, let me know what you think?)

atanasster:
excellent, lgtm

flo-sch:
Sweat :)

I have no clue how the release works here, so I would rather have someone knowing it merging the PR whenever they think it's time

Do not hesitate if you feel something should be added?

PR #1303, opened 2020-11-25
Center the logo

hasparus:
I feel a bit guilty about making a PR out of this.


lachlanjc (approved):
You're good! I think no PR is too small…we're here not to make everything immediately perfect, but incremental correctness & successive improvement. Small PRs give us all momentum & make more things better sooner, so go for it.

Issue #1307, opened 2020-11-27
0.5.0-alpha.1 - Property 'sx' does not exist on type 'IntrinsicAttributes & ...


atanasster:
Since upgrading 0.5.0-alpha.1, I see this error on external(non-theme-ui) components

```
Property 'sx' does not exist on type 'IntrinsicAttributes & ...
```

example:

https://github.com/ccontrols/component-controls/blob/0335fb94440b8dc67a4b00c7db0560a2f42947aa/ui/components/src/CopyContainer/CopyContainer.tsx#L30

hasparus:
Okay, so I have one case of this bug, and I think I have a suspect. I didn't check if it's the same thing that happens in `component-controls`, and don't have a solution yet, but let me share my thoughts.

## my case and the investigation

```tsx
import { jsx, Styled as s } from "theme-ui";

export type BoxedTextProps = ComponentPropsWithoutRef<typeof s.p>;
```

![image](https://user-images.githubusercontent.com/15332326/100469410-780a2f00-30d6-11eb-85cd-2d0adedd0796.png)

My `BoxedText` components spreads all of its props onto `Styled.p`, so the type of its props is just the type of props of `Styled.p`.

Simple, right?

Well, no. I must have wrote it before `Styled.p` or, since we're speaking of types, `ThemedComponent<'p'>` (defined below) accepted `as` prop.

```tsx
export interface ThemedComponent<Name extends ElementType> {
    <As extends ElementType | undefined = undefined>(props: WithPoorAsProp<ComponentProps<Name>, As>): JSX.Element;
}

export declare type WithPoorAsProp<Props, As extends ElementType | undefined = undefined> = {
    as?: As;
} & (As extends undefined ? Props : AnyComponentProps);

interface AnyComponentProps extends JSX.IntrinsicAttributes {
    [key: string]: unknown;
}
```

`WithPoorAsProp` is a way to opt out of typechecking a component usage when you add `as={"something"}`.

But the problem with it is that when we read the props on type level, as becomes `ElementType | undefined`, and we get a bunch of props mapped to undefined.

![image](https://user-images.githubusercontent.com/15332326/100470731-14cdcc00-30d9-11eb-8aeb-9f153183ac23.png)


Okay, but it should allow any prop, right?

```tsx
declare type WithConditionalSxProps<P> =
    'className' extends keyof P
        ? P extends { className?: string; }
            ? Omit<P, 'sx'> & { sx: SxProps }
[...]

PR #1308, opened 2020-11-27
fix(types): add sx prop whenever className can be string

hasparus:
I'm not exactly sure if we have the best possible implementation, as we don't produce `className` at runtime, we produce `css` and call Emotion's `jsx`, so it seems that we should also use their types for correctness.

cc @atanasster @dcastil

Relevant to #1307, but I'm not sure if it closes.

---

There are three changes here.

> - **breaking TypeScript**: Renamed and removed types.
>   - `SxProps` to `SxProp`.
>   - `SxStyleProp`, an alias for `ThemeUIStyleObject` removed. Use `ThemeUIStyleObject` instead.
> - Fix: Add `sx` props types to all props accepting `className`.
> - Fix WithPoorAsProp to work with ComponentProps utility type.

The change to `WithPoorAsProp` allows to read props of `@theme-ui/mdx` `Styled` on type level. Previously, after passing throught ComponentProps, they became dicts of `unknown`.

```tsx
type MyProps = ComponentPropsWithoutRef<typeof Styled.p>
```

The change to `WithConditionalSxProp` allows to use `sx` with all components accepting `className` of type `string`, even if they also accept other values for `className`.


packages/core/src/jsx-namespace.ts
```diff
@@ -1,10 +1,8 @@
-import { SxProps } from './types'
+import { SxProp } from './types'
 
-type WithConditionalSxProps<P> = 'className' extends keyof P
-  ? P extends { className?: string }
-    ? Omit<P, keyof SxProps> & SxProps
-    : Omit<P, keyof SxProps>
-  : Omit<P, keyof SxProps>
+export type WithConditionalSxProp<P> = { className: string } extends P
```
dcastil, line 3: I like this one much more than the previous solution. I don't know why Emotion did it that way.

hasparus:
Okay, I got greedy, and there's going to be another change, what you can probably see in failed tests. I'm making `@theme-ui/core` and `theme-ui` strict TypeScript.

packages/core/src/jsx-namespace.ts
hasparus: Sorry, Dany. Now it's even uglier. I forgot that components can have more than one prop while writing this, and that's how I got that nice one-liner :D

dcastil: haha 😂

hasparus:
I didn't even touch this :cry: 
image

LibraryManagedAttributes can literally wreak generic components inference :D Okay, this needs to be fixed on our side. I don't want to add "our jsx pragma is incompatible with react-hook-form" to the readme :D 

image

hasparus:
cc @beerose

Issue #1313, opened 2020-11-30
gatsby-plugin-theme-ui@next errors with automatic JSX runtime


hoobdeebla:
Theme UI now supports the automatic JSX runtime as of `0.5.0-alpha.1`. Gatsby also supports it but currently has it disabled by default ([gatsby#27615](/gatsbyjs/gatsby/pull/27615)). When manually enabling it via a custom Babel config in my Gatsby site, the following error occurs on `gatsby develop` with `gatsby-plugin-theme-ui@0.5.0-alpha.1`:

```console
error undefined failed

[my project]/node_modules/gatsby-plugin-theme-ui/src/provider.js: pragma and pragmaFrag cannot be set when runtime is automatic.
> 1 | /** @jsx jsx */
    | ^
  2 | import { jsx, ThemeProvider, merge } from 'theme-ui'
  3 | import localTheme from './index'
  4 | import components from './components'
not finished Building development bundle - 1.398s
```

I also tried disabling gatsby-plugin-theme-ui and manually wrapping my site's layout component with `<ThemeProvider/>`, and the dev build completed; however, when I loaded localhost:8000, I received the following error:

```console
TypeError: Cannot read property 'variant' of null
(anonymous function)
node_modules/@theme-ui/css/dist/index.esm.js:276
  273 | var result = {};
  274 | var obj = typeof args === 'function' ? args(theme) : args; // insert variant props before responsive styles, so they can be merged
  275 | 
> 276 | if (obj['variant']) {
  277 |   obj = _extends({}, get(theme, obj['variant']), obj);
  278 |   delete obj['variant'];
  279 | }
```

Here is the custom Babel config I am using:
```json5
{
  presets: [
    'babel-preset-gatsby',
    [
      '@babel/preset-react',
      { runtime: 'automatic' },
    ],
  ],
}
```

lachlanjc:
This should be fixed on the latest (alpha) release. Can you try 0.6 which supports the automatic runtime/Emotion 11?

hasparus:
I didn't get to testing it yet, but it should be fixed in `0.6.0-alpha.3`.

hoobdeebla:
Yes, it is. Thank you!

PR #1321, opened 2020-12-01
fix: check if object is non-null before variant

atanasster:
issue: #1313

Seems it was possible to have null values for `obj` - added check

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -281,7 +281,7 @@ export const css = (args: ThemeUIStyleObject = {}) => (
   let result: CSSObject = {}
   let obj = typeof args === 'function' ? args(theme) : args
   // insert variant props before responsive styles, so they can be merged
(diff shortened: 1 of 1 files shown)


hasparus:
How do you feel about skipping the entire iteration when `x` is false or empty?

I have all type-level changes for it already in #1090, but we could merge this already as a step towards closing https://github.com/system-ui/theme-ui/issues/1297.

```tsx
  for (const key in styles) {
    const x = styles[key as keyof typeof styles]

    if (x === false || x == null) {
      continue
    }

```

atanasster:
> How do you feel about skipping the entire iteration when `x` is false or empty?

Hmm, its outside the iteration now to resolve the variants before the iteration

hasparus:
> Hmm, its outside the iteration now to resolve the variants before the iteration

Oh. True! I misread the code. Sorry!

PR #1322, opened 2020-12-01
(dev): Set arrowParens in Prettier config

lachlanjc:
Since this default is different across various versions of Prettier, we’ve all had a bunch of commits going both ways across various files. Setting this in our Prettier config so our diffs are smaller/easier to read/more consistent.


hasparus:
YES. It was always suspicious to see so many changes to these parens...

PR #1323, opened 2020-12-01
(mdx): Rename Styled component to Themed

lachlanjc:
Following #1312, making this breaking change to use a clearer name for `Styled`. I did my absolute best to make the change everywhere across docs, code, docs code, & examples, & I added a redirect on the docs site from /styled to /themed. Lmk if anything looks off!


hasparus (approved):
LGTM

Issue #1319, opened 2020-11-30
bugs from development docs


hasparus:
1. It seems that something (#1273?) broke how variants compose with mdx styles, what's visible in docs home page.
2. Did `@theme-ui/prism` stop working or did we break something in the docs?

lachlanjc:
for context:

![image](https://user-images.githubusercontent.com/5074763/100681682-45fd0500-3342-11eb-9aa4-eaedaf590dd4.jpeg)

![image](https://user-images.githubusercontent.com/5074763/100681692-4b5a4f80-3342-11eb-8709-db2c89721258.jpeg)

atanasster:
I'll take a look tonight, seems something is off

atanasster:
@hasparus - was there something else besides the Prism code highlighting?

hasparus:
There are link buttons on home page.
![image](https://user-images.githubusercontent.com/15332326/100725016-ce6fba00-33c3-11eb-8c28-28459b4b261f.png)
![image](https://user-images.githubusercontent.com/15332326/100725065-db8ca900-33c3-11eb-9402-76636b0e65b1.png)

I fiddled with it for a while, but this is quite tricky. The color from variant should override color from anchor styles.

atanasster:
Thanks @hasparus - the order for variant sx props is now preserved, and the issue seems fixed

atanasster:
#1326 and #1320 should close this issue

lachlanjc:
Down to 75 issues!! Hurray!

PR #1326, opened 2020-12-02
fix: preserve order of variant expansion props

atanasster:
issue: #1319 

#1273 changed the order of variant props expansion - where local sx props were overwriting any props inherited from variants. This led to incompatibility with older code relying on props order.

Now the variant props are expanded during a loop to place them at the same order where the variant itself was declared.


hasparus (approved):
LGTM

PR #1327, opened 2020-12-02
perf: remove default values for css custom vars

atanasster:
removed the default values for css custom vars, since they are coming from the exact same theme keys. 
As discussed with @hasparus and @fcisio ry to check in while still in the 0.xxx stage.

### DIFF
--- a/packages/color-modes/src/custom-properties.ts
+++ b/packages/color-modes/src/custom-properties.ts
@@ -1,8 +1,8 @@
 import { css, get, Theme } from '@theme-ui/css'
 
 const toVarName = (key: string) => `--theme-ui-${key}`
-const toVarValue = (key: string, value: string | number) =>
-  `var(${toVarName(key)}, ${value})`
+const toVarValue = (key: string) =>
+  `var(${toVarName(key)})`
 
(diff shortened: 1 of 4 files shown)


fcisio (approved):

hasparus:
I suppose we can just revert it if somebody who knows what it's for creates the issue to notify us about a bug :)

hasparus:
Okay, I found the problem.
This PR breaks `@theme-ui/color`.

I'll revert it, and we can rethink the issue considering `@theme-ui/color` and `@theme-ui/custom-properties`.

hasparus:
https://github.com/system-ui/theme-ui/issues/1336

hasparus:
I couldn't revert it from GitHub UI, because the files have changed since, so the status isn't changed.

If we end up fixing the issue on this branch, we need to remember to revert the revert (per the legendary [revert-a-faulty-merge.txt](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/howto/revert-a-faulty-merge.txt)). If we don't want to preserve the history of this branch, we'll rebase or squash its changes and the second revert won't be needed.

PR #1328, opened 2020-12-02
fix: add missing `sx` prop to Box

hasparus:
cc @atanasster @lachlanjc 

Box handles `sx` even without the pragma. Right now it's missing.


lachlanjc (approved):

atanasster (approved):
LGTM

PR #1346, opened 2020-12-08
Media match breakpoint index

### LINKED ISSUE #1191: useBreakpointIndex defaultIndex option doesn't allow you to select the largest breakpoint
**Describe the bug**
You can't specify the largest breakpoint as the `defaultIndex` in `useBreakpointIndex` when using ssr.

**To Reproduce**
1. Create a new app using the base theme
2. Import `@theme-ui/match-media`
3. Create a component that uses `useBreakpointIndex`, setting the `defaultIndex` to 3
4. Render the component using ssr

**Expected behavior**
`useBreakpointIndex` by default returns 0-3, because it gets the length of how many breakpoints it matches (of which there are three by default). I expect to be able to pass the max breakpoint index (3) as the default.

**Additional context**
There's a check for if the `defaultIndex` passed is `< 0 || > breakpoints.length - 1`, but really you're not passing in the actual breakpoint index you want, but the number of breakpoints the screen currently matches:

```
return breakpoints.filter(
  breakpoint =>
  window.matchMedia(`screen and (min-width: ${breakpoint})`).matches
).length
```

Which produces 0-3.

atanasster:
issue: #1191

useBreakpointIndex was returning the number of breakpoints matched, while the default value (defaultIndex) and the name of the hook were implying that the returned values is an index into the array of breakpoints.

The tests were also not correct (adjusted to the wrong implementation) and had to be changed - unfortunately now this can break existing code that would have been relying on the previous code. Usually in only happens if you provided more values than breakpoints are available

For example this used to work
```
const Component = (props) => {
  //4 values provided - previously it would return 'd' - the last value
    value = useResponsiveValue(['a', 'b', 'c', 'd'])
    return null
  }
<ThemeProvider
      theme={{ //3 breakpoints
        breakpoints: ['30em', '45em', '55em'],
      }}>
      <Component />
</ThemeProvider>
```
Now an exception will be generated if more values are used than the number of available breakpoints


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## v0.6.0 UNRELEASED
 
 - Remove recursive default values from CSS custom properties. PR #1327
-
+- Fix match-media useBreakpointIndex defaultIndex. Issue #1191
```
```suggestion
- Fix match-media `defaultIndex` on `useBreakpointIndex`. Issue #1191, PR #1346

```

lachlanjc (approved):
Looks good! Is this breaking? Do we need to update any docs?

CHANGELOG.md
atanasster: thanks !

atanasster:
> Looks good! Is this breaking? Do we need to update any docs?

its is breaking for users who relying on the buggy implementation before - hopefully not that many. IN normal use, it is not breaking, butin any case I marked it now as **BREAKING** in CHANGELOG

hasparus:
Wait, I don't get it.

Given 3 breakpoints, there are 4 possible responsive values, right?

```
value-0 breakpoint-A value-1 breakpoint-B value-2 breakpoint-C value-3  
```

How does this change affect usage of `useBreakpointIndex`?

lachlanjc:
Is this bug still a problem/do we want this PR now?

lachlanjc:
Closing as stale, but we can totally re-open!

korac:
hi there 👋 
I would like to re-open a discussion on this one.

I believe there's a mismatch in terminology here and consequently, in the implementation.

As stated in the [first comment](https://github.com/system-ui/theme-ui/pull/1346#issue-759791417), `useBreakpointIndex`: 
- returns the _number_ of breakpoints matched.
- receives an _index_ of a specific breakpoint (`defaultIndex`).


In the implementation:
- the hook uses the `defaultIndex` (which is an _index_ in breakpoints array) to initialize `value` (local state).
- but `getIndex` function filters the _number_ matched breakpoints (or better to say, matches the index of a responsive value) - which allows the `value` to become larger than the index in breakpoints array.

Consider this example:
```ts
const breakpoints = ['640px', '1240px'];

const bpIndex = useBreakpointIndex(); // Can be one of: `0 | 1 | 2` because the hook returns the *number* of matched breakpoints

const bpIndex2 = useBreakpointIndex({ defaultIndex : 2 }); // throws an error because defaultIndex, in this implementation, tracks the index of the breakpoint (not their number).
```

Basically, what `useBreakpointIndex` receives as an argument and returns are, currently, two distinct notions.

### Quick fix

If believe an easy fix for this is simply allowing `defaultIndex` to be the same length as breakpoints array.
```ts
// /match-media/src/index.ts

if (typeof defaultIndex !== 'number') {
  // ...
} else if (defaultIndex < 0 || defaultIndex > breakpoints.length) { // Not: breakpoints.length - 1
  // ...
}

```

Issue #1345, opened 2020-12-08
theme.styles are not applied in Gatsby site


454de6e:
**Describe the bug**
I have a Gatsby site that does not apply any styles from `theme.styles` during MDX transformations with `theme-ui` and `gatsby-plugin-theme-ui` both at v0.3.4. 

**To Reproduce**

1. Clone https://github.com/maiertech/gatsby-themes/tree/theme-ui-styles-broken
1. Run yarn
1. Run yarn workspace digital-garden-example run dev
1. Launch http://127.0.0.1:8000/posts/ut-officia-officia-ut-lorem-quis-deserunt-irure-nisi-culpa-aliqua-ex-lorem-magna/ and note that the two links are unstyled, even though `theme.styles` styles them.

<img width="230" alt="Screen Shot 2020-12-08 at 1 07 00 AM" src="https://user-images.githubusercontent.com/1482402/101446670-b3380980-38f1-11eb-9d02-50ee2bac1e96.png">

**Expected behavior**
Both links should not be styled with browser defaults but with primary color.

**Additional context**
This looks like https://github.com/system-ui/theme-ui/issues/1148 posted by @cwgw.

Running `yarn list --pattern mdx` yields this:

```
yarn list v1.22.5
├─ @mdx-js/mdx@1.6.22
├─ @mdx-js/react@1.6.22
├─ @mdx-js/util@1.6.22
├─ @theme-ui/mdx@0.3.4
│  └─ @mdx-js/react@1.6.21
├─ babel-plugin-apply-mdx-type-prop@1.6.22
├─ gatsby-plugin-mdx@1.6.0
├─ gatsby-recipes@0.5.0
│  └─ remark-mdx@2.0.0-next.8
├─ remark-mdx@1.6.22
└─ remark-mdxjs@2.0.0-next.8
   └─ @mdx-js/util@2.0.0-next.8
```
Not sure why `@theme-ui/mdx@0.3.4` wants `@mdx-js/react@1.6.21`. I added

```
"resolutions": {
    "@mdx-js/react": "^1.6.22",
    "@mdx-js/util": "^2.0.0-next.8",
    "remark-mdx": "^2.0.0-next.8"
  }
```

and that did not fix the issue (after a Gatsby clean).

I am completely lost here.

hasparus:
Okay, this is certainly something we should work on, but in the meantime could you try the following?

```
"resolutions": {
    "@mdx-js/mdx": "1.6.17",
[...]

PR #1351, opened 2020-12-09
chore: bump mdxjs dependencies

atanasster:
issue: #1345 

Bump versions @mdx-js/mdx and @mdx-js/react to "^1.6.22", gatsby-plugin-mdx to "^1.6.0"


lachlanjc (approved):

454de6e:
See https://github.com/system-ui/theme-ui/issues/1345#issuecomment-742225675.

hasparus:
Pasting @454de6e's comment.

> When using theme-ui in an MDX project (which of course is optional because theme-ui can be used without any MDX), I would expect you would already bring your own @mdx-js/react as peer dependency of whatever framework you use, e.g. gatsby-plugin-mdx. In that case the hard-wired dependency is in the way.
> 
> But I also see your point that you want theme-ui usable without the hassle of having to install peer dependencies. So I wonder if maybe Theme UI should also have an official way of using it modularly, where rather than installing theme-ui you reach for specific packages like @theme-ui/components or @theme-ui/mdx (complemented by @theme-ui/core) depending on your specific styling needs. Then @theme-ui/mdx can have @mdx-js/react as peer dependency and whoever wants the complete experience can install theme-ui which comes with @mdx-js/react as dependency.

We can merge this, but let's make `@mdx-js/react` a peer dep in `theme-ui` and `@theme-ui/mdx`, and align the docs with it.

1. We minimize our surface our surface a bit. Right now Theme UI seems like a huge colossus which does everything. If it's a dependency, people don't have to read its docs, right? And then we have problems.
2. It exports a context provider. Any version mismatch here is breaking the _provider-consumer_ link, so the entire thing stops working.

I have mdx-js versions locked in resolutions in all of my Theme UI + MDX projects. It works but it isn't an experience we'd like the users to have, especially since you can't do it with `npm` AFAIK.

hasparus (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## v0.6.0 UNRELEASED
 
+- Bump versions @mdx-js/mdx and @mdx-js/react to "^1.6.22", gatsby-plugin-mdx to "^1.6.0"
```
```suggestion
- Bump versions `@mdx-js/mdx` and `@mdx-js/react` to `^1.6.22`, gatsby-plugin-mdx to `^1.6.0`. PR #1351
```

lachlanjc (approved):

atanasster:
> Pasting @454de6e's comment.
> 
> > When using theme-ui in an MDX project (which of course is optional because theme-ui can be used without any MDX), I would expect you would already bring your own @mdx-js/react as peer dependency of whatever framework you use, e.g. gatsby-plugin-mdx. In that case the hard-wired dependency is in the way.
> > But I also see your point that you want theme-ui usable without the hassle of having to install peer dependencies. So I wonder if maybe Theme UI should also have an official way of using it modularly, where rather than installing theme-ui you reach for specific packages like @theme-ui/components or @theme-ui/mdx (complemented by @theme-ui/core) depending on your specific styling needs. Then @theme-ui/mdx can have @mdx-js/react as peer dependency and whoever wants the complete experience can install theme-ui which comes with @mdx-js/react as dependency.
> 
> Let's merge this, but let's make `@mdx-js/react` a peer dep in `theme-ui` and `@theme-ui/mdx`, and align the docs with it.
> 
> 1. We minimize our surface our surface a bit. Right now Theme UI seems like a huge colossus which does everything. If it's a dependency, people don't have to read its docs, right? And then we have problems.
> 2. It exports a context provider. Any version mismatch here is breaking the _provider-consumer_ link, so the entire thing stops working.
> 
> I have mdx-js versions locked in resolutions in all of my Theme UI + MDX projects. It works but it isn't an experience we'd like the users to have, especially since you can't do it with `npm` AFAIK.

I believe also just installing a newer version of mdxjs should solve, not just using resolutions. Can you pls confirm?

hasparus:
> I believe also just installing a newer version of mdxjs should solve, not just using resolutions. Can you pls confirm?

Yup, it's still a problem for me. Regardless of the version, we can still get conflicts.

hasparus:
Let's follow up in #1388.

PR #1368, opened 2020-12-14
Remove @emotion/styled from components

### LINKED ISSUE #1183: Typescript forwarding ref issue
Hey, 

If i use Box and then i apply ref forwarding on a new component called button. It would seem that in typescript when using forward ref it seems that box is set to HTMLDivElement and if i supply HTMLButtonElement we get error that it needs to be a div element, i assume this is to do with the 'as' prop?

Just wandered what i'm doing wrong - i even tried the typings for ForwardRef and it still did not work. TS version 4.0.2

for example:
```Javascript
import * as React from 'react';
import { Box, BoxProps } from 'my-lib';

export interface IButtonProps extends BoxProps {
  label: string;
  variant?: 'primary' | 'secondary' | 'tertiary';
  children: React.ReactNode;
}

const Button = React.forwardRef<
  HTMLButtonElement,
  IButtonProps,
>(({ label = 'change me', variant = 'primary', ...rest }, ref) => {
  return (
    <Box
      as="button"
      ref={ref}
      variant={variant}
      aria-label={label}
      {...rest}
    />
  );
});

export { Button };
```

```
Type '((instance: HTMLButtonElement | null) => void) | MutableRefObject<HTMLButtonElement | null> | null' is not assignable to type 'string | ((instance: HTMLDivElement | null) => void) | RefObject<HTMLDivElement> | null | undefined'.
  Type '(instance: HTMLButtonElement | null) => void' is not assignable to type 'string | ((instance: HTMLDivElement | null) => void) | RefObject<HTMLDivElement> | null | undefined'.
    Type '(instance: HTMLButtonElement | null) => void' is not assignable to type '(instance: HTMLDivElement | null) => void'.
      Types of parameters 'instance' and 'instance' are incompatible.
        Type 'HTMLDivElement | null' is not assignable to type 'HTMLButtonElement | null'.
          Type 'HTMLDivElement' is missing the following properties from type 'HTMLButtonElement': disabled, form, formAction, formEnctype, and 13 more.ts(2322)
index.d.ts(143, 9): The expected type comes from property 'ref' which is declared here on type 'IntrinsicAttributes & ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & Pick<...> 
```

Thanks in advance ;-)

atanasster:
- remove @emotion/styled in favor of @emotion/react's css prop - working this PR with @lachlanjc.

a few questions we can discuss
 - the new mergeProps function is merging "arrays" vs "objects" - this can be used for overwriting css props when they are not supported by a browser - ie multiple sx props with the same name (we will need to modify the internal css from '@theme-ui/css' to return an array of props)
- do we need to use the jsx pragma still for the emotion css prop
- any performance/bundle size measurements


packages/components/src/Box.js
```diff
@@ -1,38 +1,64 @@
...
 
-const shouldForwardProp = createShouldForwardProp(boxSystemProps)
-
 const sx = (props) => css(props.sx)(props.theme)
 const base = (props) => css(props.__css)(props.theme)
-const variant = ({ theme, variant, __themeKey = 'variants' }) =>
-  css(get(theme, __themeKey + '.' + variant, get(theme, variant)))
+const variant = ({ theme, variant, __themeKey = 'variants' }) => {
+  return css(get(theme, __themeKey + '.' + variant, get(theme, variant)))(theme)
+}
```
lachlanjc:
```suggestion
const variant = ({ theme, variant, __themeKey = 'variants' }) =>
  css(get(theme, __themeKey + '.' + variant, get(theme, variant)))(theme)
```

packages/components/src/Box.js
```diff
@@ -1,38 +1,64 @@
...
+    sx,
+    (props) => props.css
+  )
+  boxSystemProps.forEach((name) => {
+    delete rest[name]
+  })
+  return <Component ref={ref} css={style} {...rest} />
+})
+
+Box.withComponent = (component) => ({ as, ...props }) => (
```
lachlanjc: Personally I'm super down for deprecating this API—it's really difficult to use at scale in my experience, & I think `as` is highly preferable. If we want to keep this shim around & announce that it's deprecated & axe it in future releases, I'm cool with that too

lachlanjc (commented):
Thanks so much for submitting this :)

packages/components/src/Box.js
atanasster: thanks - fixed. You can guess that I had a few console.logs in there :)

packages/components/src/Box.js
atanasster: Agreed - lets announce it as deprecated (where?) but keep for a while.

lachlanjc: Our Upgrading/Migrating guide I think, & GitHub releases

lachlanjc:
Quick aside—we’re also depending on `@emotion/styled` in the `mdx` package, to power the `Themed` component. So it will still be bundled inside `theme-ui` until we replace that as well, though I think that should be a separate PR.

lachlanjc:
I just tried running [size-limit](http://npm.im/size-limit) on the uncompressed, built versions of the current release vs this branch, & it reduces size by 1KB in this package (17.44KB vs 16.55KB). Less than I thought/hoped, but it’s a solid step nonetheless, & once we remove `@emotion/styled` from `mdx` as well, that’s a much larger reduction (that package alone is 5% of our current bundle size, plus its dependencies).

packages/components/src/Box.js
lachlanjc: We could also do a `console.warn` right here?

packages/components/src/Box.js
```diff
@@ -1,40 +1,63 @@
...
+  )
+  boxSystemProps.forEach((name) => {
+    delete rest[name]
+  })
+  return <Component ref={ref} css={style} {...rest} />
+})
 
+Box.withComponent = (component) => ({ as, ...props }) => (
+  <Box as={component} {...props} />
+)
```
lachlanjc:
```suggestion
Box.withComponent = (component) => ({ as, ...props }) => {
  console.warn('[theme-ui] You’re using the `.withComponent` API on a Theme UI component. This API will be deprecated in the next version; pass the `as` prop instead.')
  return <Box as={component} {...props} />
}
```

packages/components/package.json
```diff
@@ -12,9 +12,7 @@
   },
   "dependencies": {
     "@emotion/react": "^11.1.1",
-    "@emotion/styled": "^11.0.0",
```
hasparus, line 15: _index.d.ts_ imports StyledComponent and we leak it in our public API.

```ts
import { StyledComponent } from '@emotion/styled'
```
```ts
export const Box: StyledComponent<BoxOwnProps, BoxProps>
```

We'll need an update to `index.d.ts` here.

atanasster: I can update it for the sake of it, however it is imported by `@emotion/react` as a devDependency anyway:

https://github.com/emotion-js/emotion/blob/a89d4257b0246a1640a99f77838e5edad4ec4386/packages/react/package.json#L54

hasparus: >  as a devDependency anyway

It's not installed in user's node_modules then.

packages/components/src/Box.js
```diff
@@ -1,40 +1,64 @@
...
+    sx,
+    (props) => props.css
+  )
+  boxSystemProps.forEach((name) => {
+    delete rest[name]
+  })
+  return <Component ref={ref} css={style} {...rest} />
+})
 
+Box.withComponent = (component) => ({ as, ...props }) => {
```
hasparus, line 60: This is actually still breaking.

Previously you could repeat it as much as you want.
```
Box.withComponent('div').withComponent('form')
```

I'm not really opinionated about it, so it's not a suggestion — I just wanted to let you know.

atanasster: yeah, we marked it as deprecated anyway :)

packages/components/package.json
atanasster: Not sure what to do about it - all types are dev dependencies and they are not part of the production bundles.

1. We can add the type and pretty much duplicate `StyledComponent`, with all the risks duplicating code entails. 
2. We can add  `@emotion/styled` as a devDependency (different from using it at run-time as previously). 
3. We can leave as is.

I took #3 but  open to different opinions.

hasparus: > We can add @emotion/styled as a devDependency (different from using it at run-time as previously).
> Not sure what to do about it - all types are dev dependencies and they are not part of the production bundles.

`"dependencies"` and `"devDependencies"` have no inherent meaning in the context of bundling. If I remember correctly, **brunch** bundler used to bundle all `"dependencies"`, but **webpack** doesn't care about them at all (It traverses your module tree.)

They have meaning for package managers however. If a package A has package B in `"dependencies"`, both will appear in user's node_modules when they do `npm install A`. This is not the case for `"devDependencies"`. In case of `"peerDependencies"`, the user will get a warning in the terminal prompting them to install the peerDependency.

If we use types from a devDependency, users won't know to install the package, but the types will collapse to `any` or fail their build when their tsconfig option `skipLibCheck` is set to false.

You can read more about it in npm docs:
- [dependencies](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#dependencies)
- [devDependencies](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#devdependencies)
- [peerDependencies](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#peerdependencies)

There are also `bundledDependencies` and `optionalDependencies`, but I haven't seen anybody use them, so I'm not sure what's the support like.

atanasster: Thanks for the review, I think it should be a devDependency (as it is already through `@emotion/react`, yes its just the typescript type and not sure its worth duplicating the same code. Once we remove the `withComponent` obsolete, we can “rewrite” our own and drop it completely

lachlanjc: What part of the type do we need from Emotion?

packages/components/src/Box.js
lachlanjc: Personally I'm fine with this — this was never a documented feature anyway. Did we want to add a note to the deprecation warning?

packages/components/package.json
atanasster: Its just ‘StyledComponent’ 
https://github.com/system-ui/theme-ui/blob/90b7c10b2c2533c0126c9e12787997c67778eb39/packages/components/index.d.ts#L2

hasparus: > Thanks for the review, I think it should be a devDependency.

So if somebody is depending just on `theme-ui`, and doesn't have `@emotion/styled` in their own dependencies or devDependencies, this change will remove `@emotion/styled` from their node_modules.

If they use TypeScript with `skipLibCheck: false`, we'll crash their build.
If they use JavaScript or skip lib check, we'll de facto disable editor support on Box and Flex, what is still an regression.

If we remove @emotion/styled from runtime, we need to stop depending on their types, especially since the types for Box and Flex is this PR are now incorrect.

lachlanjc: @hasparus Will your PR to move `@emotion/styled` to user/peer deps fix this?

hasparus: @lachlanjc I started working on removing @emotion/styled on `remove-emotion-styled-2` branch — take a look at https://github.com/system-ui/theme-ui/compare/develop...remove-emotion-styled-2.

I think we'll release peer-deps branch and @emotion/styled removal in the same version — both might be breaking changes.

lachlanjc:
Closing in favor of #2043!

PR #1373, opened 2020-12-16
feat(color-modes): Default useColorModeMediaQuery to true

### LINKED ISSUE #624: Default useColorSchemeMediaQuery to true
I seems like the default use case/expected behavior when having a `dark` & a `light` theme would be to match the darkmode media query. Could we default `useColorSchemeMediaQuery` to true to avoid folks having to dig through the docs to figure out whether that's the way it works or not?

lachlanjc:
Closes #624. Breaking change :)


hasparus (approved):
Sweet

lachlanjc:
Ok, so all the checks were passing here so I didn’t investigate further, but it looks like this broke some tests.

PR #1374, opened 2020-12-16
docs: Add section on vendor prefixes to object styles guide

lachlanjc:


hasparus (approved):
LGTM

hasparus:
image

I'm wondering if we should mention vendor prefixed pseudo selectors.

lachlanjc:
Do you mean like input placeholders? You just use those in strings, right?

lachlanjc:
Since they work exactly like non-vendor-prefixed pseudo selectors, I'm cool with omitting for now. I'm going to merge & we can add more docs later, in the spirit of incremental correctness

PR #1390, opened 2020-12-22
fix color modes tests

hasparus:


lachlanjc (approved):
Looks great! I've read through the code & nothing looks out of place. Haven't tested it myself but if tests are passing now, let's do it.

PR #1421, opened 2021-01-07
feat: Move theme configuration options to config object

### LINKED ISSUE #1291: Nest variants and configuration options inside theme

Currently, the Theme object is flat. We're compliant with [System UI Theme](https://system-ui.com/theme) specification, so scales have to be flat, but Variants (and there are quite a lot of them), and options (similarly) don't have to be, and they're not _as important_ as scales.

Right now we kind of enforce on the user to keep their theme object "in order" manually, with whitespaces, regions or manual sorting. Somebody with an ESLint rule sorting object keys will have a really bad time reading interweaved scales, variants and options. 

I think we could keep them nested under `variants` and `options` properties.
Of course, we'd have deprecate the old API and support both scenarios for a while, what's a bit annoying from maintenance perspective.

Leaving it as it is.

**Additional context**

We might have even more options in the near future.
image

---

@lachlanjc @atanasster @dcastil, what do you think?

lachlanjc:
This is a breaking change that affects basically all projects, but improves clarity of themes & gives us a more unified place for future configuration options.

Closes #1291


packages/color-modes/src/custom-properties.ts
```diff
@@ -70,19 +70,28 @@ export const objectToVars = (parent: string, obj: Record<string, any>) => {
...
+  const {
+    colors = {},
+    colors: { modes = {} } = {},
+    config: {
+      useBodyStyles,
+      useCustomProperties,
+      initialColorModeName,
+      printColorModeName,
+    } = {},
+  } = theme
```
lachlanjc: If anyone has suggestions for making this not be so ugly, I’m all-ears. I don’t have a lot of TypeScript experience & this solved all the errors I was running into, but it’s not the best.

hasparus: Why is this ugly? :P

hasparus:
Do you think we should make it non-breaking?
I'm wondering if reading these config options from two places with a deprecation `console.warn` makes sense.
It would ease the migration, but I'm not 100% sure if it's worth the effort.

lachlanjc:
I was thinking we already are changing `useColorSchemeMediaQuery`, & introducing the printing option, so most people are going to be heading to their themes anyway. We could check for `useColorSchemeMediaQuery` directly on the theme & warn then?

hasparus:
I'm a bit afraid of the number of issues we'll get after a stable release. The easier we make the migration the less headache for us in the future.

lachlanjc:
Thoughts @atanasster? Worth making this non-breaking? I'm leaning toward no because it's such a small change for most projects (vs the Styled component might be in dozens of places, this is almost never in several).

atanasster:
> Thoughts @atanasster? Worth making this non-breaking? I'm leaning toward no because it's such a small change for most projects (vs the Styled component might be in dozens of places, this is almost never in several).

Sorry, I am not a fan of breaking changes, actually didnt especially favor `Themed`/ `Styled` change either. Just my 0.02- i realize `config` field is a better design, but not an eye sore to keep current state as well.

lachlanjc:
That's totally valid. Should we read all of these off of `config` first, then fall back to what's defined on the theme?

atanasster:
> That's totally valid. Should we read all of these off of `config` first, then fall back to what's defined on the theme?

That sounds perfect

hasparus:
image

Since we agree on that, I'm marking this PR with "request changes" to avoid merging it accidentally.

hasparus:
Hey @lachlanjc, I started fixing conflicts on this and decided to rebase — here's my WIP branch — https://github.com/system-ui/theme-ui/compare/develop...hasparus:config?expand=1. I'll ask you for code review in a while :)

I need to do a few fixes in places where we had conflicts, but it's looking good. We could release it as 0.7.0.

lachlanjc:
@hasparus Thanks so much for working on this! I'd still love it to happen but haven't put the energy into making it happen.

lachlanjc:
@hasparus Running into an issue with the tests here, where the tests aren't finding the root `html` element—could you take a quick look? Not sure if my code has an issue or the tests need an update

hasparus:
I'll try to find some time this week to take a look at your newest changes, these tests and my previous work at https://github.com/system-ui/theme-ui/compare/develop...hasparus:config?expand=1 (Did I finish all the fixes? I... think so? Why didn't I ask you for review of my fixes?) and consolidate work a bit.

hasparus:
Let's continue on this in https://github.com/system-ui/theme-ui/pull/1688. The PR should be considered as merged after we merge _1688_ to develop.

hasparus:
:rocket: PR was released in [`v0.8.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.8.0) :rocket:

PR #1422, opened 2021-01-09
Update example link to Base Preset

sam-b-rose:
The current Base Preset link points to the old path of the `@theme-ui/preset-base` index file. This file has been renamed from `index.js` to `index.ts`, so the URL pointing to this resource also needs to be updated.


lachlanjc (approved):

hasparus:
Good catch! Thanks for the PR, @samrose3 :pray:

Issue #1434, opened 2021-01-11
Styles root, body vs. html


fcisio:
**Is your feature request related to a problem? Please describe.**
Theme UI root element is `body` which, for most use cases works fine.
There are although some times when you really need to apply styles to to `html` tag.

One of these uses cases is when dealing with `rem` units (which are relative the root `html`). So `rem` units currently do not work with Theme UI.

It would be nice to have a way to target the `html` tag somehow.

**Describe the solution you'd like**
Creating a version of `<Global />` with Theme UI capabilities could be useful.

I'll be working on something like that on one of my projects. We could potentially package it for the library.

I was thinking of a component inside the provider with an instance of Emotion `<Global />` onto which we could inject a part of the Theme Context.

**Additional context**
I'd love to know what you guys think about something like this and if you've experienced similar issues.

fcisio:
Here's my working "hack":

```jsx
import React from 'react'
import { css, useThemeUI } from 'theme-ui'
import { pickBy, omitBy, isObject } from 'lodash'
import { Global } from '@emotion/core'

const ThemeGlobalRoot = ({ key = '_root' }) => {
  const theme = useThemeUI(null).theme.styles[key]
  const _ = (func) => func(theme, isObject)

  return <Global styles={css({ html: _(omitBy), ..._(pickBy) })} />
}

export default ThemeGlobalRoot
```

Theme:

```js
styles: {
  _root: {
    // HTML scope
  },
  root: {
    // default scope
  },
}
```

```jsx
<ThemeProvider theme={theme}>
  <ThemeGlobalRoot />
  {children}
</ThemeProvider>
```

atanasster:
@fcisio i havent encountered such issue, but it seems legit.

for the implementation, have you thought about making it a theme prop - similar to useCustomProps, this way it will all work automatically without the need for another ThemeGlobal?

fcisio:
I think it would be the best solution yes!
I'm looking into a PR, but I'm not familiar with the codebase yet. Could take a while.

I did find a comment from @hasparus dating from a year ago:
`packages/theme-ui/src/index.ts`:

```ts
// TODO: work this into root provider/global styles?
export const BaseStyles = (props: Record<string, unknown>) =>
  jsx('div', {
    ...props,
    sx: {
      fontFamily: 'body',
      lineHeight: 'body',
      fontWeight: 'body',
      variant: 'styles',
    },
  })
```

atanasster:
[...]

PR #1441, opened 2021-01-17
New useRootStyles flag

fcisio:
In reference to #1434 
Here's my implementation draft.

We discussed either creating a new flag or exporting a new Global component. I feel like creating a new flag is best, since having the styles on the `<html>` element should be the default in the future. If it becomes the default, then `useBodyStyles` would need to be renamed anyway.

So with this, we have the `useRootStyles`, which is optional and overrides `useBodyStyles`, and down the line, we can deprecate `useBodyStyles`.

---
This is my first PR on the project, it still needs some love 😅
@atanasster could you give me a hand with:
- Adding a test for `BodyStyles` now named `RootStyles`
- Making sure that my `useRootProp()` function respects the right structure and is in the right place


fcisio:
I found a bug with the `noflash` script on the `develop` branch.
Seems to come from @hasparus commit `b16a44` : "fix(color-modes): fix tests, read preferred scheme synchronously"

Not sure if I should create an issue, let me know what's best and I'll share what I found.

packages/css/src/types.ts
```diff
@@ -602,6 +602,16 @@ export interface ThemeStyles {
   [key: string]: ThemeUIStyleObject | undefined
 }
 
+export function useRootProp(theme: Theme = {}): any {
```
atanasster: I am used to have react hooks as `useXXX` function names. Maybe we can change this to a simple function name like `getRootStyles`or `getUseRootStyles` ?

packages/css/src/types.ts
```diff
@@ -642,10 +652,15 @@ export interface Theme {
   printColorModeName?: string
 
   /**
-   * Adds styles defined in theme.styles.root to the <body> element along with color and background-color
+   * Adds styles defined in theme.styles.root to the <html> element along with color and background-color
    */
-  useBodyStyles?: boolean
+  useRootStyles?: boolean
```
atanasster: I think this might be a bit confusing - `useRootStyles` is declared as optional boolean, however it is tested as potentially `null` and we also allow for `useBodyStyles`. 

What do you think about having a theme prop 'rootElement?: string` - by default this is 'body' but user can change it to any other tag.
And we keep just `useBodyStyles` (even though confusing since it can be not just body anymore) for compatibility sake?

atanasster (commented):
Great job!!! 
I added a comment - its mostly about the structure of the theme option
`useRootStyles` vs `rootElement`. For me it is a bit confusing, but I can also understand your initial decision and its perfectly valid as well.

packages/css/src/types.ts
fcisio: My idea of this is basically that the HTML should ideally always be the root.
Having a prop like `rootElement` would potentially create a window for breaking themes (not compatible with presets, and most importantly the `noflash` script).

The only reason why I kept `useBodyStyles` is for backward compatibility reasons. The flag might be used in theme presets and people styles might be written to work with the body as root (ex: `'&.random-body-class'`).

We could also simply rename `useBodyStyles`, change the docs, and introduce it as a breaking change.

I'm allowing `useRootStyles` to be null, for uses cases like the flag being present in a preset, but the user want's to revert to the body.

---

There are no downsides about using HTML over BODY as root, apart from backward compatibility, and it would allow using HTML CSS selectors. So I think that should be the only root option in the future.

What do you guys think?

atanasster: Thanks for the explanation, makes sense, I was just expressing my pov :)

However, we definitely need to keep backward compatibility (as you have it in your implementation) imho.

atanasster (approved):
lgtm !

packages/css/src/types.ts
```diff
@@ -602,6 +602,16 @@ export interface ThemeStyles {
   [key: string]: ThemeUIStyleObject | undefined
 }
 
+export function getUseRootStyles(theme: Theme = {}): any {
```
hasparus: image
Pourquoi, Francis? Pourquoi?

hasparus:
```suggestion
export function getUseRootStyles(theme: Theme = {}) {
```

hasparus (changes requested):
The PR is so good I have to assume that `: any` is just there just to mess with me.

hasparus:
Vercel deploy fails because there's a bash script missing here and present on develop, but it's okay.

packages/css/src/types.ts
```diff
@@ -602,7 +602,7 @@ export interface ThemeStyles {
   [key: string]: ThemeUIStyleObject | undefined
 }
 
-export function getUseRootStyles(theme: Theme = {}): any {
+export function getUseRootStyles(theme: Theme = {}) {
```
hasparus: Actually... Why is this in types.ts file?

packages/css/src/types.ts
fcisio: Haha, look at you leaving comments in my native language 😂

I think I mimicked from another function, I'm not super familiar with typescript.

packages/color-modes/src/index.tsx
```diff
@@ -155,6 +156,7 @@ export const ColorModeProvider: React.FC = ({ children }) => {
 const noflash = `(function() { try {
   var mode = localStorage.getItem('theme-ui-color-mode');
   if (!mode) return
+  document.documentElement.classList.add('theme-ui-' + mode);
```
hasparus: Why do we add the class both to the `html` and to the `body`? Shouldn't it be _one or the other_?

packages/css/src/types.ts
fcisio: I was expecting this one. I didn't find a better place where to put this, also not being familiar with the project.
Where would you put it?

packages/color-modes/src/index.tsx
fcisio: Yeah it should match what is set with the flag.
I wasn't sure if it would be a good idea to add a prop in the `noflash` function.

hasparus: We have the theme in scope, so we could possibly read that option and add only the one line that's needed to the string.

packages/css/src/types.ts
hasparus: This is the only piece of executable runtime code in `types.ts`. I'd probably stick it in `index.ts`. It's growing, but I don't think we have to split it into multiple files yet.

packages/css/src/types.ts
```diff
@@ -642,7 +652,12 @@ export interface Theme {
   printColorModeName?: string
 
   /**
-   * Adds styles defined in theme.styles.root to the <body> element along with color and background-color
+   * Adds styles defined in theme.styles.root to the <html> element along with color and background-color
+   */
+  useRootStyles?: boolean
+
+  /**
+   * Legacy : Adds styles defined in theme.styles.root to the <body> element along with color and background-color
```
hasparus:
```suggestion
   * @deprecated
   * Adds styles defined in theme.styles.root to the <body> element along with color and background-color
```

This should make `useBodyStyles` display with strikethrough in popular IDEs AFAIK (VSCode, and possible JetBrains).
There are also lint rules checking for usage of deprecated APIs which respect this JSDoc comment.

hasparus:
**To-Do for myself**

- [x] Make changes we discussed with fcis on Discord.

Issue #1439, opened 2021-01-16
[0.6.0-alpha.4]: Scales with key named "default" are broken


fcisio:
**Describe the bug**
Installed `0.6.0-alpha.4` on Gatsby Starter, and it seems that the `boxShadows`, `borders` & `zIndices` keys are broken.

**To Reproduce**
[See the repro](https://github.com/fcisio/repro-theme-ui-broken-keys)

```json
"react": "^17.0.0",
"react-dom": "^17.0.0",
"theme-ui": "^0.6.0-alpha.4",
```

1. `yarn && yarn develop`
2. Inspect the `<h1 />` on the homepage

> You should see that the theme properties are working, with the exception of `box-shadow`, `border` & `z-index`

**Expected behavior**
All references should output the value set on the theme

**Screenshots**
![Shot_0 26](https://user-images.githubusercontent.com/46095852/104806927-03d03b80-57a9-11eb-9443-a3c971ea14f9.jpg)

**Additional context**
I extended the tests on a fork, but they seemed to all pass.

lachlanjc:
Having this issue too. `borderRadius` is broken for me too

hasparus:
I cherry-picked @fcisio's test on top of last 0.6.0-alpha.4 commit, and tests passed :/

So I made a tiny repro to see how it works on CodeSandbox, and it seems it's more complex than broken scales
https://codesandbox.io/s/lucid-cohen-sw28l?file=/src/App.js

hasparus:
I just released 0.6.0-alpha.5 with a fix for this.

Proof by CodeSandbox:

Before: https://codesandbox.io/s/lucid-cohen-sw28l?file=/src/App.js
After: https://codesandbox.io/s/elastic-keller-uumrz?file=/src/App.js:230-276

PR #1442, opened 2021-01-17
Update system prop test

fcisio:
There is an issue currently on the latest Alpha #1439
It seems to be fixed on the `develop` now; not sure what fixed it. Should be good on the next alpha release I presume?

`borders`, `shadows` & `zIndices` keys were missing from the system props tests. I added them so we make sure it's fixed on the next release 👍


hasparus:
Thanks for the PR, fcisio!

Issue #1406, opened 2020-12-30
Color modes not using CSS custom properties


okonet:
Looking at the code I see that to set the color mode, it relies on the emotion theme provider:

https://github.com/system-ui/theme-ui/blob/6d9e29369a76c2802a1bb8253efaef072a087fbc/packages/color-modes/src/index.tsx#L111-L117

This generates unnecessary class names in order to change the color even though all the custom props are in DOM:

<img width="854" alt="CleanShot 2020-12-30 at 19 29 59@2x" src="https://user-images.githubusercontent.com/11071/103373509-736dd780-4ad5-11eb-9a7e-d6d9548937e0.png">

I've build a quick experiment where I'm relying on changing only the className of the `body` element and it works as expected without any additional re-rendering caused by emotion.

I'm wondering if the current behavior is intentional and what's the motivation to rely on have JS runtime for changing color mode?

atanasster:
Hi @okonet - not sure i understand 100% the issue you are reporting, however if you mean that new emotion classes are generated because all the color references use a default value for css props - then this is changed in the latest beta versions.

what version are you using? Also if this is still an issue, can you please post an example to reproduce and help us fix it.

lachlanjc:
I'm also not really understanding the issue here. So we do generate all the custom properties in advance, explicitly so less JavaScript needs to run when the theme changes. But there's no way to build a reliable theme switcher without JavaScript: we're a React library, so we need JS to offer a hook for executing the switch from any UI or background component in your app, & there's not any decent way to switch body classes or switch all the custom properties used without JS. & since there's a hook for reading the current color mode, if we didn't run through JS, React wouldn't re-render with new data in that hook, so various parts of the app would be out of date.

Not requiring JS is a valid requirement for some apps/components, but Theme UI isn't the right library if you're looking for approaches close to the DOM.

okonet:
@atanasster hmm, let me try the latest beta. I'm on the latest stable version: 3.5.0 and this is still an issue. 

I do understand why you'd like to rely on the React context here, but I also am confused why is there the need to pre-render all CSS custom properties then. That's why I raised this question. To me having them in DOM should eliminate the need to rely on the context but it looks like I'm mistaken?

lachlanjc:
[...]

PR #1453, opened 2021-01-18
fix: add theme.original space to save colors

atanasster:
issue #1406
added tests
added field theme.original.colors to save colors in emotion theme object


packages/css/src/types.ts
```diff
@@ -808,4 +808,14 @@ export interface Theme {
    * @see https://theme-ui.com/components/message#variants
    */
   messages?: Record<string, ThemeUIStyleObject>
+
+  /**
+   * When the theme values are replaced with css custom vars, save original values
+   */
+  original?: {
```
hasparus: I'm skeptical towards including it in the types for our public API.
1. Users won't ever add it manually. If they have a key called original, we'd overwrite it.
2. We don't have that value on the theme in **our context**. We add it to theme's shallow copy passed to **Emotion's context**.

packages/color-modes/src/index.tsx
```diff
@@ -131,8 +131,10 @@ export const ColorModeProvider: React.FC = ({ children }) => {
 
   if (theme.useCustomProperties !== false) {
     emotionTheme.colors = toCustomProperties(emotionTheme.colors, 'colors')
+    emotionTheme.original = {
```
hasparus: There's a tiny risk that users will have a key named "original" in their themes, what makes it a breaking change.
A use-case for a Symbol perhaps? We could keep this more "private".

packages/css/src/types.ts
atanasster: removed

hasparus:
Awesome. Looks good to merge.

lachlanjc (approved):

hasparus:
Okay, after fighting with color mode flash for _ages_ and reading https://github.com/system-ui/theme-ui/pull/259 for a tenth time I'm positive that we need to bring back `rawColors`
-  either by saving them to `__original.colors` like this PR 
- **or** saving transformed colors to `theme.__colorsVars`, and changing scales to use `__colorsVars` instead of `colors`.
  - This way we'd leave `colors` intact and avoid crashing @LekoArts's Cypress tests again :P 
  - Leaving `colors` intact is cool in case users want to transform them with functions from polished / pass them to threejs canvas or anywhere else.

atanasster:
I dont see any reference to ‘rawColors’ in this PR?

hasparus:
> I dont see any reference to ‘rawColors’ in this PR?

Sorry for the confusion.
`__original.colors` already existed in Theme UI in the past and was called `rawColors`. Debugging color mode flash led me to the CSS properties and the observation that sometimes we read colors from Emotion context, and sometimes (e.g. code snippets in Theme UI docs) we read them unprocessed from our context.

I'm in the process of fixing merge conflicts and merging this PR right now — The alternative to `theme.original.colors` - saving processed css properties to `theme.colorVars` is infeasible due to how our packages are separated.

PR #1454, opened 2021-01-19
(docs): Move CSS Logical Properties to separate table

lachlanjc:
As discussed when the list was added, the logical properties makes the table super long, so I’ve summarized them & moved them into a details tag.

Preview live: https://theme-ui-git-docs-separate-logical.systemui.vercel.app/sx-prop

### DIFF
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -357,16 +357,18 @@ Because MDX uses its own custom pragma and `createElement` function, the Theme U
 You can use any of the [Theme UI components](/components),
 which support the `sx` prop, in an MDX file as a workaround.
 
-```js
+```mdx
 import { Box } from 'theme-ui'
 
-;<Box
+<Box
   sx={{
     padding: 3,
     bg: 'highlight',
   }}
 >
-  # Hello
+
+# Hello
+
 </Box>
 ```
 
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -66,12 +66,6 @@ The following CSS properties will use values defined in the theme, when availabl
 | `ml`, `marginLeft`        | `space`          |
 | `mx`, `marginX`           | `space`          |
 | `my`, `marginY`           | `space`          |
-| `marginBlock`             | `space`          |
-| `marginBlockEnd`          | `space`          |
-| `marginBlockStart`        | `space`          |
-| `marginInline`            | `space`          |
-| `marginInlineEnd`         | `space`          |
-| `marginInlineStart`       | `space`          |
 | `p`, `padding`            | `space`          |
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/sx-prop.mdx
```diff
@@ -344,15 +360,13 @@ which support the `sx` prop, in an MDX file as a workaround.
 ```js
 import { Box } from 'theme-ui'
 
-<Box
+;<Box
```
atanasster: `;<Box` - any idea why this semi-colon made it in there? I tried and my prettier also messes up the source code

atanasster (approved):
Great job!!!

packages/docs/src/pages/sx-prop.mdx
lachlanjc: Huh. Yeah, I ran this through Prettier to fix the tables & I didn't notice this. Maybe adding `return` or something would help? It's obviously a typo on our page but I don't want to be manually changing things from Prettier as they might slip by reverted in the future.

atanasster: Agreed, no idea howto fix it in prettier

hasparus: We could try to turn on semicolons for the docs. (I think Prettier doesn't like lines starting from JSX when the previous line has no semi.)

hasparus (approved):
LGTM

packages/docs/src/pages/sx-prop.mdx
hasparus:
```suggestion
<Box
```

packages/docs/src/pages/sx-prop.mdx
```diff
@@ -344,15 +360,13 @@ which support the `sx` prop, in an MDX file as a workaround.
...
   sx={{
     padding: 3,
     bg: 'highlight',
   }}
 >
-
-# Hello
-
+  # Hello
```
hasparus: Wait, this doesn't work, does it?

If you don't wrap markdown in empty lines you're still in JSX, so this won't render a heading, and that # is just printed. I mean, I know that's just a code snippet in the docs, but still.

packages/docs/src/pages/sx-prop.mdx
lachlanjc: I'd like to avoid that since they're unnecessary clutter & not how we write the library itself. I'll add export default I guess?

packages/docs/src/pages/sx-prop.mdx
lachlanjc: You're correct, I believe. I guess we'll just counter Prettier here & figure out how to avoid it in the future.

packages/docs/src/pages/sx-prop.mdx
lachlanjc: Ah. The problem is the language on the syntax highlighting is wrong.

PR #1455, opened 2021-01-20
fix: custom vars to be created even if no color modes

atanasster:
@fcisio reported on discord:

`Just realized, that the custom properties are only printed on the body if modes is in the colors declaration.`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 ## v0.6.0 UNRELEASED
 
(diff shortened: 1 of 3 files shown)


hasparus (approved):
Do you think we should update CHANGELOG with this?

lachlanjc:
May as well—let's be on the side of completeness in the CHANGELOG, then only mention things broadly relevant in our Upgrading guide we point people toward for stable releases

PR #1460, opened 2021-01-22
feat(css): skip false before passing style object to Emotion

### LINKED ISSUE #1297: [typescript] Sx properties should accept booleans

hasparus:
Closes https://github.com/system-ui/theme-ui/issues/1297

Allows TypeScript users to write
```
sx={{ color: isActive && 'blue' }}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.6.0 UNRELEASED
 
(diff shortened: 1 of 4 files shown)


atanasster (approved):
Looking great !

PR #1461, opened 2021-01-23
chore: deprecate mdx/Styled

atanasster:
Bring back Styled but with a console warning. The code is a bit hacky, but couldn't come up with anything better.


lachlanjc (approved):

lachlanjc:
Let's add a line to CHANGELOG as well 👍

atanasster:
> Let's add a line to CHANGELOG as well 👍

True - will you add it or me?

packages/mdx/src/index.ts
```diff
@@ -156,12 +158,29 @@ export const Themed: ThemedDiv & ThemedComponentsDict = styled('div')(
   themed('div')
 ) as ThemedDiv & ThemedComponentsDict
 
+// to be deprecated
```
hasparus: Okay, I think this is **already deprecated**, and deprecated means "to be removed in future releases".

We can use a JSDoc comment to mark that and point to the new name.
https://jsdoc.app/tags-deprecated.html

```suggestion
/**
 * @deprecated since 0.6.0.
 * 
 * `Styled` was renamed to `Themed` to avoid confusion with styled components.
 */
```

This is how it shows up in user's editors.

image

CHANGELOG.md
```diff
@@ -2,10 +2,11 @@
 
 ## v0.6.0 UNRELEASED
 
-- Skip `false` values before passing style objects to Emotion. Allow `false` as style property value in TS types. Issue #1297, PR #1460.
+- Make the deprecation of `Styled` → `Themed` non-breaking, with backwards compatability & a deprecation warning on `Styled` until a future release. PR #1461
```
hasparus:
```suggestion
- Make the rename of `Styled` to `Themed` non-breaking. Add a deprecation warning on `Styled` until a future release. PR #1461
```

packages/mdx/src/index.ts
```diff
@@ -156,12 +158,29 @@ export const Themed: ThemedDiv & ThemedComponentsDict = styled('div')(
...
+  return createElement(tag, props)
+}
+
 export const components = {} as ThemedComponentsDict
 
 tags.forEach((tag) => {
   // fixme?
   components[tag] = styled(alias(tag))(themed(tag)) as any
   Themed[tag] = components[tag] as any
+  Styled[tag] = styled(warnStyled(tag))(themed(tag)) as any
```
hasparus, line 188: What do you think about using a proxy instead of copying the code for Themed?

I've been thinking about something like this: [**TS Playground**](https://www.typescriptlang.org/play?#code/MYewdgzgLgBADgJxMAphCMC8MDeMVgBuAXLgL4wCGGASiqAgCYA80CAlmAOYA0MArmADWYEAHcwAPhhkA3AFgAUEtCRYXBChRQUCLLhgRKATwASKADYWQACgCUB1RBAWUAOmtcbAcnNWQMADqIAgWjACE3g4UZEoq4NAwjCiI9JQ6jADimtq6+mAoYjAACkgAHsY2Glo6CHw4SjBNMFzaNlCUCK1QfEIoxg4Nis0jMOwAZjA2iMhoEG4EhG4AcgDyACIAogD6m8sAajDhmNjeM4z8wFDs4FG4jaOPE1N9xmNgMB1d2oMPj-8wJwudxiTpgGx-AFQgAGAGUoMZXIw3AASHDwjjcGyvOwUcZIAC2MHxICJACIAAJQAAWKAJKAAtPx2AB6AmMMpksYYZKpYDpFCMKhgIVidhWGAAIxQME0BJAhEF72hkKh-wA1DBoVRifwoPxNDBFQgIDcwHw4K5qDL+BAZQAVWn05FojGcLw4iicaAoSjI6F2VX-WLDYNxUOPTT6hAfL7dADaryoGFeIEmCLgKDTn063QAugoIzIlGRA8pFNVcgg3EYzJZrPYlLzNPyMtkaroayY-A27LIgA)

```ts
// after rename
const greeter = { sayHello() { console.log('Hello World!') } }

// before rename (like our Styled)
const deprecatedGreeter = new Proxy(greeter, {
    get(target, key) {
        if (process.env.NODE_ENV !== 'production') {
            if (key in target) {
                console.warn(
                    `Styled.${String(key)} from from "@theme-ui/mdx" is deprecated and will be removed in`
                    + ` a future version, please use Themed.${String(key)} instead.`)
            }
        }

        return target[key as keyof typeof target];
    }
})
```

hasparus:
Do you guys think it would be an overkill to test it?

image

We already have a bunch of tests testing warning messages. We could keep up this tradition.

hasparus (changes requested):
Made some comments. Just marking this here so I don't merge it accidentally.

lachlanjc:
> Do you guys think it would be an overkill to test it?

The only reason not to test things is so we can build more things faster—let’s absolutely add tests for that kind of thing too. We can always remove them later very easily

packages/mdx/src/index.ts
atanasster: I would prefer to keep the old implementation - and avoid any new bugs.

atanasster:
> > Do you guys think it would be an overkill to test it?
> 
> The only reason not to test things is so we can build more things faster—let’s absolutely add tests for that kind of thing too. We can always remove them later very easily

In my book, it's an overkill to spend any more time on a deprecated component that will be pulled out soon, while it holds back a stable release.

Btw there is a small test `test('Deprecated Styled test'`

packages/mdx/src/index.ts
hasparus:
```
styled(warnStyled(tag))(themed(tag))
```

is a _new_ implementation. And it's bug-prone — right now `alias(tag)` is missing. Any accidental change can break it which won't get caught by Themed tests. 

image

This is a new bug. 

The approach I'm advising is 
1. Themed is the new API. We should thoroughly test that it works.
2. Styled is just an alias/proxy. We should briefly test that it's an alias/proxy. We should also test that it prints the warning.

hasparus:
> a deprecated component that will be pulled out soon

I assume that if we're deprecating something, we'll keep it to next stable "major" release.

packages/mdx/src/index.ts
atanasster: 1. Regarding using a Proxy, I think we can use it for the new implementation of `Themed` - but keep `Styled` as close as possible to the old `Styled`. 

2. Great catch on alias - I am adding the aliased tags as well as a test for inlineCode.

3. There was already a test for `Styled`, just to make sure it functions. I am now adding tests for alias and console. warning

atanasster: > I assume that if we're deprecating something, we'll keep it to next stable "major" release.

agreed, I assume we are heading for the major release soon :)

lachlanjc (approved):

hasparus (approved):

CHANGELOG.md
```diff
@@ -36,6 +37,7 @@
 
 ## v0.6.0-alpha.6 2021-01-22
 
+- Skip `false` values before passing style objects to Emotion. Allow `false` as style property value in TS types. Issue #1297, PR #1460.
```
```suggestion
```
Removing this since I’m not sure why it’s in this branch

packages/mdx/src/index.ts
```diff
@@ -156,12 +158,34 @@ export const Themed: ThemedDiv & ThemedComponentsDict = styled('div')(
...
+ */
+export const Styled: ThemedDiv & ThemedComponentsDict = styled('div')(
+  themed('div')
+) as ThemedDiv & ThemedComponentsDict
+
+const warnStyled = (tag: keyof IntrinsicSxElements): FC => (props) => {
+  useEffect(() => {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(
+        'The Styled component from "@theme-ui/mdx" is deprecated and will be removed in a future version, please use Themed instead'
```
lachlanjc:
```suggestion
        '[theme-ui] The Styled component from "@theme-ui/mdx" is deprecated and will be removed in a future version, please use Themed instead'
```

packages/mdx/src/index.ts
```diff
@@ -156,12 +158,34 @@ export const Themed: ThemedDiv & ThemedComponentsDict = styled('div')(
...
+ */
+export const Styled: ThemedDiv & ThemedComponentsDict = styled('div')(
+  themed('div')
+) as ThemedDiv & ThemedComponentsDict
+
+const warnStyled = (tag: keyof IntrinsicSxElements): FC => (props) => {
+  useEffect(() => {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(
+        '[theme-ui] The Styled component from "@theme-ui/mdx" is deprecated and will be removed in a future version, please use Themed instead'
```
lachlanjc:
```suggestion
        '[theme-ui] The Styled component from "@theme-ui/mdx" is deprecated and will be removed in a future version. It has been renamed to Themed with the same API.'
```

PR #1475, opened 2021-02-03
Remove JSX pragma from gatsby-theme-style-guide

aaronadamsCA:
Should fix `pragma and pragmaFrag cannot be set when runtime is automatic` when installing version `next` of Theme UI and this Gatsby plugin.


lachlanjc:
Good call! I think this will break `Row` though — want to make that use `Flex` instead of `div`?

aaronadamsCA:
Give me a little time to look further into it, just a bit swamped right now.

(Did not know this could have that effect, hmm!)

lachlanjc:
No worries! I just invaded the branch with the fix :)

lachlanjc (approved):

hasparus (approved):

Issue #1477, opened 2021-02-06
Change "__default" key CSS Custom Properties output


fcisio:
**Is your feature request related to a problem? Please describe.**

I love the new `__default` key behavior, but I did run into a use case issue.

I have a function that transforms color strings into custom properties. I use it to reference colors on other custom properties.

```js
--color: toCSSVar('primary'),
--color-hover: toCSSVar('primary.light'),
```

```js
--color: 'var(--theme-ui-colors-primary)',
--color-hover: 'var(--theme-ui-colors-primary-light)',
```

Currently a color with a `__default` key prints `--theme-ui-colors-primary-__default`.

For me, it means that my function doesn't work 100% of the time. I can easily adapt my function, but it makes me wonder...

**Describe the solution you'd like**

Should this be transformed internally so that it stays consistent with past behavior?

### From
```css
--theme-ui-colors-primary-__default: red;
--theme-ui-colors-primary-light: pink;`
```

### To
```css
--theme-ui-colors-primary: red;
--theme-ui-colors-primary-light: pink;
```

> It also looks a bit cleaner.

PR #1481, opened 2021-02-06
Remove "__default" key from CSS custom properties

fcisio:
This change follows my issue #1477 
Seems simple enough, let me know if you guys are down with this.

I had some type errors on my test, some color mode keys don't seem to allow for objects, only for color strings.

`primary?: CSS.Property.Color;`


packages/color-modes/src/custom-properties.ts
```diff
@@ -1,8 +1,9 @@
 import { css, get, Theme, __internalGetUseRootStyles } from '@theme-ui/css'
 
-const toVarName = (key: string) => `--theme-ui-${key}`
+const stripKeyDefault = (key: string) => key.replace('-__default', '')
+const toVarName = (key: string) => `--theme-ui-${stripKeyDefault(key)}`
 const toVarValue = (key: string, value: string | number) =>
-  `var(${toVarName(key)}, ${value})`
+  `var(${toVarName(stripKeyDefault(key))}, ${value})`
```
hasparus: I don't get this part.

Isn't `toVarValue` now

```
const toVarValue = (key: string, value: string | number) =>
  `var(`--theme-ui-${stripKeyDefault((stripKeyDefault(key)))`}, ${value})`
```

stripKeyDefault is called twice.

fcisio: You're right, I did this pretty quickly I didn't realize they were fed into each other.
Just fixed it in a new commit.

hasparus:
Okay it looks good, but I have a slight worry we have a few implicit dependencies between modules which are not covered with tests.

- [ ] Should `@theme-ui/custom-properties` also be updated?
- [x] Does this work with `getColor` from `@theme-ui/color`? (It should be fine, but I'll add tests for it.) 

---

> I had some type errors on my test, some color mode keys don't seem to allow for objects, only for color strings.
>
> primary?: CSS.Property.Color;

This is because many components from Theme UI components use `color: 'primary'`, so before `__default` was supported, if you had primary as object, you had to change their styles. Components predate both types and `__default`, and I wanted to suggest with TypeScript that "primary" as simple color will cause less problems down the road.

I'm gonna update types. I'll send another PR and ask for your reiew, okay?

hasparus (approved):

hasparus:
As a side note for myself, [CI](
https://github.com/system-ui/theme-ui/blob/develop/.github/workflows/ci.yml) didn't run on this PR.

PR #1482, opened 2021-02-07
Update the License Year

sampoder:
It's 2021, yay(?)!


lachlanjc:
Appreciate it Sam, but we're not planning to keep updating this since Brent authored it in 2019.

PR #1492, opened 2021-02-08
Support nested scales with __default key as style values

hasparus:
I updated types to allow nested color scales in known color names to fit @fcisio's test from https://github.com/system-ui/theme-ui/pull/1481 (and the changelog), and I noticed this is a slight breaking change on type-level.
 
Our existing tests didn't typecheck, and what's more, a lot of them didn't **work** in cases where `primary` and other colors were objects with `__default` key.
  
I was wondering how can I make it "less breaking" and smooth out the experience around `__default` so you don't need to change much code when you go from `primary: 'blue'` to `primary: { __default: 'blue' }`. Well, this is it. 

I didn't spend too much time on it, and I'm considering taking a step back, because I'm not sure if the change to css/src/index.ts isn't superflous (it overlaps with handling `__default` inside `get`, though I subjectively like it a bit more as it doesn't add so many undocumented features)

cc @fcisio @lachlanjc, would you like to take a look here and tell me what you think?


fcisio:
@hasparus the current looks good to me 👍

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## v0.6.0 UNRELEASED
 
+- **Breaking TypeScript**: Known colors (_primary_, _text_, _background_, _accent_, _secondary_) in `ColorMode` can now be nested scales.
+
+  Following no longer typechecks, as `colors.primary` can be an object.
```
hasparus:
```suggestion
  The following no longer typechecks, as `colors.primary` can be an object.
```

Issue #1130, opened 2020-08-14
Next.js not applying styles to MDX components


jonashaefele:
**Describe the bug**
I ran into the same issue described [here](https://github.com/vercel/next.js/issues/13047) and [here](https://github.com/hashicorp/next-mdx-enhanced/issues/53)

When rendering MDX pages in Next.js, the styles defined in theme.js don't get applied.

The styles get applied perfectly well when using `Styled.h1` and its sibling components in a JSX file, but the MDX parser seems to render plain HTML tags without the classes needed to apply the styles defined the `theme.js`

**To Reproduce**

The issue is present in the [next example](https://github.com/system-ui/theme-ui/tree/master/examples/next) in this repo. Run the example and you'll see that only the root styles are applied, but the MDX components are rendered plain, without styles.

The H1 should be `color: 'primary'` (#07c) but gets rendered with the black (#000) inherited from the root styles applied to the body.

Change the first few lines in `index.mdx` to make it easier to see:

```
import { Styled } from 'theme-ui'

# MD H1

<Styled.h1>Styled H1</Styled.h1>
```

**Expected behavior**

I'd expect MDX files to render the MD components with all the ThemeUI styles defined in the `styles` object of `theme.js` or in other words to `# Title` should render identically to `<Styled.h1>Title</Styled.h1>`, etc.

**Screenshots**

<img width="1180" alt="Screen Shot 2020-08-14 at 18 45 14" src="https://user-images.githubusercontent.com/1152888/90277916-64f08d80-de5e-11ea-94ab-ab0e8ab378a5.png">
<img width="1182" alt="Screen Shot 2020-08-14 at 18 45 23" src="https://user-images.githubusercontent.com/1152888/90277923-6752e780-de5e-11ea-919f-c6b7e74492f2.png">

**Additional context**
It looks like it might be a [next.js bug](https://github.com/hashicorp/next-mdx-enhanced/issues/53), but all the issues related to it have been closed.

I tried to debug it for two days now. I changed the versions for `next` and `@next/mdx` to
- 9.3.3
- 9.3.6
- 9.5.1
- 9.5.2

All of which were suggested should work at some point in issue comments, but none of them seem to work in the next example or my project (which is more complex, so I tried to  isolate the issue).

I appreciate this might not be a theme-ui issue, but thought maybe someone has already solved it.:

Edit: I also tried to swap `@next/mdx` for https://github.com/hashicorp/next-mdx-enhanced (latest) but that didn't help either.

[...]

Issue #839, opened 2020-04-10
Understanding how the sx prop works, sx prop object composition, etc


erikdstock:
I think understand the jsx pragma but have a couple questions about `sx`.
1: The pragma does not seem to be necessary for theme-ui components such as `<Box />`, but is is for the native/intrinsic html elements such as `<div />`s.  Is this correct?

2: Perhaps ill-advised, I wanted to create a custom component that would have its own internal styles that could be overridden, essentially `({ sx = {}, ...props }) => <Text sx={ ...rootStyles, ...sx } ...props />`. This did not work when using the jsx pragma, which leads me to think babel is actually wiping away my non-standard sx prop - I can even log it and see that it is always `{}`. Is this assumption correct? I can probably just work around it using variants and plain text in any case, but wondering.

dburles:
1. Correct, though there's no reason not to use it as you'll probably be mixing them with your own components or elements.

2. All you need to do is ensure that the `className` prop is passed through to the child element.

```js
const Button = (props) => {
  return (
    <button
      {...props}
      sx={{
        backgroundColor: 'red',
        padding: 3,
      }}
    />
  );
};

export default Button;
```

erikdstock:
Ah, thank you. I probably would have continued overlooking that. Caught some bugs, forgotten spread operators mostly thanks to this. For the record this is what I was trying to do to match a legacy api- if anyone has thoughts on ways to clean it up more I'd love to hear them:
```tsx
export const StyledText: StyledTextComponent = ({
  bold,
  italic,
  danger,
  faded,
  sx = {},
  children,
  ref,
  ...props
}) => {
  // { color, fontWeight, etc }
  const textOverrides = getTextStyle({ bold, italic, faded, danger })

  return (
    <Text
      as="p"
      sx={{ ...textOverrides, ...sx }}

      // necessary for type-checking: Type 'string' is not assignable to type 'Ref<HTMLDivElement>
      ref={ref as any}
      {...props}
    >
      {children}
    </Text>
  )
}

// usage with text variant
export const Lead: StyledTextComponent = ({ sx, ...props }) => {
  return <StyledText as="h3" variant="lead" sx={sx} {...props} />
}

// usage with an html tag
const aStyles = themeGet<SystemStyleObject>("styles.a")
export const StyledA: StyledTextComponent<React.ComponentProps<"a">> = ({
  sx = {},
  ...props
[...]

PR #1494, opened 2021-02-08
docs(variants): Clearly explain variant notation

lachlanjc:
At @erikdstock’s excellent suggestion, clarifying how the notation for accessing variants works.

### DIFF
--- a/packages/docs/src/pages/guides/variants.mdx
+++ b/packages/docs/src/pages/guides/variants.mdx
@@ -30,21 +30,40 @@ For example, you can define `primary` and `secondary` variants for buttons and u
 }
 ```
 
-With the `theme` object above, the `buttons` variants can be referenced in the `sx` prop.
+With the `theme` object above, the `buttons` variants can be referenced by any tag through the `sx` prop.
 
 ```jsx
 <button sx={{ variant: 'buttons.primary' }} />
 ```
 
-When using the built-in [`Button` component](/components/button), you can set these styles with the `variant` prop.
+When using the [built-in components](/components), you can use the `variant` prop directly on components,
+instead of inside the `sx` object.
+Most [components have their own variant group](/components/variants)
+(e.g. the [`Button` component](/components/button) uses `buttons`),
+and some also have a default variant they’ll utilize from that variant group (e.g. `Button` uses `primary`).
+
+This means using `<Button />` without specifying a variant has the same result as the snippet above,
+since the default variant name (`primary`) matches a variant name inside the `buttons` group.
+If you want a different variant from the same group, you can give its key without specifying the group:
 
 ```jsx
 <Button variant="secondary" />
 ```
 
+You can also use variants outside of a component’s default variant group with dot notation.
+This is especially useful for combining layout components without further nesting your DOM,
+such as adding container styles to a grid or flexbox. By using the [`Grid` component](/components/grid)
+here, you’re still able to use `Grid`’s props, and the variant styles are combined with the grid’s.
+
+```jsx
+<Grid variant="layouts.container" columns={3} gap={3} />
+```
+
 <Note>
 
 Variants can use any name you choose, and deeply nested objects can be accessed with dot notation.
+You could use this to organize more complex variants, like `<Button variant="outline.small" />`,
+which would access the `buttons.outline.small` object from the theme.
 
 </Note>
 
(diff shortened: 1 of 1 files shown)


erikdstock:
Looks great! I'm not sure how useful any of this would be useful for documentation but I'd like to clarify - styles are applied and overridden in this order:

1. `theme.styles.root`: automatically applied from the `body` element
2. `theme.styles.<tagName>` for MDX or `<Themed[tagName] / > or `Styled.[tagname]`  components
3. variants (probably best not to mix them):
  - built-in `variant` prop for theme-ui/components`sx` eg `<Button variant="outline"`
  - `sx` prop `variant` key with full path eg `<div sx={{variant: "buttons.outline"}}`
4. `sx` prop styles
5. [(maybe an anti-pattern)](https://github.com/system-ui/theme-ui/issues/839) `className` manually passed down from custom parent components, eg 
```tsx
const CustomButton = ({ children, className }) => (
  <Button
    className={className}
    variant="primary"
    sx={{ borderColor: "purple" }}
  >
    {children} (but custom)
  </Button>
)

// sx is converted to className
<CustomButton sx={{p: 5}}>Big Button</ CustomButton>
```

Finally, within the `theme-ui/components`: `variant` prop, which may have its own special key/default (eg `theme.buttons`), using a full theme path with the dot is an option as well- eg `<Button variant="secondary"` is equivalent to `variant="buttons.secondary"` and `<Button variant="cards.primary"` or `variant="styles.button"` would also be permitted, if not advisable.

[Example codesandbox](https://codesandbox.io/s/theme-ui-variants-and-sx-cc2ue?file=/pages/index.js)

I'm also wondering if there are any identified anti-patterns - In the example above radical departures like adding link styles to a button don't seem to work, and I've also found that there is always a balance to strike as far as what belongs isolated in a theme file vs declared where it is used or where some defaults are applied.

Another thought i'd be willing to take on - it would be nice to add doc comments for default theme keys to the custom components.

hasparus (approved):
Great improvement, folks.

I'm approving both the current addition and any changes / more content based on Erik's comment.

lachlanjc:
Thanks @erikdstock! I added a section about `styles` linking to various places those are used. I’m going to merge this now as I don’t want making this perfect to interfere with getting something better live, but we could definitely keep making this better. Appreciate the feedback as always.

packages/docs/src/pages/guides/variants.mdx
```diff
@@ -162,7 +165,19 @@ These variants can then be used in the `theme.styles` object, with the `sx` prop
 />
 ```
 
-## Themeable Layout Components
+## `styles` variants
+
+Variants inside the `styles` object, while usable through the same mechanisms as regular variants,
+are also used for other Theme UI APIs.
+
+- The entire [`styles` object](/theme-spec#styles) can style child HTML elements or MDX content,
```
hasparus, line 173: Hey @lachlanjc, I'm not sure if this bullet doesn't "conflict" with the last one.
BaseStyles is used to style embedded HTML content, but MDX is styled with `Themed` (i.e. `##` maps to `Themed.h2`)

lachlanjc: Hmm. Using Next.js that's not the case (MDX is treated the same as HTML children, so you need BaseStyles), so I've never experienced that. I guess we can just remove the MDX reference for now?

hasparus: > MDX is treated the same as HTML children

That MDXProvider rendered by our ThemeProvider doesn't work in Next? (We pass an object with components through context)

lachlanjc: Nope: https://github.com/system-ui/theme-ui/issues/1130

hasparus: Doesn't a new version of MDX packages specified in resolutions solve that issue?

https://github.com/system-ui/theme-ui/issues/1130#issuecomment-678727533

lachlanjc: Perhaps! Forgot about that & have never tried it

hasparus: > I guess we can just remove the MDX reference for now?

Yeah, this sound like a good idea :D

lachlanjc: https://github.com/system-ui/theme-ui/commit/fbb3451d43457ba041528c2bdb5ae9a602c8ff5b

erikdstock:
Thanks yall, this is great!

PR #1495, opened 2021-02-08
docs(styled-components): Clarify & update instructions

lachlanjc:
This page has been a bit unclear, & we’re no longer worked on Styled Components migration paths, so clarifying those instructions.

### DIFF
--- a/packages/docs/src/pages/guides/styled-components.mdx
+++ b/packages/docs/src/pages/guides/styled-components.mdx
@@ -8,7 +8,7 @@ Theme UI itself doesn’t expose an API for styled components, but works seamles
 the `styled` API from the [@emotion/styled][] package. Components written with it
 should have access to the same theming context that Theme UI uses.
 
-Instead of using the `ThemeProvider` component from `emotion-theming`,
+Instead of using the `ThemeProvider` component from `@emotion/react`,
 import and use the Theme UI provider at your app’s root level.
 
 ```jsx
--- a/packages/docs/src/pages/guides/styled-components.mdx
+++ b/packages/docs/src/pages/guides/styled-components.mdx
@@ -4,8 +4,12 @@ title: 'Styled Components'
 
 # Styled Components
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/guides/styled-components.mdx
```diff
@@ -32,16 +37,20 @@ import styled from 'styled-components'
...
 ```
 
+[@emotion/styled]: https://emotion.sh/docs/styled
+[styled components]: https://styled-components.com
+
 ## Using the `sx` prop
 
-To avoid the need for an additional dependency,
-you can use the `sx` prop to create custom styled components.
+Theme UI is actively working on removing its internal dependency on `@emotion/styled` to reduce bundle size.
```
hasparus, line 45: > Theme UI is actively working on removing its internal dependency on `@emotion/styled` to reduce bundle size.

I'm not sure if this information is relevant here.

packages/docs/src/pages/guides/styled-components.mdx
```diff
@@ -4,8 +4,12 @@ title: 'Styled Components'
 
 # Styled Components
 
-If you're already using the `styled` API from the [@emotion/styled][] package, those components should have access to the same theming context that Theme UI uses.
-Instead of using the `ThemeProvider` component from `emotion-theming`, you can import and use the Theme UI provider.
+Theme UI itself doesn’t expose an API for styled components, but works seamlessly with
+the `styled` API from the [@emotion/styled][] package. Components written with it
+should have access to the same theming context that Theme UI uses.
+
+Instead of using the `ThemeProvider` component from `emotion-theming`,
```
hasparus:
```suggestion
Instead of using the `ThemeProvider` component from `@emotion/react`,
```

Emotion 11 should be the most popular already.

hasparus (approved):

packages/docs/src/pages/guides/styled-components.mdx
lachlanjc: I debated, but I want people who set them up now to know there's a breaking change coming in the next few months

PR #1499, opened 2021-02-13
Loosening ResponsiveStyleValue to accept false

ed0wolf:
I have a bunch of code that does something like:
```js
sx={{ margin: [0, isTrue && 4] }}
```

The current types don't allow this because `isTrue && 4` resolved to type `false | number` and `false` is not accepted as `ResponsiveStyleValue`. With the current types you need to write `isTrue ? 4 : undefined`.

Any chance we could loosen the types to allow for this kind of short form?

### DIFF
--- a/packages/css/src/types.ts
+++ b/packages/css/src/types.ts
@@ -4,6 +4,8 @@ import type {} from '../emotion-theme'
 
 type StandardCSSProperties = CSS.Properties<number | string>
 
+type Empty = undefined | null | false
+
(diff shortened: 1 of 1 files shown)


lachlanjc:
Seems good to me!

hasparus:
Thanks for the PR, @ed0wolf.
I'm gonna merge it in a moment, update changelog and release it as part of 0.6.0-alpha.7.

PR #1560, opened 2021-03-08
Bugfix: remove Paragraph hardcoded style

### LINKED ISSUE #1476: Remove sx props from Paragraph component
https://github.com/system-ui/theme-ui/blob/29b9b37f37b22924c5afe516972acc0c02266d71/packages/components/src/Paragraph.js#L14-L22

While I appreciate the intent, this makes it really tough to use `Paragraph` except for one very specific use case. It takes a specificity hack to defeat it using the theme:

```js
export const theme = {
  text: {
    paragraph: {
      ':not(_)': {
        my: 3,
        maxWidth: 'unset'
      }
    }
  }
}

```

I think the `margin` reset probably belongs in a base theme, and the `maxWidth` is an interesting idea that could be mentioned in the documentation or included in a paragraph variant in one of the themes.

flo-sch:
- fix(components): remove hardcoded Paragraph style
  - remove hardcoded sx props (such style is easy to opt-in with the theme / variants, but harder to opt-out)

Fixes #1476

Ping @atanasster since you commented on the original PR, just for awareness and potential feedback?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v0.6.0 UNRELEASED
 
+- Paragraph component's hardcoded responsive style has been removed (issue #1476)
+
 ## v0.6.0-alpha.7 2021-02-15
 
 - **Breaking TypeScript**: Known colors (_primary_, _text_, _background_, _accent_, _secondary_) in `ColorMode` can now be nested scales.
--- a/packages/components/src/Paragraph.js
+++ b/packages/components/src/Paragraph.js
@@ -2,7 +2,7 @@ import React from 'react'
 import Box from './Box'
 
 export const Paragraph = React.forwardRef(function Paragraph(
-  { sx, ...props },
+  props,
   ref
 ) {
   return (
@@ -11,15 +11,6 @@ export const Paragraph = React.forwardRef(function Paragraph(
       as="p"
       variant="paragraph"
       {...props}
-      sx={{
-        // reset margin by default: avoid relying on user-agent margins (not aware of theme-ui space scale)
-        margin: 0,
-        // set a max-width: avoid full-page paragraphs
-        '@media screen and (min-width: 36em)': {
-          maxWidth: '48rem',
-        },
-        ...sx,
-      }}
       __themeKey="text"
       __css={{
         fontFamily: 'body',
--- a/packages/components/src/Paragraph.js
+++ b/packages/components/src/Paragraph.js
@@ -2,7 +2,7 @@ import React from 'react'
 import Box from './Box'
 
 export const Paragraph = React.forwardRef(function Paragraph(
-  props,
+  { sx, ...props },
   ref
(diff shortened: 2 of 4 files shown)


lachlanjc:
Let's absolutely axe the max width, but I think we should keep the margin: 0 we've currently got. Otherwise, as the source comment indicates, you can use a Theme UI component & end up with non-theme spacing from the user agent styles.

flo-sch:
Yeah, also uncertain about the margin...
I'm struggling overall with SoC here, is theme-ui intention to do CSS-reset, or should that be the role of a preset // something else than theme-ui?

In the issue, part of the feedback from @aaronadamsCA is:

> I think the margin reset probably belongs in a base theme

But theme-ui does not have a "base theme", that's its goal actually (not to have one, allowing anyone to provide such base style for their UI)
At the same time, it is trivial to opt in AND out of that margin with the default `paragraph` variant 🤷‍♂️

hasparus:
> you can use a Theme UI component & end up with non-theme spacing from the user agent styles.

Only if you change the variant to one you added, innit? Good to merge from my perspective if I'm not mistaken here.

hasparus (approved):

lachlanjc:
I'm okay to go along with removing the margin, but I don't think it's worth breaking people's designs with this change, especially because I would see it as a bug in Theme UI that components have default margins that are not coming from my theme

flo-sch:
> Only if you change the variant to one you added, innit?

Not necessarily I think, if you do not inherit a preset setting `margin: 0` on that default variant, and we remove it from the component itself (as this PR currently does), then "Paragraph" rendering a "p" by default would inherit the browser's margin

flo-sch:
Alright, then let's re-add the margin reset, which can easily be opted out by setting whichever margin value on the `text.paragraph` default variant :)

lachlanjc (approved):

hasparus:
And it's merged! Thanks a lot @flo-sch!

vanbujm:
I have been having trouble overriding the margin style.

Can we change:
```javascript
    <Box
      ref={ref}
      as="p"
      variant="paragraph"
      {...props}
      sx={{
        // reset margin by default: avoid relying on user-agent margins (not aware of theme-ui space scale)
        margin: 0,
        ...sx
      }}
      ...
```

To:
```javascript
    <Box
      ref={ref}
      as="p"
      variant="paragraph"
      sx={{
        // reset margin by default: avoid relying on user-agent margins (not aware of theme-ui space scale)
        margin: 0,
        ...sx
      }}
      {...props}
      ...
```

So that you can use margin props (m, my, mx, etc) to override margin?

lachlanjc:
Seems fine to me @vanbujm! PR welcome

vanbujm:
> Seems fine to me @vanbujm! PR welcome

[PR](https://github.com/system-ui/theme-ui/pull/1658) is up.

PR #1624, opened 2021-04-02
Updated div to box as it doesn't work using div.

aimeetacchi:
I've updated this file to use Box for theme ui, as the docs say use div, when I tried this it didn't work. it worked with Import Box and use Box

### DIFF
--- a/packages/docs/src/pages/sx-prop.mdx
+++ b/packages/docs/src/pages/sx-prop.mdx
@@ -13,11 +13,11 @@ The `sx` prop lets you add any valid CSS to an element,
 while using values from your theme to keep styles consistent.
 You can think of the style object that the `sx` prop accepts as a _superset_ of CSS.
 
-```jsx
-/** @jsxImportSource theme-ui */
+```/** @jsx jsx */
+import { jsx, Box } from 'theme-ui'
 
 export default (props) => (
-  <div
+  <Box
     {...props}
     sx={{
       // values referencing scales defined in a theme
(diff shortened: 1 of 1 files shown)


hasparus:
Thanks for the PR, @aimeetacchi! I feel this should be a discussion or a bug report, though. Configuring JSX pragma with either new (@jsxImportSource) or classic runtime (@jsx) ([see babeljs.io/docs/en/babel-preset-react](https://babeljs.io/docs/en/babel-preset-react#importsource)) should allow you to use `sx` prop on any intrinsic element (like div).

Could you reproduce the problem on CodeSandbox or share your Babel/TS config here?

Do you have any ideas how we could explain JSX transpilation and JSX runtimes in the docs so they're less confusing?

hasparus (changes requested):
I'll set it to _Request changes_ just so we don't merge it by accident.

aimeetacchi:
> Thanks for the PR, @aimeetacchi! I feel this should be a discussion or a bug report, though. Configuring JSX pragma with either new (@jsxImportSource) or classic runtime (@jsx) ([see babeljs.io/docs/en/babel-preset-react](https://babeljs.io/docs/en/babel-preset-react#importsource)) should allow you to use `sx` prop on any intrinsic element (like div).
> 
> Could you reproduce the problem on CodeSandbox or share your Babel/TS config here?
> 
> Do you have any ideas how we could explain JSX transpilation and JSX runtimes in the docs so they're less confusing?

Hello, I guess may it is a bug, as it seems to work with div now. But I did have to use that @jsx at the top in a comment. I'm new to Theme UI. But I thought I would add it to the document.

hasparus:
It's all good! All contributions are welcome, and even PRs which don't end up merged are helpful.

The docs assume a familiarity with Babel JSX runtimes, what was okay when there was only one runtime (now called "classic" with `@jsx` comment), but got more complex when new runtime (with `@jsxImportSource` pragma) was introduced. It seems to be one of main friction points in Theme UI and we definitely need to improve the docs to ease the adoption.

lachlanjc:
Though we'd need to slightly tweak the syntax in this PR, I'm actually in favor of suggesting using Components to most users. I think it's more consistent with how other libraries work, & for most use cases they give you a huge head start on styling & layout, without the complexity of the JSX imports & understanding pragmas & Babel & all.

hasparus:
> I'm actually in favor of suggesting using Components to most users

@lachlanjc I think we'd have to ensure that code-splitting problems with components don't occur anymore and write some tests for that.

hasparus:
Closing as it's inactive.

PR #1640, opened 2021-04-12
chore: More explicit Gatsby peer dep

LekoArts:
Hello <USER>, Gatsby maintainer here 👋 

While looking at the plugins I noticed that the `peerDependency` on `gatsby` is set incorrectly. In https://github.com/system-ui/theme-ui/commit/55dab22f6c17286dbbfd806ba281f4d310da4548 it was changed to `>=2`. We're in the process of providing more helpful information on the `/plugins` page of our website and for that we need plugins to set their `peerDependencies` correctly/more specific.

At the moment you say that _any_ version (v3, v4, v5, v6, etc.) will work with this even when in between we'd introduce breaking changes.

Making it more explicit is the safer choice (in the future you'll be able to mark breaking changes with explicit version ranges) and will allow us to display the plugin as compatible to versions X, Y, Z.

Thanks!

P.S.: Please let me know where I should change the `CHANGELOG`? I don't see a `UNRELEASED` bulletpoint anymore :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## UNRELEASED
+
+- Set `gatsby` peerDependency more explicit to `^2.0.0 || ^3.0.0`. [#1640](https://github.com/system-ui/theme-ui/pull/1640) ([@LekoArts](https://github.com/LekoArts))
+
 ## v0.6.2 (Mon Apr 05 2021)
 
 ### 🐛 Bug Fix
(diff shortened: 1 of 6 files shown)


lachlanjc (approved):
Sounds great, thank you!

lachlanjc:
On the CHANGELOG, I believe you can just add that heading to the top of the document. Thanks!

LekoArts:
Done 👍

hasparus:
Just FYI folks, you don't need to edit the changelog anymore. I've set up auto which should take care of this before release (and even if it breaks we can run `auto changelog` so we should be good).

hasparus:
:rocket: PR was released in [`v0.7.0-develop.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.7.0-develop.0) :rocket:

PR #1664, opened 2021-04-20
Adding `scroll-margin` props

dios-david:
Hello 👋 

Recently I started to use `scrollMargin` in one of my ThemeUI projects, and I noticed the value is not resolved to a `space` value from the theme definition.
At the end I could solve it by using a theme callback, but I thought it would be a good addition for the project to add a proper support for `scroll-margin*` props.

Please let me know if I missed something!


lachlanjc (approved):
Gorgeous, thanks so much!

hasparus:
A great PR, @dios-david. Defnitely a good addition.

hasparus:
Released with 0.7.3

PR #1665, opened 2021-04-21
Return negative number from lookup if theme value is a number

kenny-f:
On a previous pull request https://github.com/system-ui/theme-ui/pull/865 I added the feature to allow negative lookups of theme objects to be resolved.

eg:
given `{ space: { s: '8px' }` and `sx={{ margin: '-s' }}` would be resolved to `margin: '-8px'`

however when the theme value is a `number` it resolves to `'-8'` (string) instead of `-8` (number) and then in the browser you see `margin: -80` instead of `margin: -80px`

This pr will fix this by checking if the theme value type is a number, and if it is will return a number.

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -246,6 +246,9 @@ const positiveOrNegative = (scale: object, value: string | number) => {
     if (typeof value === 'string' && value.startsWith('-')) {
       const valueWithoutMinus = value.substring(1)
       const n = get(scale, valueWithoutMinus, valueWithoutMinus)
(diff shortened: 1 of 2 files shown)


dios-david (approved):
Nice catch!

lachlanjc (approved):
Appreciated!

hasparus:
Solid PR @kenny-f! Thanks!

hasparus:
Released with 0.7.3

kenny-f:
> Released with 0.7.3

@hasparus Thanks!!!

PR #1672, opened 2021-04-25
fix: prevent unwanted rerender

Bram-Zijp:
I noticed that after initial render of the `ThemeColorProvider`, it was re-rendering.This MR prevents this re-render from taking place by populating initial color state with the localStorage cache state on first render.

Edit: It now just uses the classNames to populate initial state.

### DIFF
--- a/packages/color-modes/src/index.tsx
+++ b/packages/color-modes/src/index.tsx
@@ -69,6 +69,7 @@ const getModeFromClass = (): string | undefined => {
 }
 
 const useColorModeState = (theme: Theme = {}) => {
+  let stored: ReturnType<typeof storage.get>
   let [mode, setMode] = useState(() => {
     const modeFromClass = getModeFromClass()
     if (modeFromClass) {
@@ -78,22 +79,17 @@ const useColorModeState = (theme: Theme = {}) => {
     const preferredMode =
       theme.useColorSchemeMediaQuery !== false && getPreferredColorScheme()
 
-    return preferredMode || theme.initialColorModeName
+    stored = theme.useLocalStorage !== false ? storage.get() : undefined;
+
+    return stored || preferredMode || theme.initialColorModeName
   })
 
   // on first render, we read the color mode from localStorage and
   // clear the class on document element body
   useEffect(() => {
-    const stored = theme.useLocalStorage !== false && storage.get()
-
     if (typeof document !== 'undefined') {
       document.documentElement.classList.remove('theme-ui-' + stored)
       document.body.classList.remove('theme-ui-' + stored)
-    }
-
-    if (stored && stored !== mode) {
-      mode = stored
-      setMode(stored)
     }
   }, [])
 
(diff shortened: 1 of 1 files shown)


lachlanjc:
Super appreciate the PR! I don't 100% understand what's happening in the new code though: why do we need to remove the body class on first render?

Bram-Zijp:
At the [bottom](https://github.com/system-ui/theme-ui/pull/1672/files#diff-5ded92a0b88f8632b50f513c22d8317d0ae7ba32f833d9f2c689d4edbfa7c5e6L194) of the changed file, a class is set on the body. This specifies the initial color mode based on the local storage data, I believe. To be honest, I'm not sure if we need this "body className" in the first place. Perhaps this is something to look into. My gut feeling is that this body className is used, to consume server side rendered styles, bound to this body className.

In terms of change, the code does the exact same thing as what was implemented, with the exception of the re-render.

Why not remove classes on the useState? I believe this action will mutate the DOM while still in the process of rendering, whereas mutating the dom in the useEffect will happen after rendering of the initial DOM is complete. Correct me if I'm wrong.

packages/color-modes/src/index.tsx
```diff
@@ -81,24 +81,17 @@ const useColorModeState = (theme: Theme = {}) => {
...
-
-    if (typeof document !== 'undefined') {
-      document.documentElement.classList.remove('theme-ui-' + stored)
-      document.body.classList.remove('theme-ui-' + stored)
-    }
-
-    if (stored && stored !== mode) {
-      mode = stored
-      setMode(stored)
+    const classMode = getModeFromClass()
```
Bram-Zijp, line 86: Perhaps it's faster to just attempt to remove classes without checking if they exist.

packages/color-modes/src/index.tsx
```diff
@@ -81,24 +81,17 @@ const useColorModeState = (theme: Theme = {}) => {
     return preferredMode || theme.initialColorModeName
   })
 
-  // on first render, we read the color mode from localStorage and
-  // clear the class on document element body
+  // remove initial color mode classes from the document and body
   useEffect(() => {
-    const stored = theme.useLocalStorage !== false && storage.get()
-
-    if (typeof document !== 'undefined') {
```
Bram-Zijp, line 89: removed because useEffect only runs client side.

Bram-Zijp:
Instead of using [getModeFromClass](https://github.com/system-ui/theme-ui/pull/1672/files#R59), we could specify a global on [the same place we set the theme className, obtained from localStorage.](getModeFromClass). I think this will result in less LoC and better performance (noticable or not).

lachlanjc:
Thanks for the explanation! @hasparus could you take a look?

hasparus:
Thanks for the PR @Bram-Zijp! Nice catch!

@lachlanjc I'll need to take a look at corner cases (how does it look like with SSR? that rerender triggers update) and merge a few outstanding PRs to this part first (fcis' fixes).

Issue #1602, opened 2021-03-26
colors from outside of theme are not recomputed on initial render after Gatsby SSR


ehowey:
**Describe the bug**

Minimal Repo: https://github.com/ehowey/color-mode-bug
Relevant code: https://github.com/ehowey/color-mode-bug/blob/main/src/pages/index.js
Live site: https://condescending-blackwell-39cd84.netlify.app/ 

Previously I could use code like the following to dynamically update images in dark mode, e.g. setting `invert(1)` to invert an image for dark mode. This code is still working fine in development mode but is breaking production mode. I have created a minimal repo and a deploy to Netlify to test this. You can see in the console log it is properly catching the mode as dark but for some reason the text and images are not updating their css based on the color mode.

To test this make sure you are visiting the live site in incognito mode as it will correctly change the colors and image on the second load once the cache is set.

```js
const [mode] = useColorMode()
const isDark = mode === "dark"

<GatsbyImage sx={{ filter: isDark ? "invert(1)" : "none"}} />
<Theme.p sx={{color: isDark ? "tomato" : "cornflowerblue"}}>Some text </Themed.p>
```

**Expected behavior**
Previously `isDark` could be used like a boolean flag in the CSS to toggle different colors based on the color mode.

hasparus:
Thanks for the issue and the reproduction @ehowey!

hasparus:
This is definitely something I broke while fixing color modes flash — previously we always started with default color mode and you could see the colors (except background and text) _blink_, but in this case you need the flash.

When the website SSRs, "mode" is `undefined`, so `isDark` is false. On client side we update all CSS custom properties, so colors from your theme update.

Based on the quick fiddling I gave it today, `sx` is recomputed on the client on rehydration, but it doesn't affect the HTML. 
After changing `color` to a function with console.logs I can see it returns `"tomato"`, but the text is still blue. :thinking:
After setting `key` to Math.random, text is tomato.

> This code is still working fine in development mode but is breaking production mode.

BTW You can set `flags.DEV_SSR` to true in your Gatsby config to see bugs like this in developent.

As a quick workaround, you could add the color to the theme, but I realize this is less than ideal.

ehowey:
Thanks for looking into this.  Just as an FYI - this shows up in development even with the `DEV_SSR` flag set to true, so not a reliable indicator of whether it is working correctly.

fcisio:
@ehowey 
Here is my fix for this.

```js
[...]

PR #1685, opened 2021-04-26
Gatsby SSR colorMode Fix

fcisio:
This PR follows the issue: #1602 

> Also extends on my previous PR #1639 

I wasn't able to repro the issue inside the package, but basically replicated what I do on the user-side.
It should work, and once there is a canary version, I'll be able to test in my own projects.

Let me know what you guys think 👌

---
I additionally fixed some things that were breaking Gatsby builds internally

### DIFF
--- a/packages/color-modes/src/index.tsx
+++ b/packages/color-modes/src/index.tsx
@@ -109,7 +109,7 @@ const useColorModeState = (theme: Theme = {}) => {
   }, [])
 
   // when mode changes, we save it to localStorage
-  React.useEffect(() => {
+  useEffect(() => {
     if (mode && theme.useLocalStorage !== false) {
       storage.set(mode)
     }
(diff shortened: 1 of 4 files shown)


fcisio:
Alright 😅
Diving deeper into this, I realized I don't really have a fix.

I did draft one, but it only works when local storage is turned off.

I'm not sure what causes the issue in the first place, but my guess is it's from `preferredMode`  and `modeFromClass`.

Steps to reproduce:
1. build docs
2. serve docs
3. delete the `theme-ui-color-mode` key from the storage
4. navigate to http://localhost:9000/home/

> A pink div should show up on dark mode (if it's the `preferredMode`). The issue is only present on the first load. If you switch to dark mode, then all works fine.

If I remember right, I first saw this issue surface around the time `useColorSchemeMediaQuery` default was changed to `true`. We could probably trace back what changes are causing this.

If anyone, wants to have a go at this, I'm kinda stuck for now.

hasparus:
Closing, because this PR is a leftover. We fixed this already on `develop` (and **fcisio** reviewed it)

PR #1718, opened 2021-05-05
Docs: update JSX pragma guide with automatic runtime section

### LINKED ISSUE #1335: Document using site-wide JSX pragma configuration
```json
{
  "plugins": [
    [
      "@babel/plugin-transform-react-jsx",
      {
        "throwIfNamespace": false, // defaults to true
        "runtime": "automatic", // defaults to classic
        "importSource": "theme-ui" // defaults to react
      }
    ]
  ]
}
```

https://babeljs.io/docs/en/babel-plugin-transform-react-jsx

flo-sch:
- docs(pragma): add "automatic runtime" section to JSX pragma guide
  - add link to JSX pragma guide from "How it works" guide

Fixes #1335

### DIFF
--- a/packages/docs/src/pages/guides/how-it-works.mdx
+++ b/packages/docs/src/pages/guides/how-it-works.mdx
@@ -4,13 +4,16 @@ title: How it Works
 
 # How it Works
 
-Theme UI combines several libraries together to form a mini-framework for styling applications.
+Theme UI combines several libraries together to form a mini-framework for
+styling applications.
 
 Theme UI is built with:
 
 - [Emotion][]: used to generate isolated CSS with theming context
-- [MDX][]: used to provide custom Emotion styled components to MDX documents, without polluting the global CSS scope
-- [Typography.js][]: optionally used for creating rich typographic styles with a simple, high-level API
+- [MDX][]: used to provide custom Emotion styled components to MDX documents,
+  without polluting the global CSS scope
+- [Typography.js][]: optionally used for creating rich typographic styles with a
+  simple, high-level API
 
 [emotion]: https://emotion.sh
 [mdx]: https://mdxjs.com
@@ -19,13 +22,16 @@ Theme UI is built with:
 ## ThemeProvider
 
 The Theme UI `ThemeProvider` component is a wrapper around MDX's `MDXProvider`,
-which adds custom React components to context,
-and Emotion's `ThemeProvider`, which adds the `theme` object to context for use with Emotion.
+which adds custom React components to context, and Emotion's `ThemeProvider`,
+which adds the `theme` object to context for use with Emotion.
 
-The Theme UI `ThemeProvider` includes a default `components` object with styled components that pick up values from the `theme.styles` object.
-For example, `theme.styles.h1` will be used in its `components.h1` component and rendered in MDX documents.
+The Theme UI `ThemeProvider` includes a default `components` object with styled
+components that pick up values from the `theme.styles` object. For example,
+`theme.styles.h1` will be used in its `components.h1` component and rendered in
+MDX documents.
 
-For illustrative purposes, the `ThemeProvider` renders something like the following:
+For illustrative purposes, the `ThemeProvider` renders something like the
+following:
 
 ```jsx
 // for demonstration only – this does not reflect the actual source code
@@ -50,18 +56,20 @@ export default ({ components, theme, children }) => (
 
 ## Custom Components
 
-When using the `components` prop to add custom MDX components,
-each component is wrapped with the `sx` prop to allow the component to pick up values from the `theme.styles` object.
+When using the `components` prop to add custom MDX components, each component is
+wrapped with the `sx` prop to allow the component to pick up values from the
+`theme.styles` object.
 
 ## JSX Pragma
 
-The Theme UI JSX function wraps the Emotion JSX function, converting the `sx` prop to a call to `@theme-ui/css`.
-The two examples below yield the same results:
+The Theme UI JSX function wraps the Emotion JSX function, converting the `sx`
+prop to a call to `@theme-ui/css`. The two examples below yield the same
+results:
 
 ```jsx
 // with Emotion's JSX function
 // this is typically handled with the Emotion Babel preset
-/** @jsxImportSource @emotion/reaact */
+/** @jsxImportSource @emotion/react */
 import { css } from 'theme-ui'
 
 export default (props) => (
@@ -91,3 +99,10 @@ export default (props) => (
   />
 )
 ```
+
+<Note>
+
+Confused about this `@jsxImportSource` comment? Read more about
+[JSX Pragma](/guides/jsx-pragma).
+
+</Note>
--- a/packages/docs/src/pages/guides/jsx-pragma.mdx
+++ b/packages/docs/src/pages/guides/jsx-pragma.mdx
@@ -47,7 +47,7 @@ Technically Babel also calls `jsxs` from `react/jsx-runtime` and `jsxDEV` from
 
 To change the underlying create element functions, you can either add an option
 to the Babel plugin or you can set a _pragma comment_ at the beginning of a
-module. Changing the function import source in the Babel plugin will transform
+module. Changing the function import source in Babel config will transform
 all JSX in an application into the same set of functions. Using a pragma comment
 limits the change to only the modules that it's added to. This lets you default
 to the React `jsx` function in most places and use custom functions only where
--- a/packages/docs/src/pages/guides/jsx-pragma.mdx
+++ b/packages/docs/src/pages/guides/jsx-pragma.mdx
@@ -4,13 +4,15 @@ title: JSX Pragma
 
 # JSX Pragma
 
-Theme UI uses custom JSX functions and JSX import source comments to allow you to style elements with values from your theme using the [`sx` prop][].
+Theme UI uses custom JSX functions and JSX import source comments to allow you
+to style elements with values from your theme using the [`sx` prop][].
 
 ## What is JSX
 
-JSX is an XML-like syntax extension to JavaScript.
-It's a syntax sugar usually used for React's `jsx` function.
-You don't need to write JSX to use React, but it's meant to make code more readable, especially for tree structures with attributes.
+JSX is an XML-like syntax extension to JavaScript. It's a syntax sugar usually
+used for React's `jsx` function. You don't need to write JSX to use React, but
+it's meant to make code more readable, especially for tree structures with
+attributes.
 
 Given the following JSX:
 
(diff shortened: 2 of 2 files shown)


packages/docs/src/pages/guides/jsx-pragma.mdx
```diff
@@ -55,3 +64,120 @@ The preferred way of using these functions is by adding the custom pragma commen
...
+}
+```
+
+NOTE: this requires
+[babel >= 7.9.0](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#react-automatic-runtime)
+
+### Using TypeScript
+
+If you use TypeScript to transpile your source code with `tsc` (or only
+typecheck), or for instance to run tests with `ts-jest`
```
hasparus, line 132: Sorry mate, I was under impression we already merged this PR.

Could we write something like "If you use TypeScript to transpile your JSX"?
A lot of people are using TypeScript together with Babel, and they don't need to configure jsxImportSource globally twice (you only need it in Babel config if you're using Babel)

packages/docs/src/pages/guides/jsx-pragma.mdx
```diff
@@ -55,3 +64,120 @@ The preferred way of using these functions is by adding the custom pragma commen
...
+    "jsxImportSource": "theme-ui"
+  }
+  // ...
+}
+```
+
+NOTE: this requires
+[TypeScript >= 4.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#react-17-jsx-factories)
+
+NOTE: in order to typecheck `sx` using automatic runtime, you will also need to
```
hasparus: Is this also the case when `"jsxImportSource": "theme-ui"` is configured in TypeScript config?

packages/docs/src/pages/guides/how-it-works.mdx
```diff
@@ -50,18 +56,20 @@ export default ({ components, theme, children }) => (
...
-The Theme UI JSX function wraps the Emotion JSX function, converting the `sx` prop to a call to `@theme-ui/css`.
-The two examples below yield the same results:
+The Theme UI JSX function wraps the Emotion JSX function, converting the `sx`
+prop to a call to `@theme-ui/css`. The two examples below yield the same
+results:
 
 ```jsx
 // with Emotion's JSX function
 // this is typically handled with the Emotion Babel preset
-/** @jsxImportSource @emotion/reaact */
```
hasparus, line 64: :D

packages/docs/src/pages/guides/jsx-pragma.mdx
```diff
@@ -34,19 +36,26 @@ jsx('div', {
...
+`react/jsx-dev-runtime` in some cases, but the concept still holds.
 
-To change the underlying create element functions, you can either add an option to the Babel plugin or you can set a _pragma comment_ at the beginning of a module.
-Changing the function import source in the Babel plugin will transform all JSX in an application into the same set of functions.
-Using a pragma comment limits the change to only the modules that it's added to.
-This lets you default to the React `jsx` function in most places and use custom functions only where you need it,
-giving the author more control over where it's used.
+To change the underlying create element functions, you can either add an option
+to the Babel plugin or you can set a _pragma comment_ at the beginning of a
+module. Changing the function import source in the Babel plugin will transform
```
hasparus:
```suggestion
module. Changing the function import source in Babel config will transform
```

packages/docs/src/pages/guides/jsx-pragma.mdx
```diff
@@ -4,13 +4,15 @@ title: JSX Pragma
...
-Theme UI uses custom JSX functions and JSX import source comments to allow you to style elements with values from your theme using the [`sx` prop][].
+Theme UI uses custom JSX functions and JSX import source comments to allow you
+to style elements with values from your theme using the [`sx` prop][].
 
 ## What is JSX
 
-JSX is an XML-like syntax extension to JavaScript.
-It's a syntax sugar usually used for React's `jsx` function.
-You don't need to write JSX to use React, but it's meant to make code more readable, especially for tree structures with attributes.
+JSX is an XML-like syntax extension to JavaScript. It's a syntax sugar usually
```
hasparus:
```suggestion
[JSX](https://facebook.github.io/jsx/) is an XML-like syntax extension to JavaScript. It's a syntax sugar usually
```

packages/docs/src/pages/guides/jsx-pragma.mdx
```diff
@@ -4,13 +4,15 @@ title: JSX Pragma
...
+used for React's `jsx` function. You don't need to write JSX to use React, but
```
hasparus:
```suggestion
used for [React's `jsx`](https://reactjs.org/docs/introducing-jsx.html) function. You don't need to write JSX to use React, but
```

packages/docs/src/pages/guides/jsx-pragma.mdx
```diff
@@ -55,3 +64,120 @@ The preferred way of using these functions is by adding the custom pragma commen
...
+          importSource: 'theme-ui', // or '@theme-ui/core'
+          runtime: 'automatic',
+          throwIfNamespace: false,
+        },
+      },
+    ],
+  ],
+  // ...
+}
+```
```
hasparus, line 167:
````suggestion
```

You can read more about [customizing Babel config in Next.js docs](https://nextjs.org/docs/advanced-features/customizing-babel-config).
````

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 1 above)
Comment (flo-sch) [reply packages/docs/src/pages/guides/jsx-pragma.mdx:132]: You still need it for instance for eslint and other tools like `ts-jest` that do not use Babel, even if you use `@babel/preset-typescript` to build, right?

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 2 above)
Comment (flo-sch) [reply packages/docs/src/pages/guides/jsx-pragma.mdx]: Huuuh, I think this is above my TS understanding to be honest, but I can re-try in a project 😁

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 1 above)
Comment (hasparus) [reply packages/docs/src/pages/guides/jsx-pragma.mdx:132]: Good point. I assumed people would use `babel-jest`, but `typescript-eslint` will probably need it in the tsconfig when it's doing typechecking.

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 2 above)
Comment (hasparus) [reply packages/docs/src/pages/guides/jsx-pragma.mdx]: I was under impression that TS correctly found types from configured importSource, but I'll have to double-check.

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 1 above)
Comment (flo-sch) [reply packages/docs/src/pages/guides/jsx-pragma.mdx:132]: Should we still be more explicit here then? I don't mind 🤷

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 2 above)
Comment (flo-sch) [reply packages/docs/src/pages/guides/jsx-pragma.mdx]: Seems like you're right actually, good catch :)
I will remove it

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 1 above)
Comment (hasparus) [reply packages/docs/src/pages/guides/jsx-pragma.mdx:132]: I think it's okay now. @lachlanjc could you take a look at this PR when you have a moment?

File: packages/docs/src/pages/guides/jsx-pragma.mdx (hunk 2 above)
Comment (flo-sch) [reply packages/docs/src/pages/guides/jsx-pragma.mdx]: @hasparus just figured out why I thought this was still necessary:

If you apply `sx` to a non theme-ui component without manually providing `sx` types, then it does not typecheck even with `jsxImportSource: "theme-ui"` inside tsconfig.json (even though sx gets correctly compiled)

```tsx
// StyledForm.tsx
import { Button, Flex, Input, Label } from 'theme-ui';

// Some interface not explicitly typing "sx"
export interface StyledFormProps {
  onChange(data: Record<string, unknown>): void;
}

const StyledForm: React.FC<StyledFormProps> = ({
  onChange,
  ...props
}) => (
  <form {...props}> // spreading props, including sx --> Emotion correctly compiles it to a className
    <Flex sx={{ flexFlow: 'column', p: 's' }}>
      <Label htmlFor="input.test">Test</Label>
      <Input
        id="input.test"
        type="text"
        onChange={
          (event) => onChange({
            test: event.target.value
          })
        }
      />
    </Flex>
  </form>
);

export default StyledForm;
```

TS error when forwarding `sx`

```tsx
// App.tsx
const App = () => (
  <StyledForm
    onChange={console.log}
    sx={{
      // Type '{ /* ... */ sx: { ...; }; }' is not assignable to type 'IntrinsicAttributes & StyledFormProps & { children?: ReactNode; }'.
      //   Property 'sx' does not exist on type 'IntrinsicAttributes & StyledFormProps & { children?: ReactNode; }'.
      'px': 'm'
    }}
  />
);
```

This can be easily fixed by adding `sx?: ThemeUIStyleObject` to `StyledFormProps` in that case, just slightly sad that we get a TS error even though the className works fine IMHO, but I don't mind adding the types wherever it is needed

Could/Should the automatic `sx` typings be added to all JSX components instead of only theme-ui ones, or is that anti-pattern? 🤔

beerose: Hey @flo-sch 👋 @hasparus doesn't have access to his computer, and he asked me to write this 😄 

Theme UI types add `sx` prop only to components which props already accept `className`, as `sx` transpiles to `className`. 

The code responsible for this is at [packages/core/src/jsx-namespace.ts#L3](https://github.com/system-ui/theme-ui/blob/develop/packages/core/src/jsx-namespace.ts#L3).

Here's a CodeSandbox, so you can try it out: https://codesandbox.io/s/frosty-https-wpp7f?file=/src/index.tsx

flo-sch: Hi @beerose, nice to meet you and thanks for relaying the answer 😁
(not sure if the sandbox link was the correct one though, seems more about GraphQL in there?) 

That make sense in term of typings I guess, I did reproduce it (even without automatic runtime):
https://codesandbox.io/s/sx-props-typings-automatic-runtime-dte1v?file=/src/App.tsx

My only claim was that, since Emotion successfully compiles it anyway, it felt like a "false-negative" from TS to complain on missing `sx` types in that case (in that App.tsx, there is a TS complain on the first `<Form sx={{ }} />` since it does not add types for className or sx)

But I guess this is another discussion here, not related to automatic runtime in itself

beerose: Hey! I'm not sure how it happened but yes, the link was wrong 🤦‍♀️
Here's the valid one: https://codesandbox.io/s/affectionate-architecture-utrwm?file=/src/index.tsx
Sorry!

hasparus: > My only claim was that, since Emotion successfully compiles it anyway, it felt like a "false-negative" from TS to complain on missing sx types in that case (in that App.tsx, there is a TS complain on the first <Form sx={{ }} /> since it does not add types for className or sx)

IMHO this isn't a false negative — `sx` is compiled to `className` which is not handled so we see almost the same error as if we passed a `className`, meaning "this component cannot be styled from the outside".

hasparus (approved):

hasparus:
:rocket: PR was released in [`v0.10.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.10.0) :rocket:

PR #1775, opened 2021-05-25
fix: Allow overriding Paragraph margins via theme

### LINKED ISSUE #1774: Paragraph margin can't be overridden by theme
**Describe the bug**
Paragraph margins can't be changed via theming because the `m={0}` value is always added after applying the theme. This was first fixed in #1476, but reintroduced by #1560.

**To Reproduce**
I created a sandbox in https://codesandbox.io/s/aged-silence-42b7m
See the output - all 4 paragraphs should have margins applied.

**Expected behavior**
Margins should be set to 0 by default. But no matter what method of styling you prefer, they should always be overrideable. 

**Screenshots**
image

**Additional context**
It seems that margin:0 is set twice. Maybe it helps to remove the `m={0}` completely because it's set anyway.
image

bernharduw:
This fixes #1774 - I found that margin: 0 is set by default (coming from Box maybe? Not 100% sure).

So if we don't set the margin for Paragraph manually, the specificity issues disappear.

I also added two tests using a theme override - the other two cases are already covered by the existing tests.

# Version

Published prerelease version: `v0.9.1-develop.0`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - `@theme-ui/components`
    - fix: Allow overriding Paragraph margins via theme [#1775](https://github.com/system-ui/theme-ui/pull/1775) ([@bernharduw](https://github.com/bernharduw))
  
  #### Authors: 1
  
  - Bernhard Gschwantner ([@bernharduw](https://github.com/bernharduw))
</details>

### DIFF
--- a/packages/components/test/__snapshots__/index.js.snap
+++ b/packages/components/test/__snapshots__/index.js.snap
@@ -1097,7 +1097,6 @@ exports[`Paragraph renders 1`] = `
   font-family: body;
   font-weight: body;
   line-height: body;
-  margin: 0;
 }
 
 <p
--- a/packages/components/src/Paragraph.js
+++ b/packages/components/src/Paragraph.js
@@ -1,17 +1,12 @@
 import React from 'react'
 import Box from './Box'
 
-export const Paragraph = React.forwardRef(function Paragraph(
-  props,
-  ref
-) {
+export const Paragraph = React.forwardRef(function Paragraph(props, ref) {
   return (
     <Box
       ref={ref}
       as="p"
       variant="paragraph"
-      // reset margin by default: avoid relying on user-agent margins (not aware of theme-ui space scale)
-      m={0}
       __themeKey="text"
       __css={{
         fontFamily: 'body',
--- a/packages/components/test/index.js
+++ b/packages/components/test/index.js
@@ -341,6 +341,26 @@ describe('Paragraph', () => {
     const json = renderJSON(<Paragraph m={margin} />)
     expect(json).toHaveStyleRule('margin', margin)
   })
+
+  test('renders with theme override', () => {
+    const margin = '8px'
+    const json = renderJSON(
+      <ThemeProvider theme={{ text: { paragraph: { margin } } }}>
+        <Paragraph />
+      </ThemeProvider>
+    )
+    expect(json).toHaveStyleRule('margin', margin)
+  })
+
+  test('renders with theme variant', () => {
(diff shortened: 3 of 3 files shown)


hasparus:
Thanks for the PR @bernharduw!

hasparus:
Doesn't this break the scenario when users depend on margin being set to 0 by default? (See snapshot in the diff.)

packages/components/test/__snapshots__/index.js.snap
```diff
@@ -1097,7 +1097,6 @@ exports[`Paragraph renders 1`] = `
   font-family: body;
```
bernharduw, line 1097: The snapshot still has `margin: 0` in it, but at the very beginning.

bernharduw:
I think that way the effective margin on paragraphs is still 0, but can be overridden by any styling method, regardless of theme, styled-system-style or direct sx overrides. The tests still all pass, only the duplicate margin: 0 is reduced to a single occurrence.

The margin:0 in line 1095 comes from the Box which sets margin:0 as a default, so we should be safe here.

hasparus:
Thanks! Yup, I can see that first `margin: 0`. Sorry!

hasparus (approved):

hasparus:
:rocket: PR was released in [`v0.9.1-develop.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.9.1-develop.0) :rocket:

hasparus:
:rocket: PR was released in [`v0.9.1`](https://github.com/system-ui/theme-ui/releases/tag/v0.9.1) :rocket:

hasparus:
:rocket: PR was released in [`v0.9.2-develop.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.9.2-develop.0) :rocket:

sergeylukin:
WD for removing duplication. I, however, struggle to understand the purpose of Box's default margin: 0... which causes troubles when parent intends to set inner Boxes' margin via `'> * + *': 'marginLeft: 30px'` for instance.
Why do we need to set margin: 0? If it's inherited, we can set `m={0}` in case it's required (which is an edge case really).
Would love to know more about that. Thanks!
Hope that's not too far from the subject lol

lachlanjc:
@sergeylukin I believe it's just a reset, since elements like form fields & lists often have margin applied by the default browser stylesheet. Happy to accept better solutions to that problem, though!

PR #1812, opened 2021-06-10
fix(color-modes): combine color modes in nested theme providers

### LINKED ISSUE #1456: sx prop does not correctly apply theme in nested ThemeProviders with ColorMode
**Describe the bug**

When having an application that has a nested ThemeProviders with the same theme and color modes but different values, toggling the colorMode changes the colorMode and theme in both the top level and nested theme provider, but when utilizing that theme in the nested component with the sx prop, the wrong colour gets applied.

**To Reproduce**
Here is a simple codesandbox that reproduces the issue. https://codesandbox.io/s/exciting-banzai-hrjom?file=/src/index.js
**Expected behavior**

In the example provided, I would expect the inner components background value for light mode to be "darkcyan", and it is if you print out the theme object. But when its applied to the div using `sx={{ bg: "background" }}` it applies "red".

Screenshot 2021-01-21 at 11 48 01

Kapture 2021-01-21 at 11 54 02

hasparus:
Should close https://github.com/system-ui/theme-ui/issues/1456.

Updated repro: https://codesandbox.io/s/theme-ui-nested-color-modes-bug-qm0lt?file=/src/index.js.

@rezaabedian could you take a look at the [CodeSandbox](https://codesandbox.io/s/theme-ui-nested-color-modes-bug-qm0lt?file=/src/index.js) and see if this is what you expected in #1456?

### DIFF
--- a/packages/color-modes/src/index.tsx
+++ b/packages/color-modes/src/index.tsx
@@ -7,8 +7,9 @@ import React, {
 } from 'react'
 import {
   jsx,
+  ThemeUIContextValue,
   useThemeUI,
-  __ThemeUIInternalBaseThemeProvider,
+  __ThemeUIInternalBaseThemeProvider as ThemeUIInternalBaseThemeProvider,
 } from '@theme-ui/core'
 import {
   get,
@@ -34,22 +35,22 @@ const storage = {
   get: () => {
     try {
       return window.localStorage.getItem(STORAGE_KEY)
-    } catch (e) {
+    } catch (err) {
       console.warn(
         'localStorage is disabled and color mode might not work as expected.',
         'Please check your Site Settings.',
-        e
+        err
       )
     }
   },
   set: (value: string) => {
     try {
       window.localStorage.setItem(STORAGE_KEY, value)
-    } catch (e) {
+    } catch (err) {
       console.warn(
         'localStorage is disabled and color mode might not work as expected.',
         'Please check your Site Settings.',
-        e
+        err
       )
     }
   },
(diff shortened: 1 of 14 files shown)


hasparus:
Hey @fcisio, you're probably the best person to review these changes.

packages/color-modes/src/index.tsx
```diff
@@ -242,29 +254,60 @@ export const ColorModeProvider: React.FC = ({ children }) => {
...
+  const newTheme = useThemeWithAppliedColorMode({
+    outerTheme: outerCtx.theme,
+    colorMode: outerCtx.colorMode,
+  })
+
+  return (
+    <ThemeUIInternalBaseThemeProvider
+      context={{ ...outerCtx, theme: newTheme }}>
+      {/* Changed CSS Variables will cascade from the wrapping div */}
+      {jsx('div', {
```
fcisio, line 284: We could make this a `section` tag.
Not a big deal if it's a div, but would probably be more syntactically right.

fcisio:
I tested in a real project and It looks good to me!
I was worried that it might not work with `useCustomProperties: false`, but works perfectly.

Might be a good idea to supplement the docs about nested providers. Add an example like what's in the sandbox.

packages/color-modes/src/index.tsx
hasparus: I think I'd use a div here, because I truly only need it to apply styles, so want it to be as "semantically transparent" as possible.

I'm a bit worried that nested providers will now affect user's DOM and make layout less obvious. We could add "as" prop, accept styles from props, and expose NestedColorThemeProvider in public API, but I'd rather avoid it until people ask for it a lot.

hasparus:
:rocket: PR was released in [`v0.10.1-develop.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.10.1-develop.0) :rocket:

PR #1839, opened 2021-06-20
Create deno.yml

devkadirselcuk:


lachlanjc:
Hey! Can you provide some context on what this change is for & why we should make it?

hasparus:
@lachlanjc I'd guess we need to test if Theme UI can SSR in Deno for [Aleph.js](https://alephjs.org/), but some PR description would indeed be nice.

PR #1840, opened 2021-06-22
(color-modes) Add support for forEach on classList for IE11

thiniulian:
I've been working in a project using next.js and theme-ui and we have to support IE11. 
During debugging I could see that the classList.forEach was throwing an error due to IE11 not havving support for forEach on the NodeList data type, although modern browsers and the specs do.

This is the simplest form of adjusting this without adding polyfills.

### DIFF
--- a/packages/color-modes/src/index.tsx
+++ b/packages/color-modes/src/index.tsx
@@ -70,7 +70,7 @@ const getPreferredColorScheme = (): 'dark' | 'light' | null => {
 const getModeFromClass = (): string | undefined => {
   let mode: string | undefined
   if (typeof document !== 'undefined') {
-    document.documentElement.classList.forEach((className) => {
+    Array.from(document.documentElement.classList).forEach((className: string) => {
       if (className.startsWith('theme-ui-')) {
         mode = className.replace('theme-ui-', '')
       }


hasparus:
Hey @thiniulian, this might not be the only change needed to make it work in IE. I'd transpile Theme UI (and possibly other dependencies) and polyfill if I were you.

Here's our Babel config:

```
    [
      '@babel/preset-env',
      {
        bugfixes: true,
        loose: true,
        modules: false,
        targets: '> 0.25%, not dead, not ie 11',
      },
    ],
```

I'm skeptical towards merging this, because I we don't test for IE 11 in any way, and since we explicitly don't transpile for it, I'm sure that even if this is the only change required to work today, it's going to eventually break.

hasparus (changes requested):
Do you think we should document the fact that support for IE 11 is on user's side better? I think we only mentioned it in changelogs https://theme-ui.com/migrating/#v09.

thiniulian:
> Hey @thiniulian, this might not be the only change needed to make it work in IE. I'd transpile Theme UI (and possibly other dependencies) and polyfill if I were you.
> 
> Here's our Babel config:
> 
> ```
>     [
>       '@babel/preset-env',
>       {
>         bugfixes: true,
>         loose: true,
>         modules: false,
>         targets: '> 0.25%, not dead, not ie 11',
>       },
>     ],
> ```
> 
> I'm skeptical towards merging this, because I we don't test for IE 11 in any way, and since we explicitly don't transpile for it, I'm sure that even if this is the only change required to work today, it's going to eventually break.

Haven't tried the transpile method yet. We are very cautious on the bundle size and we try to keep to a minimum on polyfills and so on. I'll give it a spin now based on the change log instructions.

As for the documentation on not supporting IE11, I think bringing more visibility on it, and the things that should be done, would be more helpful for people like me that should (still) support IE11.

Thanks for the prompt answer!

hasparus:
Okay, transpilation instructions will be bundler-dependent so it's not trivial to add them to docs, but let's try to talk about it here so we can have some notes for the future docs page.

The gist is that we need to include deps shipping "modern" JS in our transpilation machinery.

With regards to polyfills, I'd use this one for classList.forEach:

```js
  if ('NodeList' in window && !NodeList.prototype.forEach) {
    NodeList.prototype.forEach = function (callback, thisArg) {
      thisArg = thisArg || window;
      for (var i = 0; i < this.length; i++) {
        callback.call(thisArg, this[i], i, this);
      }
    };
  }
```

Including stuff from Create React App polyfills might be also be needed: https://github.com/facebook/create-react-app/blob/master/packages/react-app-polyfill/ie11.js.

This blog post is pretty good: https://developer.epages.com/blog/coding/how-to-transpile-node-modules-with-babel-and-webpack-in-a-monorepo/

thiniulian:
> Do you think we should document the fact that support for IE 11 is on user's side better? I think we only mentioned it in changelogs https://theme-ui.com/migrating/#v09.

This  next-transpile-modules seems to do the trick for us.

PR #1856, opened 2021-07-04
Define correct types in most presets.

tornewuff:
Exported values in presets should have the correct types so they can be
used properly when imported. Define types for all the relatively-easy
cases, and fix various outdated property references.


lachlanjc (approved):
Thank you! I'm cool with this if the master of types @hasparus is

packages/preset-system/package.json
```diff
@@ -12,5 +12,8 @@
   "publishConfig": {
     "access": "public"
   },
+  "dependencies": {
```
hasparus: I’m not sure about adding dependencies here. Shouldn’t it be a peer dependency instead? We expect people to have core packages installed already.

packages/preset-system/src/index.ts
```diff
@@ -67,10 +68,10 @@ export const system = {
     body: 1.5,
     heading: 1.25,
   },
-  textStyles: {
+  text: {
```
hasparus: "textStyles" here was a bug, wasn’t it?

lachlanjc: I think so 😬

packages/preset-system/package.json
lachlanjc: Yep. Let's do that. I guess there's no way we can make it only download for TS users?

hasparus:
Okay, one concern I have is that this is breaking for people who read nested properties from exported objects, as `: Theme` obscures exact types. This is actually the main reason we didn’t merge a similar PR a while ago.

I'll try to add some test cases and send you some proposed changes this week.

Do you have any examples where presets cannot be used when imported currently?

packages/preset-system/src/index.ts
tornewuff: I think it was left over from before @theme-ui/components existed; the theme specification on styled-system suggests textStyles as one possible style variant key you might use. I updated this (and some other similar cases) to use `text` as that matches the current usage in components.

packages/preset-system/package.json
tornewuff: Yeah a peerDependency is probably more correct; we want the same version that the parent project is using (and if that causes a type error then presenting the error to the user is the right thing to do).

tornewuff:
Right now it's a type error to use most (maybe all?) of the preset themes directly:
```typescript
import { ThemeProvider } from '@theme-ui'
import { base } from '@theme-ui/presets'
const App = () => <ThemeProvider theme={base}/>
```
results in non-assignable type errors for various nested parts:
```
Type '{ space: number[]; fonts: { body: string; heading: string; monospace: string; }; fontSizes: number[]; fontWeights: { body: number; heading: number; bold: number; }; lineHeights: { body: number; heading: number; }; colors: { ...; }; styles: { ...; }; }' is not assignable to type 'Theme | ((outerTheme: Theme) => Theme)'.
  Type '{ space: number[]; fonts: { body: string; heading: string; monospace: string; }; fontSizes: number[]; fontWeights: { body: number; heading: number; bold: number; }; lineHeights: { body: number; heading: number; }; colors: { ...; }; styles: { ...; }; }' is not assignable to type 'Theme'.
    The types of 'styles.th' are incompatible between these types.
      Type '{ textAlign: string; borderBottomStyle: string; }' is not assignable to type 'ThemeUIStyleObject | undefined'.
        Type '{ textAlign: string; borderBottomStyle: string; }' is not assignable to type 'ThemeUICSSObject'.
          Types of property 'textAlign' are incompatible.
            Type 'string' is not assignable to type 'StylePropertyValue<TextAlign | undefined>'.
```

Now that you mention using nested parts and exact types though, it may be preferable here to just define them `as const` if that is sufficient for all the cases. I'll try it out.

tornewuff:
OK - making the whole thing `as const` breaks the assignability of a bunch of things to Scale, but making the styles dict `as const` fixes the issues for almost all the presets.

Added a new version here that does not require importing the type into the presets, but just has it as a devDependency for test purposes, with trivial 'tests' that verify that the exported preset is assignable to `Theme` when typechecked (but don't actually do anything when run).

How does this look?

This still doesn't fix preset-funk or preset-tailwind. The spread operator usage to compose objects there seems to defeat the type inference somehow; may think about this more later.

hasparus:
@tornewuff I'm gonna send a PR pointing at your branch. I added a few [_Constrained Identity Functions_](https://kentcdodds.com/blog/how-to-write-a-constrained-identity-function-in-typescript) to `@theme-ui/css/utils` and used them to get the best of both worlds in defining presets (we get typechecking and autocomplete as with `: Theme` annotation, and we preserve exact type as with `as const` assertion).

hasparus:
Okay, it seems I can't target your fork in the PR form, @tornewuff, so here's the link with my proposed changes:

https://github.com/system-ui/theme-ui/compare/preset-types-with-cif?expand=1

- As described above, I added a few identity functions and used them to define presets.
- I moved tests into `/presets` package.
- I got rid of that `undefined` from Tailwind preset color scales. This preset is on its way to deprecation / removal regardless, and sparse arrays didn't align with tests in many places in the codebase.

PR #1981, opened 2021-10-24
Use media query value when useColorSchemeMediaQuery is set to 'system'

### LINKED ISSUE #787: Unexpected `prefers-color-scheme` behaviour

draekien:
This PR is based on the changes proposed in https://github.com/system-ui/theme-ui/pull/981 and fixes https://github.com/system-ui/theme-ui/issues/787

This change adds a new option for `useColorSchemeMediaQuery` called `'system'` which will
listen for changes to the `prefers-color-scheme` media query and update the color mode
accordingly.

This change also updates the documentation to include a short description for the `'system'` option.

I decided to go with `'system'` over `'live'` as I've mostly encountered websites referring to the
user's preferred color scheme as the 'system' setting. Let me know if you would prefer `'live'` instead.

Testing: `'system'`
color-scheme-change

Testing: `'initial'` | `true`
color-scheme-initial


lachlanjc:
Wow, phenomenal work!! Haven't tested it out but looks great. Super appreciate the contribution.

draekien:
> Wow, phenomenal work!! Haven't tested it out but looks great. Super appreciate the contribution.

Thanks! I'm loving the library so wanted to see if I could contribute. Keep up the great work!!

packages/css/src/options.ts
```diff
@@ -25,7 +25,7 @@ export interface ThemeUIConfig {
   /**
    * Initializes the color mode based on the prefers-color-scheme media query
    */
-  useColorSchemeMediaQuery?: boolean
+  useColorSchemeMediaQuery?: 'system' | 'initial' | boolean
```
hasparus:
```suggestion
  useColorSchemeMediaQuery?: 'system' | 'initial' | true /* same as 'initial' for compat */ | false
```

We can later update the docs to refer to `"initial"` instead of `true`.

hasparus:
Great PR @draekien! Thank you for the contribution!

lachlanjc (approved):
Tried this locally & works exactly as I'd expect!

hasparus:
:rocket: PR was released in [`v0.12.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.12.0) :rocket:

PR #1983, opened 2021-10-27
feat(components): Add gap prop to Flex

### LINKED ISSUE #556: Add gap prop to Flex component

lachlanjc:
Closes #556!

### DIFF
--- a/packages/components/index.d.ts
+++ b/packages/components/index.d.ts
@@ -32,13 +32,17 @@ export interface BoxProps
  */
 export const Box: StyledComponent<BoxOwnProps, BoxProps>
 
-export type FlexOwnProps = BoxOwnProps
-export type FlexProps = BoxProps
+export interface FlexProps extends BoxProps {
+  /**
+   * Space between child elements
+   */
+  gap?: ResponsiveStyleValue<string | number>
+}
 /**
  * Use the Flex component to create flexbox layouts.
  * @see https://theme-ui.com/components/flex
  */
-export const Flex: StyledComponent<FlexOwnProps, FlexProps>
+export const Flex: ForwardRef<HTMLDivElement, FlexProps>
 
 export interface GridProps extends BoxProps {
   /**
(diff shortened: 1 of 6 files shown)


lachlanjc:
@hasparus It looks like the test for changing the tag on the `Flex` component is failing. Since I made it no longer directly a styled component with Emotion, but instead a React component like Grid, I would expect the `as` prop to work here. Is this a problem with our test not picking up on the `as` prop, or a problem with my code?
image

lachlanjc:
There’s also a semantic question here—should `gap` be a prop if flexDirection, alignItems, etc aren’t? Maybe this isn’t adding anything useful & is too confusing? We already support gap inside sx, after all

lachlanjc:
I think I've talked myself into this not being a good idea! Closing the original issue.

PR #2045, opened 2021-12-29
fix: presets-demo and select-field styles

appsparkler:
Fixes:
- presets demo wasn't working
- minor style issue when you select the `bootstrap` in the field for selecting presets.

### DIFF
--- a/packages/docs/src/components/presets-demo.js
+++ b/packages/docs/src/components/presets-demo.js
@@ -1,6 +1,5 @@
 /** @jsx jsx */
-import { jsx, Themed, components, Select } from 'theme-ui'
-import { ThemeContext } from '@emotion/react'
+import { jsx, Themed, components, Select, ThemeProvider } from 'theme-ui'
 import { MDXProvider } from '@mdx-js/react'
 import { useState } from 'react'
 import { Helmet } from 'react-helmet'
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/presets-demo.js
```diff
@@ -46,7 +46,7 @@ export default function PresetsDemo() {
           <span>Preset:</span>
           <Select
             id="theme"
-            sx={{ display: 'inline-flex' }}
+            sx={{ display: 'inline-flex', paddingRight: 24 }}
```
lachlanjc: Good catch here. Let's increase it further though? On my iPhone:

image

packages/docs/src/components/presets-demo.js
```diff
@@ -16,7 +15,8 @@ import Lorem from './lorem.mdx'
 
 export default function PresetsDemo() {
   const [theme, setTheme] = useState('base')
-  const preset = presets[theme]
+  const preset = presets[theme]   
+
```
lachlanjc, line 19:
```suggestion
```

lachlanjc (approved):
Thank you for the contribution! Looks great.

packages/docs/src/components/presets-demo.js
hasparus:
```suggestion
            sx={{ display: 'inline-flex', paddingRight: 32 }}
```

hasparus:
Does anybody have an idea why Cypress fails here?

packages/docs/src/components/presets-demo.js
appsparkler: @lachlanjc , I've added the style to the core `Select` component : [chore: update style for core Select component](https://github.com/system-ui/theme-ui/pull/2058)

PR #2058, opened 2021-12-30
chore: update style for core Select component

appsparkler:
Fixing the minor style issue in the core component so that we don't need to fix individual ones.

### Release notes

`Select` component default `paddingRight` style has increased to `spaces[4]` (32px by default), to avoid text flowing behind the chevron icon.


flo-sch (commented):
LGTM

There is just a failing test in the CI, could you please update the snapshot?

`packages/components/test/index.js`

hasparus:
Okay, I don't have anything against this change, but are we really sure we want to add this `32px` to all user projects?

I wouldn't classify this as breaking, but it might be annoying for some users.
It would need a release note, and I think we should bundle it with some more features.

appsparkler:
> There is just a failing test in the CI, could you please update the snapshot?

@flo-sch , DONE : https://github.com/system-ui/theme-ui/pull/2058/commits/f596aaa9ddd2c9741e054160a522ab54e9de5f16

flo-sch:
> I wouldn't classify this as breaking, but it might be annoying for some users.
> 
> It would need a release note, and I think we should bundle it with some more features.

Fair enough, it might indeed, especially when using another SVG arrow.

(The padding can still be customised from the theme, or?)

flo-sch (approved):

lachlanjc:
Yeah, I do hesitate a bit with a "magic number", but at the same time, this has basically been a bug with our component if you're not making it full-width through some other means. E.g. with the theme on <https://theme.hackclub.com>, if I wasn't maximizing the width of the `Select` via CSS Grid columns, this is the default behavior:
image
We're already setting `padding: 2`, so setting `paddingRight: 4` seems pretty reasonable to me. (Let's use `4` instead of `32` though, so it's customizable via the spacing scale.)

Appreciate this contribution!

appsparkler:
> Let's use `4` instead of `32` though, so it's customizable via the spacing scale

Hi @lachlanjc; DONE : 14c58be

lachlanjc (approved):

hasparus:
Hey @lachlanjc @appsparkler, I've added release notes section to the first post. Feel free to tweak it to your liking — it will land in the changelog and GitHub releases.

@lachlanjc I've also labeled this with `minor` as it may be surprising for some users, so how about we bundle it with removing Emotion styled (I promise I'll finish that PR soon 🤦‍♂️) and release as 0.14.0?

lachlanjc (approved):
@hasparus This is ready to merge!

hasparus:
:rocket: PR was released in [`v0.15.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.15.0) :rocket:

PR #2087, opened 2022-01-17
docs: Document default theme values

### LINKED ISSUE #2026: Document theme properties' default values
**Describe the bug**
I noticed that if I omit some keys from the theme object I pass to the `ThemeProvider`, such as `breakpoints` and `fontSizes`, some default values will be used, and I couldn’t find that documented anywhere so I don’t think it’s by design. The default values for these two keys are the same as in the [base preset](https://theme-ui.com/theming#example-theme), but I can confirm that it’s not the whole base preset, as `fonts` and `fontWeights` behave as expected, for example. Could it have something to do with the fact that `breakpoints` and `fontSizes` are arrays? What’s especially strange is that I can’t find the default values in the theme returned by the `useTheme` hook; they only pop up when I try to use them (e.g. `fontSize: 1`, or any value set to an array will flip at `40em`, then `52em`, etc.).

**To Reproduce**

1. Leave out the `breakpoints` and `fontSizes` keys from your theme.
2. Render `<div sx={{ fontSize: 1 }} />`
3. Notice the rendered CSS is `font-size: 14px`.

**Expected behavior**
I would’ve expected the rendered CSS to be `font-size: 1px`, which is the case when the theme has a `fontSizes` key set to an empty array.

lachlanjc:
Closes #2026 (see reasoning in that thread), and adds a few more links/info to the Theming doc.

### DIFF
--- a/packages/docs/src/pages/theming.mdx
+++ b/packages/docs/src/pages/theming.mdx
@@ -20,7 +20,7 @@ In order to ensure color palettes are as interoperable as possible, the followin
 
 Beyond these base colors, any additional values can be added to a theme to augment the base color palette.
 
-### Color Modes
+### Color modes
 
 Multiple color modes, i.e. _dark mode_, can be handled with a nested `modes` object that matches the shape of the default colors.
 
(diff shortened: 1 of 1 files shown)


hasparus:
:rocket: PR was released in [`v0.14.0`](https://github.com/system-ui/theme-ui/releases/tag/v0.14.0) :rocket:

PR #2122, opened 2022-02-08
docs/style: Update document for sx-prop. (#2118)

PeterlitsZo:
It is hard to find that how sx prop works and how to let sx-prop works for
user-defind components.

We update the document to make it easy to get.


PeterlitsZo:
No one cares. Closed.

Issue #2123, opened 2022-02-08
docs(color-modes): Add examples of array or object color scales


PeterlitsZo:
After read source code. We find that it sounds like that color can be plain object or array and so on. But we do not find it in https://theme-ui.com/color-modes.

Does it need update? If it need, I can help you with PR.

lachlanjc:
Good call! Yes, would love a PR.

PeterlitsZo:
OK, I will try. Wait for soon.

PeterlitsZo:
Done, I am going to close this.

PR #2128, opened 2022-02-09
docs(color-modes): Add document about color value can be obj/arr/hex.

PeterlitsZo:
For issue  #2123


hasparus:
I think that part of the docs was more of _"How does Theme UI color modes logic work?"_ instead of _"How to define your color modes?_". We might need to rewrite this whole page when I think about it.

Issue #2124, opened 2022-02-08
How about remove the Label warpper for Switch?


PeterlitsZo:
If we want to to let the label be before of the switch, we need:
``` jsx
<Flex
  sx={{
    justifyContent: 'space-between',
    alignItems: 'center',
    py: 4,
  }}>
  <Label htmlFor="enable-email-alerts" sx={{ flex: 1 }}>
    Enable email alerts?
  </Label>
  <Box>
    <Switch id="enable-email-alerts" />
  </Box>
</Flex>
```

But I find that Switch is already a Label element. Why? I think this will be better:
``` jsx
<Label sx={{
    display: 'flex',
    justifyContent: 'space-between',
    alignItems: 'center',
    py: 4,
  }}>
  <span>Enable email alerts?</span>
  <Switch /> {/* A React.Fragment but not label element */}
</Label>
```

PeterlitsZo:
It will break something... so we can create a new component `PureSwitch`. How about it?

PeterlitsZo:
I will try in my project.

PeterlitsZo:
By the way, I find that the `Box` set `margin: 0`, which let the parent's `& > * + *`'s `margin-left: 5px` rule not works.

Need we set the `Box`'s `margin: 0`?

PeterlitsZo:
Here is my work, it works well for me:

<details>
  <summary>My code!</summary>

  ```tsx
  import React from 'react';
  import { Box, Label } from 'theme-ui';
  import type { CSSProperties as CSSProp } from 'theme-ui';
  
  const GUTTER = 2;
  const SIZE = 18;
  
  interface SwitchProps {
    gutter?: CSSProp['height'] & CSSProp['width'] & CSSProp['borderRadius'];
    size?: CSSProp['height'] & CSSProp['width'] & CSSProp['borderRadius'];
    className?: string;
    label?: string;
    variant?: string;
    [key: string]: any;
  }
  
  export default React.forwardRef(
    (props: SwitchProps, ref: React.ForwardedRef<HTMLInputElement>) => {
      let {
        gutter = GUTTER,
        size = SIZE,
        className,
        label,
        variant = 'switch',
        ...otherProps
      } = props;
      if (typeof gutter === 'number') gutter = gutter + 'px';
      if (typeof size === 'number') size = size + 'px';
  
      // Hidden checkbox, the really element.
      const Checkbox = (
        <Box
          ref={ref}
          as="input"
          // @ts-ignore
[...]

PR #2137, opened 2022-02-18
feat(components): Add `labelPosition` prop on `Switch`

PeterlitsZo:
Switch has a prop `label` to set its label. And its position is
defaultly at the end. To put the label at the start of `Switch` is
hard.

So I add `labelPosition`, using `gap: 2` instead of the `mr: 2` to make
sure that those children's space is right.

By the way, I move those inner components outside, to make those code
more clear than before. I hope that works.

### DIFF
--- a/packages/color-modes/package.json
+++ b/packages/color-modes/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@theme-ui/color-modes",
-  "version": "0.16.0-develop.4",
+  "version": "0.16.1-develop.0",
   "main": "dist/theme-ui-color-modes.cjs.js",
   "module": "dist/theme-ui-color-modes.esm.js",
   "types": "dist/theme-ui-color-modes.cjs.d.ts",
--- a/packages/color/package.json
+++ b/packages/color/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@theme-ui/color",
-  "version": "0.16.0-develop.4",
+  "version": "0.16.1-develop.0",
   "source": "src/index.ts",
   "main": "dist/theme-ui-color.cjs.js",
   "module": "dist/theme-ui-color.esm.js",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@theme-ui/components",
-  "version": "0.16.0-develop.4",
+  "version": "0.16.1-develop.0",
   "main": "dist/theme-ui-components.cjs.js",
   "module": "dist/theme-ui-components.esm.js",
   "types": "dist/theme-ui-components.cjs.d.ts",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@theme-ui/core",
-  "version": "0.16.0-develop.4",
+  "version": "0.16.1-develop.0",
   "source": "src/index.ts",
   "main": "dist/theme-ui-core.cjs.js",
   "module": "dist/theme-ui-core.esm.js",
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@theme-ui/css",
-  "version": "0.16.0-develop.4",
+  "version": "0.16.1-develop.0",
   "source": "src/index.ts",
   "main": "dist/theme-ui-css.cjs.js",
   "module": "dist/theme-ui-css.esm.js",
--- a/packages/custom-properties/package.json
+++ b/packages/custom-properties/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@theme-ui/custom-properties",
   "description": "Generate CSS custom properties for use with Theme UI",
-  "version": "0.16.0-develop.4",
+  "version": "0.16.1-develop.0",
   "source": "src/index.ts",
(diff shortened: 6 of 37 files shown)


PeterlitsZo:
For issue #2124 

Weekend, right?

PeterlitsZo:
Well, I find that you like to use package name rather than the component name...

I will fix it with force push. (I am not sure that it is a good idea...)

I will go to sleep. See you tomorrow.

packages/components/src/Switch.js
```diff
@@ -6,11 +6,12 @@ const GUTTER = 2
...
 
 export const Switch = React.forwardRef(function Switch(
-  { className, label, sx, variant = 'switch', ...props },
+  { className, label, labelPosition = 'end', sx, variant = 'switch', ...props },
   ref
 ) {
-  return (
-    <Label sx={{ cursor: 'pointer' }}>
+  // Inner really checkbox (but hidden).
+  const ReallyHiddenCheckbox = ({ ref, label, ...props }) => {
```
hasparus: Creating new components within components breaks reconciliation. Could you move this to the module scope or even inline it again?

In this particular case, there won't be any visible bug, but it's still a good idea to just avoid nesting components like this.

I've made a small repro for you to illustrate the problem: https://codesandbox.io/s/lucid-rain-oeq4ph?file=/src/App.tsx

Notice that when you change the text in the buggy input it resets — it mounts again, because from React's perspective this is a new component.


Further reading: https://reactjs.org/docs/reconciliation.html

hasparus (changes requested):
https://github.com/system-ui/theme-ui/pull/2137/files#r810646380

PeterlitsZo:
OK! I will work on it tonight!

PeterlitsZo:
- tonight `X`
- after 11 days `V`

By the way, I find that if running `yarn format`, it changes a lot of files.

I think we need to commit with those files which is formatted.

packages/components/src/Switch.js
PeterlitsZo: Done!

packages/docs/src/pages/components/switch.mdx
```diff
@@ -42,8 +44,25 @@ To show different style based on the input state, you can use
...
+```jsx live=true
+<Switch label="Do Not Click Me!" labelPosition='end' />
+```
+
 ## External label
 
+<Note>
+Try to use `labelPosition` prop rather than another Label layer!
+</Note>
+
```
lachlanjc:
```suggestion
```
Don't think this note is needed since it's directly after the `labelPosition` section, but don't feel strongly. If you want to keep it, let's drop the exclamation point, to make it a more chill suggestion & not come across as shouty.

packages/docs/src/pages/components/switch.mdx
```diff
@@ -42,8 +44,25 @@ To show different style based on the input state, you can use
 />
 ```
 
+## `labelPosition`
+
+May you want to change the label's position, so there are props
+`labelPosition`. Its value can be `'start'`, `'end'`(default).
```
lachlanjc:
```suggestion
By default, the switch comes first, with the label to the right of it. If you'd like to change the ordering, set the `labelPosition` prop to `"start"`, or the default value, `"end"`.
```

packages/docs/src/pages/components/switch.mdx
```diff
@@ -42,8 +44,25 @@ To show different style based on the input state, you can use
...
+
+```jsx live=true
+<Switch label="Click Me!" labelPosition='start' />
```
lachlanjc:
```suggestion
<Switch label="Use 24-Hour Time" labelPosition="start" />
```
Use consistent quote style & realistic label example

packages/docs/src/pages/components/switch.mdx
```diff
@@ -42,8 +44,25 @@ To show different style based on the input state, you can use
...
+```
+
+```jsx live=true
+<Switch label="Do Not Click Me!" labelPosition='end' />
```
lachlanjc:
```suggestion
<Switch label="Auto-Correction" labelPosition="end" />
```

lachlanjc (approved):
Thanks for contributing this, sorry for the long wait! Suggested some docs improvements but the code functionality looks good.

packages/docs/src/pages/components/switch.mdx
PeterlitsZo: OK. In my opinion, removing the exclametion point will be better.

PeterlitsZo:
If it is OK, please tell me to rebase the git history to get more clear history.

chrislombaard:
Hi @PeterlitsZo @hasparus @lachlanjc , is there an update on this PR? 

We're currently experiencing this issue in our project with the Switch/ Label stye props being only targetable for the `Switch` component and not the `Label`.  I.e we want to either disable, or move the label to the left side of the switch. Is there another work-around for this in the meantime?

PeterlitsZo:
@chrislombaard Unbelievable that this PR is still not merged. I will deal with it in this weekend (maybe this weekend :) ).

PeterlitsZo:
Wait one more week...

ghost:
> Wait one more week...
 🙏

PeterlitsZo:
-> Move to new PR: #2452

PR #2176, opened 2022-03-27
small refactor in ThemeProvider

TKDev7:
Small refactor in ThemeProvider


lachlanjc (approved):

hasparus:
I’m sorry, but this isn’t a refactor.
`useRootStyles === false` returns `false` when `useRootStyles` is `undefined` or `null`. We treat it as `true` unless it’s explicitly set to `false`. That optional chaining is also not equivalent to the old code.

PR #2211, opened 2022-04-26
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,8 @@ jobs:
   # but secrets and env vars cannot be read in `job.if`, so we check if
   # GH_TOKEN is present before attempting to release.
   release_check:
+    permissions:
+      contents: none
     runs-on: ubuntu-latest
     needs: [build]
     steps:


hasparus (changes requested):
I think you forgot to add the same code to the `release` job, but I wonder if this can even work. Don't we need content permissions to push a commit and tag? 

Take a look at these commits 
image

It's my avatar there, but it's our automation that creates them.

I'd gladly talk about how we can make our workflow more secure, but I'm not sure if this PR improves the state of it.

naveensrinivasan:
> I think you forgot to add the same code to the `release` job, but I wonder if this can even work. Don't we need content permissions to push a commit and tag?
> 
> Take a look at these commits image
> 
> It's my avatar there, but it's our automation that creates them.
> 
> I'd gladly talk about how we can make our workflow more secure, but I'm not sure if this PR improves the state of it.

You are right there are additional permissions required. This would not be enough.

lachlanjc:
Closing for now since this has gone stale, but happy to take a look at future work! Thank you!

PR #2212, opened 2022-04-27
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -96,3 +96,8 @@ updates:
       - dependency-name: 'cypress'
         versions:
           - '> 9.0.0, < 10'
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: monthly
+    open-pull-requests-limit: 10


hasparus:
Hey @naveensrinivasan.

Thank you for the interest in Theme UI.
I'm sorry to say that, but we're unfortunately not going to merge this.
As a rule of thumb, we are not merging first-time contributors Pull Requests affecting CI.

Our GitHub Actions setup is internal. It doesn't affect the users, but it's a possible vector for attack that can affect them very much.

PR #2326, opened 2022-09-22
components: Remove style props

### LINKED ISSUE #706: [Components/Select] Expose access to container props

I'm using the `Select` components and use them inline. When the component width is small, the position of the arrow down is not appropriate anymore. I'd love to have access to the container div to make changes to the overall layout of the `Select` component.

Have a `containerSx` prop:

```jsx
<Select containerSx={{ justifyContent: "space-betweent", etc }} />
```

Currently, to solve this issue I have to create and maintain my own `InlineSelect` component, which is far from ideal.

---

PS: Since we're in the topic of `Select`, I just wonder what if users of `theme-ui` wants to use a different arrow down icon. It seems like there is not a way to change that. Just wonder what you think about it and if you're open to adding some other prop to expose that functionality.

lachlanjc:
We'd mentioned this in #2275 & #832 as part of simplifying our API, as it can lead to confusing CSS ordering & overrides: if I have the same property in a style prop & sx, which wins out? It's especially unclear from a component that's wrapping a Theme UI component. This is a big, breaking change, so it should definitely not be in our MDX release, in case folks don't want to upgrade immediately. Providing a codemod would be great too; I haven't worked on that yet.

## Release notes

- **BREAKING:** Theme UI Components no longer accept style props (e.g. `<Box bg="primary" mt={3}`), in favor of only supporting styling through the `sx` prop.
  - The `Field` component now accepts a second `containerSx` prop, intended to replace whitespace style props (e.g. to convert `<Field my={3} sx={{ … }}` to `<Field containerSx={{ my: 3 }} sx={{ … }}`). `Checkbox`, `Select`, and `Switch` have this new prop as well.


lachlanjc:
@hasparus thoughts on this?

devatrox:
Looks good to me. I very much appreciate the containerSx prop!

packages/docs/src/pages/components/index.mdx
```diff
@@ -85,29 +83,33 @@ All Theme UI components include the following props.
...
-These props work the same as properties within the `sx` prop, but allow for a more terse syntax for readability.
+Each Theme UI component includes shorthand props for applying margin, padding,
+and color styles. These props work the same as properties within the `sx` prop,
+but allow for a more terse syntax for readability.
 
 <Note>
 
-If you’ve used [Styled System][] or [Rebass][] before, these props work the same way.
+If you’ve used [Styled System][] or [Rebass][] before, these props work the same
+way.
```
hasparus: @lachlanjc Should we remove this?

hasparus:
@lachlanjc I can write a simple codemod if you think it's good to merge.

lachlanjc:
That'd be amazing! Definitely my main hesitation.

PR #2341, opened 2022-10-05
fix(css): Prevent transforming @page properties like size

### LINKED ISSUE #2166: Size shorthand is overwriting an existing CSS property
I was mighty confused why my printing stylesheet wasn't working, until I finally discovered that the `size` property was being replaced with `width` and `height`.

[size is a property used for defining @page size](https://developer.mozilla.org/en-US/docs/Web/CSS/@page/size)

I might recommend `wh` as a shorthand instead, as it describes the format of the string to be passed, is shorter, and is generally more descriptive than `size`. On top of not conflicting with an existing property.

lachlanjc:
Closes #2166. This may not be ideal because it prevents transforms of _all_ properties inside `@page`, but the way our transform function is setup, it doesn't seem easy to fix otherwise. This means use of margin properties inside `@page` will no longer be transformed; in my experience you're only using those properties so you can define physical units like `in` & `cm` instead of pixels, so doesn't seem like a big deal to me.

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -368,7 +368,7 @@ export const css =
       const x = styles[key as keyof typeof styles]
       const val = typeof x === 'function' ? x(theme) : x
 
-      if (val && typeof val === 'object') {
+      if (val && typeof val === 'object' && key !== '@page') {
         if (hasDefault(val)) {
           result[key] = val[THEME_UI_DEFAULT_KEY]
           continue
--- a/packages/css/test/past-bugs.ts
+++ b/packages/css/test/past-bugs.ts
@@ -31,4 +31,15 @@ describe('theme scales, get and default object property (#1439)', () => {
 
     expect(actual).toStrictEqual({ zIndex: 1 })
   })
+
+  // https://github.com/system-ui/theme-ui/issues/2166
+  test('size property works inside at page', () => {
+    const actual = css({ size: '2rem', '@page': { size: 'A4' } })({})
+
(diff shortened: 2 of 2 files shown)


packages/css/src/index.ts
```diff
@@ -368,7 +368,7 @@ export const css =
       const x = styles[key as keyof typeof styles]
       const val = typeof x === 'function' ? x(theme) : x
 
-      if (val && typeof val === 'object') {
```
hasparus, line 375: I agree that this is the implementation for the solution, but I'm very conflicted about having a special case for `@page`.

It feels bad, doesn't it? Could we maybe... choose a different solution?
There's no perfect solution here, but I suppose I'd rather resign from the `size` shorthand and add `w` and `h` shorthands in its place.

An alternative we could consider — could we just not implement anything?
There's an escape hatch with `css` prop, Safari doesn't support `@page`. Seems less _correct_ than dropping the `size` alias, and we'd probably need an error thrown only in development, but it's something 🤷

lachlanjc: w/h shorthands will require changes by a significant percentage of Theme UI users & seems like a price they shouldn’t have to pay to me. The number of people using `@page` is likely tiny, & it’s very easy to update what code in the extremely rare case this change breaks it (imagine usages of that selector to find & replace vs “size” in a codebase). This isn’t very correct but it just doesn’t seem like a common issue to be sad about not getting your (screen-optimized) theme as automatically accessible in printing page setup for some browsers.

PR #2362, opened 2022-11-17
fix: Destructure Emotion version in import

### LINKED ISSUE #2361: Spread emotion/react version import
I'm using rollup to build my application and I'm getting some warning because rollup understands as a global module as can see in the follow image:

image

My suggestion is spread the version on import, something like:

```
import { version as emotionReactVersion } from '@emotion/react/package.json'

...

const __EMOTION_VERSION__ = emotionReactVersion  
```

**Additional context**
References:

https://github.com/system-ui/theme-ui/blob/f5ae7d69bb3f63dff996f6ddf4705009d3b4429a/packages/core/src/index.ts#L8

https://github.com/system-ui/theme-ui/blob/f5ae7d69bb3f63dff996f6ddf4705009d3b4429a/packages/core/src/index.ts#L38

lachlanjc:
Closes #2361

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -5,7 +5,7 @@ import {
 import { Theme } from '@theme-ui/css'
 import * as React from 'react'
 import deepmerge from 'deepmerge'
-import packageInfo from '@emotion/react/package.json'
+import { version as __EMOTION_VERSION__ } from '@emotion/react/package.json'
 import { parseProps } from './parseProps'
 
 import { ThemeUIJSX } from './jsx-namespace'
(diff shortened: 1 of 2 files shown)


hasparus:
Does this work? I was under impression that JSON must be default imported. 

We actually used to have this code with named import if I remember correctly

```ts
import { version as __EMOTION_VERSION__ } from '@emotion/react/package.json'
```

and we needed to change to default import.

lachlanjc:
I don't know! It builds & works on CI so… :brain:

hasparus:
Okay, I've found it.

- https://github.com/system-ui/theme-ui/issues/1097
- https://github.com/aws/aws-sdk-js-v3/issues/1504
- https://github.com/quasarframework/quasar/issues/7943
- https://webpack.js.org/migrate/5/#using-named-exports-from-json-modules

I didn't test it yet, but if nothing changed, merging this will break Theme UI in Webpack 5.

lachlanjc:
Thanks for the details there! Was surprised it was written this way if the obvious destructuring also has benefits, so that makes sense.

PR #2366, opened 2022-11-21
add @emotion/react to peerDependencies

### LINKED ISSUE #2198: Ensure all transitive peerDependencies are listed correctly and test Yarn 2/3 installs
We're currently causing peerDependency warnings in Yarn 2.
Whenever we depend on something, this dependencies peerDependencies must be listed either in our peerDependencies or in our dependencies.

### Discussed in https://github.com/system-ui/theme-ui/discussions/2196

<div type='discussions-op-text'>

<sup>Originally posted by **jw-miaem** April  7, 2022</sup>
Reason I ask (before raising) is because I'm unable to get rid of 'doesn't provide ...' warning with yarn berry package.json install. I looked at the peerDependencies for @theme-ui/css for example and its "@emotion/react": "^11" and I have "@emotion/react": "^11.9.0" installed which I thought would satisfy but could be wrong</div>

ckeeney:
adds
```json
  "peerDependencies": {
    "@emotion/react": "^11"
  },
```
to package.json.

Resolves #2198

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,9 @@
     "typecov": "^0.2.3",
     "typescript": "^4"
   },
+  "peerDependencies": {
+    "@emotion/react": "^11"
+  },
   "pnpm": {
     "overrides": {
       "@types/readable-stream": "2.3.11",


lachlanjc:
Hmm, when I `pnpm i` locally:
CleanShot 2022-11-21 at 23 40 53@2x

hasparus (changes requested):
Hey @ckeeney. Thank you for your interest in Theme UI.

You've added a peer dependency to workspace root package.json, which is a private package.
It contains the dependencies and scripts used to maintain the library.

If you're looking for one of `theme-ui`'s proper package.json files, you can find them in _packages/_ directory.

However, I must say sorry — I shouldn't have labeled #2198 as _good first issue_ — the work to do here is a bit more complex than modifying `peerDependencies` in a `package.json`. To close the issue, we'd need a TypeScript or Bash script, a GitHub action step that would verify we're installing without errors.

PR #2386, opened 2023-01-22
Fix tests for the Global component

hasparus:
I've added comments below ⬇️

### DIFF
--- a/examples/next/components/Header.tsx
+++ b/examples/next/components/Header.tsx
@@ -2,9 +2,11 @@ import { Button, useColorMode } from 'theme-ui'
 
 const Header = () => {
   const [colorMode, setColorMode] = useColorMode()
+
   return (
     <header>
       <Button
+        suppressHydrationWarning
         onClick={() => setColorMode(colorMode === 'light' ? 'dark' : 'light')}
       >
         Toggle {colorMode === 'light' ? 'Dark' : 'Light'}
--- a/packages/global/test/__snapshots__/index.tsx.snap
+++ b/packages/global/test/__snapshots__/index.tsx.snap
@@ -45,32 +45,25 @@ exports[`renders global styles 1`] = `
 <html>
   <head>
     <style
-      data-emotion="css"
-      data-s=""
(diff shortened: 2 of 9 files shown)


examples/next/components/Header.tsx
```diff
@@ -2,9 +2,11 @@ import { Button, useColorMode } from 'theme-ui'
 
 const Header = () => {
   const [colorMode, setColorMode] = useColorMode()
+
   return (
     <header>
       <Button
+        suppressHydrationWarning
```
hasparus, line 9: When I launched the Next example, I noticed the hydration warning popping up straight in your face. This component _will_ cause it, because it's using `useColorMode` without waiting until we're on the client, and we depend on the color scheme value in local storage. (We could render a skeleton for the button on the server instead, but that's not much better UX IMHO.)

packages/global/test/__snapshots__/index.tsx.snap
```diff
@@ -45,32 +45,25 @@ exports[`renders global styles 1`] = `
...
       
       @font-face{font-family:some-name;}
     </style>
     <style
-      data-emotion="css"
+      data-emotion="css-global"
       data-s=""
     >
       
-      .css-ck0s41 body{font-family:Georgia,serif;margin:0;}
```
hasparus, line 66: I've read Emotion Global's code to figure it out:

1. This class name shouldn't be there.
2. `data-emotion` should be `css-global`, not `css`.
3. Wait up, original Global uses `styles` instead of `css` prop. Eureka moment.
4. We have JSX pragma configured in our tests.
5. `sx` becomes `css`, which becomes `className` 

I've changed the name of the prop to `styles`, because __it will happen__ for the users. They're gonna use `<Global>` alongside other JSX, in files where JSX goes through `jsx` function from Theme UI.

Side effect of it is, we now have the same signature that Emotion Global has, just with different type of `styles`, so the user can just change the import if they need to access their design tokens, and even change it back if they no longer need the theme there.

PR #2389, opened 2023-02-03
fix(custom-properties): use default key name if pluralize returns empty

wahidrahim:
This fixes a bug in custom-properties where keys such as `"s"` are not generated correctly. This is due to the pluralize package not handling `"s"` in an expected way in this context of themes.

Essentially 
```ts
pluralize("s", 1) === ""
```

This will not be corrected on pluralize, see: https://github.com/plurals/pluralize/issues/36

This poses a problem for theme keys which is using "s" as a key to indicate "small", for example.

### Bug

Consider a theme where we have radius sizes defined as:

```ts
const theme: Theme = {
  // ...
  radii: {
    0: '0px',
    s: '2px',
    m: '4px',
    l: '8px',
    xl: '16px',
  },
}
```

This will create an incorrect CSS var, for the "s" key:

```css
--theme-ui-radius-0: 0px;
--theme-ui-radius-: 2px; /* Incorrect */
--theme-ui-radius-m: 4px;
--theme-ui-radius-l: 8px;
--theme-ui-radius-xl: 16px;
```

And you also get the invalid CSS custom property warning:

```
theme-ui] Theme key "2px" found will produce an invalid CSS custom property. Keys must only contain the following: A-Z, a-z, 0-9, hyphen, underscore.
```

### Solution

Since `pluralize` can at times return empty strings, in such cases it is better to just use the `key` itself. With this fix the output for the radius CSS custom properties are as expected:

```css
--theme-ui-radius-0: 0px;
--theme-ui-radius-s: 2px;
--theme-ui-radius-m: 4px;
--theme-ui-radius-l: 8px;
--theme-ui-radius-xl: 16px;
```


lachlanjc (approved):
Thank you for the thorough info & tests! Looks great

hasparus (approved):

hasparus:
:rocket: PR was released in [`v0.15.5-develop.2`](https://github.com/system-ui/theme-ui/releases/tag/v0.15.5-develop.2) :rocket:

hasparus:
:rocket: PR was released in [`v0.15.5`](https://github.com/system-ui/theme-ui/releases/tag/v0.15.5) :rocket:

PR #2357, opened 2022-11-04
Update preset-tailwind for Tailwind v3 (breaking)

lachlanjc:
Trying to make it easier to come over from Tailwind by having their latest color palettes etc in our preset!

# Version

Published prerelease version: `v0.15.4-develop.0`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - Update preset-tailwind for Tailwind v3 (breaking) [#2357](https://github.com/system-ui/theme-ui/pull/2357) ([@lachlanjc](https://github.com/lachlanjc))
  
  #### Authors: 1
  
  - Lachlan Campbell ([@lachlanjc](https://github.com/lachlanjc))
</details>

PR #2397, opened 2023-02-20
tailwind: Switch font sizes to an array

SiriusBits:
The `fontSizes` key takes an array of values (not an object). This fixes the heading size settings issue causing the text for headings to default to "6px" which can be seen on the [Demo page](https://theme-ui.com/demo) when selecting "Tailwind" from the drop down.

### DIFF
--- a/packages/preset-tailwind/src/index.ts
+++ b/packages/preset-tailwind/src/index.ts
@@ -413,21 +413,21 @@ export const fonts = {
   monospace: baseFonts.mono,
 }
 
-export const fontSizes = {
-  xs: '0.75rem',
-  sm: '0.875rem',
-  default: '1rem',
-  lg: '1.125rem',
-  xl: '1.25rem',
-  '2xl': '1.5rem',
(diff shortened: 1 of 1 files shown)


lachlanjc:
Good call, thank you!

hasparus:
This is breaking for everybody who used `fontSize: 'xs'`, right?

`fontSizes` and any scale can be both an array and an object.
I'd say the bug is just in the Demo page, and its assumption that the presets are uniform.

LekoArts:
This was changed to an object in https://github.com/system-ui/theme-ui/pull/2357 and Tailwind default theme is an object, too:

https://github.com/tailwindlabs/tailwindcss/blob/5b2c1800abc744984ec4e605903da9658118870e/stubs/config.full.js#L330-L344

So I think this PR should fix the demo page, not the theme.

lachlanjc:
Great context, thanks. I agree the `’xl’` style font sizes are a lot easier to remember/use & are an easier onboarding coming from Tailwind.

PR #2419, opened 2023-04-25
components: Added indeterminate checkbox

dev-cj:
Checkbox path is sourced from Material Design (same as the other ones)

https://user-images.githubusercontent.com/47112778/234390774-67f2fd7e-9458-4763-a38e-b7c46806a12e.mp4

### DIFF
--- a/packages/components/src/Checkbox.tsx
+++ b/packages/components/src/Checkbox.tsx
@@ -28,7 +28,9 @@ export interface CheckboxIconProps extends SVGProps {
 }
 
 const CheckboxIcon = (props: CheckboxIconProps) => {
-  const Component = props.isIndeterminate ? CheckboxIndeterminate : CheckboxChecked
+  const Component = props.isIndeterminate
+    ? CheckboxIndeterminate
+    : CheckboxChecked
   return (
     <React.Fragment>
       <Component
@@ -57,7 +59,8 @@ const CheckboxIcon = (props: CheckboxIconProps) => {
   )
 }
 
-export interface CheckboxProps extends Assign<React.ComponentPropsWithRef<'input'>, BoxOwnProps> {
+export interface CheckboxProps
+  extends Assign<React.ComponentPropsWithRef<'input'>, BoxOwnProps> {
   isIndeterminate?: boolean
 }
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Checkbox.tsx
```diff
@@ -11,41 +11,58 @@ const CheckboxChecked = (props: SVGProps) => (
...
-      })}
-    />
-    <CheckboxUnchecked
-      {...props}
-      {...__internalProps({
-        __css: {
-          display: 'block',
-          'input:checked ~ &': {
+export interface CheckboxIconProps extends SVGProps {
+  isIndeterminate?: boolean
```
hasparus: Thanks for the PR, @dev-cj!

What do you think about renaming this to `indeterminate` to match the attribute on `<input>`?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

packages/components/src/Checkbox.tsx
```diff
@@ -76,7 +100,7 @@ export const Checkbox: ForwardRef<HTMLInputElement, CheckboxProps> =
           }}
         />
         <Box
-          as={CheckboxIcon}
+          as={(SVGProps) => CheckboxIcon({ isIndeterminate, ...SVGProps })}
```
hasparus: This won't be a performance problem here, but we usually wouldn't want to create new components like this.

Could we use the [`:indeterminate`](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate) pseudoclass for with general sibling combinator `~` to control what's displayed, just like we use `:checked` right now?

hasparus (changes requested):
Huge thanks for the PR! It's looking good, and I really appreciate the video.

I left some small comments.

packages/components/src/Checkbox.tsx
dev-cj: Yeah, we should match it. 
Have made the changes.

packages/components/src/Checkbox.tsx
dev-cj: I tried that first, but the attribute `indeterminate` on input [cannot](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes) be set directly using HTML . And since none of the other components were performing side effects using hooks, I used this approach.

packages/components/src/Checkbox.tsx
hasparus: Thanks!

packages/components/src/Checkbox.tsx
hasparus: All right, so I suppose we have a few choices.
1. We can either leave it as it is,
2. Stop doing `<Box as={CheckboxIcon}>` and use `<CheckboxIcon indeterminate={indeterminate}` instead with the common props for all icons moved inside of `CheckboxIcon`
3. Maybe even inline `CheckboxIcon` and see if we can gain something by removing this layer of abstraction
4. or use a hook like you mentioned

dev-cj: I think we can go with the second option and use `<CheckboxIcon indeterminate={indeterminate}` as you said.

hasparus: Nice! I like this approach. Sorry for being so annoying here :)

dev-cj: No worries,
Have made the changes.

hasparus:
Nice one @dev-cj! Feel free to just update the snapshot with `jest -u`. Your version of the snapshot is better — the previous one had a few duplicate properties.

hasparus (approved):

hasparus:
Nice work! I'm merging to `develop` and will follow up in a stable release later today/this-weekend.

hasparus:
:rocket: PR was released in [`v0.15.8-develop.1`](https://github.com/system-ui/theme-ui/releases/tag/v0.15.8-develop.1) :rocket:

hasparus:
:rocket: PR was released in [`v0.15.8`](https://github.com/system-ui/theme-ui/releases/tag/v0.15.8) :rocket:

vkaragyaurov:
Hey, guys!

I was looking for the indeterminate state and found this PR. It says merged and released, but it doesn't seem present in the current version. I found no mentions of reverting it in the changelog, nor in the git history. After digging a bit more it looks like these changes were deleted by [Merge branch 'develop' into stable](https://github.com/system-ui/theme-ui/commit/c624b427b95d9d5aed23314e35414c873e1375ec). Any idea if this was intentional?

PR #2452, opened 2023-08-10
feat(components): Add labelPosition prop on Switch

PeterlitsZo:
`Switch` has a prop label to set its label. And its position is default at the end. Putting the label at the start of Switch is hard.

So I add `labelPosition`, using `gap: 2` instead of the `mr: 2` to make sure that those children's space is right.

Origin PR: #2137


PeterlitsZo:
Hello, I finally rebase it to the current commit. It has been delayed over one more year. If you think it is OK, please merge it please. I have only some limited time for this project.

@lachlanjc @hasparus

PR #2457, opened 2023-09-26
Filter scale tuples in Theme type inside .sx prop

### LINKED ISSUE #2456: Several TypeScript errors when upgrading to 0.16.1
**Describe the bug**
When upgrading to 0.16.1, we have introduced a bunch of TypeScript errors. An example of one such error, is when trying to access a string key on a `Scale` type property (e.g `theme.shadow`). An array index works as expected without any TS errors, but when trying to access a key (our shadow property in the theme is an object, not an array) we see a TS error.

This can be solved with type assertions, but it would be quite cumbersome to update all instances like this.

E.g
```typescript
<div
  sx={{
    boxShadow: (theme) => theme.shadows.firstLevel
  }}
/>
```

```
Property 'firstLevel' does not exist on type 'Scale<BoxShadow>'.
  Property 'firstLevel' does not exist on type 'BoxShadow[]'.ts(2339)
```

**To Reproduce**

1. Clone [this CRA sandbox](https://github.com/matt-koevort/react-theme-ui)
2. Open the project in a TypeScript enabled editor, or run `tsc --noEmit` (for example)
3. Note error when trying to access a string key on `theme.shadows` (as opposed to a numbered index)

**Expected behavior**
The type `Scale` indicates that it can be an array or an object. Out of the box it seems to assume that any property of type `Scale` is an array.

**Screenshots**

image

image

**Additional context**

hasparus:
Fixes https://github.com/system-ui/theme-ui/issues/2456.

Now the type of `theme` inside of `sx` prop doesn't contain array scales, so it should be easier to read from.

We now also allow injecting the theme, but with a smaller blast radius than in #1090.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,15 +13,15 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - uses: pnpm/action-setup@v2
         with:
-          version: 7
+          version: 8
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 18.x
+          node-version: 20.x
           cache: 'pnpm'
 
       - name: Install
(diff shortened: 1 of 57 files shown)


hasparus:
I'm converting it to a draft, because userspace type narrowing thru declaration merging creates a breaking change for all projects that contain functions `Theme => unknown`. Without `declare module "@theme-ui/core"` removed scale tuples are still a bugfix, but I should either remove the injection from here or add a disclaimer that it might be hard to adopt for existing projects.

hasparus:
:rocket: PR was released in [`v0.16.2`](https://github.com/system-ui/theme-ui/releases/tag/v0.16.2) :rocket:

PR #2461, opened 2023-11-29
Remove deprecated ThemeProvider export

lachlanjc:
It's been a year since I deprecated `ThemeProvider`, so dropping the deprecated code now.


hasparus (approved):
lgtm after fixing the build

PR #2552, opened 2025-12-26
components: [Alert] Add ARIA role

lachlanjc:
The Alert component previously was a div with no semantics. Adding role=alert to improve a11y.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role


hasparus:
@copilot can you update the snapshot here? This is intended diff.

```
      <div
        className="emotion-0"
    +   role="alert"
      />
```

PR #2553, opened 2025-12-26
components: [Textarea] Add field-sizing: content

lachlanjc:
Enabling auto-sizing textareas by default now that it's built into CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/field-sizing


hasparus:
Hey @copilot please update the snapshot so tests pass.

+   field-sizing: content;

